PR #10, opened 2015-05-06
Add a per-task enabled flag

aderuwe:
Per task - so no changes to any *Command.

### DIFF
--- a/src/GrumPHP/Runner/TaskRunner.php
+++ b/src/GrumPHP/Runner/TaskRunner.php
@@ -55,11 +55,13 @@ class TaskRunner
         $messages = array();
 
         foreach ($this->getTasks() as $task) {
-            try {
-                $task->run($files);
-            } catch (RuntimeException $e) {
-                $failures = true;
-                $messages[] = $e->getMessage();
+            if ($task->isEnabled()) {
+                try {
+                    $task->run($files);
+                } catch (RuntimeException $e) {
+                    $failures = true;
+                    $messages[] = $e->getMessage();
+                }
             }
         }
 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ parameters:
         phpunit: ~
 ```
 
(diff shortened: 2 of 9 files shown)


veewee:
Not sure I like this approach:
In my opinion, this will result in grumphp.yml files with permanent 'enabled: false' configuration.
So why not just remove the task? It doesn't make sense to configure something that is never being executed.

Maybe some ideas about a better solution:
- Replace the enabled flag with a `non-blocking` flag. The task will always be executed. When it fails you won't receive the green success message, but a yellow warnings message. However: The commit will be executed. The user will still have a visual message that shows that there are issues with the commit, but those issues are non-blocking. e.g.: you can commit a failing unit test, but receive a warning.
- Add a `threshold` flag per task. If e.g. 10% of the unit tests are failing, you can't commit. If only 5% is failing, you will be able to commit, but with warnings.

Maybe we should open a RFC about these kind of implementations before opening PRs?

src/GrumPHP/Runner/TaskRunner.php
```diff
@@ -55,11 +55,13 @@ public function run(FilesCollection $files)
         $messages = array();
 
         foreach ($this->getTasks() as $task) {
-            try {
-                $task->run($files);
-            } catch (RuntimeException $e) {
-                $failures = true;
-                $messages[] = $e->getMessage();
+            if ($task->isEnabled()) {
```
veewee: Inception!

aderuwe:
That's kinda my use-case: a project where most other developers will run maybe one task while I will run a lot more - not easy to get people to adopt something that is essentially pretty invasive. ;)

veewee:
Those people will not adopt the tools if the tasks aren't executed: they will not notice there is something wrong with the code they are trying to commit.
Maybe by 'notifying' those people, they will start to care more?

veewee:
As discussed earlier, this feature will be replaced with:
https://github.com/phpro/grumphp/issues/17

PR #14, opened 2015-05-24
Added Php-cs-fixer support

klipach:
There is very cool tool [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) it allows to fix codestyle errors automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 matrix:
   include:
-    - php: 5.3.3
+    - php: 5.3
     - php: 5.4
     - php: 5.5
     - php: 5.6
--- a/README.md
+++ b/README.md
@@ -217,6 +217,49 @@ If you want to run a particular suite only, specify it with this option.
 When this option is enabled, behat will stop at the first error. This means that it will not run your full test suite when an error occurs.
 
 
+#### PHP-CS-Fixer
+
+The PHP-CS-Fixer task will run codestyle checks.
+It lives under the `phpcsfixer` namespace and has following configurable parameters:
+
+
+**auto_fix**
+
+*Default: false*
+
+If you want to fix codestyle automatically specify it with this option.
+
+
+**config**
+
+*Default: `default`*
+
+There such predefined configs for codestyle checks: `default`, `magento`, `sf23`.
+If you want to run a particular config, specify it with this option.
+
+
+**filters**
+
+*Default: array()*
+
+There are a lot of fixers which you can apply to your code. You can specify an array of them in this config.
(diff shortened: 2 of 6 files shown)


aderuwe:
Can you update the php version requirement in composer.json so that travis can do its thing?
The PR looks good otherwise on first glance, but I'll look more later.

@veewee I think bumping the php version to 5.3.6 should be ok if we decide to accept this PR?

klipach:
Please, check, i updated PHP version and fixed codestyle errors.

veewee:
Hello @klipach 
Thanks for adding the php-cs-fixer task!
I do have some additional questions/remarks:

It sounds 'scary' to alter files during a commit. You won't have control about those files anymore. It's a good thing to make the `auto_fix` configurable and set it to false by default. Maybe an extra note could be added to the `README` file to use that option with care. Also note that the changed files won't be committed. You will need an additional `git add` to commit the changed files during the pre-commit hook.

Could you also add a configurable `--config-file` option? We tend to use the `.php_cs` file to determine the fixers, files, ... for the project. Or does this make the fixer run for all files?

These are some general remarks about the future structure of GrumPHP. These might not be fixed in this issue, but I would like to have some input on this:

Another remark I have concerning this auto_fix variable is the priority of the tasks. By changing the files you might break some existing tasks. Maybe it's a good idea to add a task priority per task configuration? For example: run the php-cs-fixer task first, next all other tests. 

At the moment the Task::run command is rather bug. We might introduce a 'Formatter' interface to display success / warning / error messages. Not sure how it will look for the moment, but it will remove the formatting stuff from the Task::run command. It seems like a good idea to me.

Another thing that ennoys me in the Task::run commands is this part: 
```php
if ($config['whatever']) { 
    $processBuilder->add(); 
}
 ```
Maybe it's a good idea to create a wrapper around the Symfony ProcessBuilder with following additional methods:
```php
// Add an optional configuration option. This will not add anything if the config is empty
$wrapper->addOptionalAttribute('--config=%s', $config['config']); 
// Add a required configuration options. Throws exception when the config is empty
$wrapper->addRequiredAttribute('--config=%s', $config['config']);
// Clone the processbuilder wrapper
$wrapper->clone()
```

veewee:
Hi @klipach,

Me and @aderuwe discussed this issue. It seems like a good idea to drop the `auto_fix` field. 
It seems like a better idea to let the task fail and show the suggestion to fix it manually.
Otherwise the user won't have any control about what he/she is commiting.

The `--config-file` option might still be usefull.

All other remarks are placed in new issues: #16, #18 and #19.

Looking forward to hear your feedback.

klipach:
Thank you for comments, i agree that `auto_fix` is no cool feature, so i removed it.

PR #21, opened 2015-05-28
Configurable ascii images

klipach:
Some people have bad sense of humor, so i added some configs for ascii images. Now you can change them or even disable.

### DIFF
--- a/src/GrumPHP/Configuration/GrumPHP.php
+++ b/src/GrumPHP/Configuration/GrumPHP.php
@@ -56,4 +56,25 @@ class GrumPHP
 
         return $tasksConfig[$taskName];
     }
+
+    /**
+     * Get ascii content path from grumphp.yml file
+     *
+     * @param $resource
+     *
+     * @return mixed
+     */
+    public function getAsciiContentPath($resource)
+    {
+        if (!$this->container->hasParameter('ascii')) {
+            return '';
+        }
+
+        $paths = $this->container->getParameter('ascii');
+        if (!array_key_exists($resource, $paths)) {
+            return '';
+        }
+
+        return $paths[$resource];
+    }
 }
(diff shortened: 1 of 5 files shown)


src/GrumPHP/Console/Helper/PathsHelper.php
```diff
@@ -80,9 +80,16 @@ public function getAsciiPath()
      */
     public function getAsciiContent($resource)
     {
-        $file = $this->getAsciiPath() . $resource . '.txt';
-        if (!$this->fileSystem->exists($file)) {
-            throw new RuntimeException(sprintf('ASCII file %s could not be found.', $resource));
```
klipach: I think it is better to continue script execution even if there is no ascii image

src/GrumPHP/Configuration/GrumPHP.php
```diff
@@ -56,4 +56,25 @@ public function getTaskConfig($taskName = null)
...
     }
+
+    /**
+     * Get ascii content path from grumphp.yml file
+     *
+     * @param $resource
+     *
+     * @return mixed
+     */
+    public function getAsciiContentPath($resource)
```
veewee: Can you provide a spec for this method?

veewee:
Hello @klipach,

Thanks again for contributing! This sure looks like a handy feature.
A little sidenode: We are planning to create a mascot for GrumPHP. Afterwords the current success and error message will be replaced by this mascot. We are trying to keep GrumPHP neutral yet fun to work with :)

Can you please add a spec for the Configuration method and fix CS.
I will try to make some time during the weekend to merge, test and release a new version.

veewee:
Hi @klipach ,
I fixed the code and added some tests.
Also the default configuration is made explicit. This way the services.yml file is the only source of truth.
Now it is also possible to rename the ASCII files without changing the code.

Thanks!

klipach:
Great! Sorry was off for several days was not able to fix comments.

veewee:
No problem! :)

PR #22, opened 2015-06-03
Added documentation for auto install via composer

klipach:
It is hard to add `pre-commit` hook for each developer manually. I added to README how to install GrumPHP checks for each developer via `composer install` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,25 @@ parameters:
 ### Set up basic configuration
 GrumPHP comes shipped with a configuration tool. Run following command to create a configuration file:
 ```
-php ./vendor/bin configure
+php ./vendor/bin/grumphp configure
 ```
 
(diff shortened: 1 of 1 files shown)


veewee:
Hello @klipach,

The actions you describe should not be executed.
This package contains a composer plugin, which hooks into the composer eventsystem.
When GrumPHP is installed / updated the `git:init` and `configure`  are automatically triggered.
When a `grumphp.yml` is already available, the configure command will do nothing. 
So when you commit the `grumphp.yml` file in git and ad grumphp to the composer dev-dependencies it should work for every person who installs it.

After composer finishes it's installation, you will see the message:
GrumPHP is sniffing your commits!

Is this feature not working for you? Because if that's so, you found a bug :)

klipach:
Thank you for explanation, will check it again.

PR #29, opened 2015-08-10
Allow phpunit to resolve the default config

NoUseFreak:
First of great tool!

I wanted to use this with a library not containing a `phpunit.xml` file. It does however have a `phpunit.xml.dist` file to which phpunit would resolve if the first was not found. ([documentation](https://phpunit.de/manual/current/en/textui.html#textui.examples.filter-shortcuts)). 

Removing the default setting allows packages to ship with a more basic config.


aderuwe:
:+1:

veewee:
Good idea. Thanks!

veewee:
The documentation is updated and this improvement is tagged as v0.3.1

PR #30, opened 2015-08-13
Check only the files that are staged for commit

mikechernev:
A small tweak to check only the files staged for commit. Currently it's checking all the files and it's not very convenient if you have files that you are still working on.

### DIFF
--- a/spec/GrumPHP/Locator/ChangedFilesSpec.php
+++ b/spec/GrumPHP/Locator/ChangedFilesSpec.php
@@ -4,6 +4,7 @@ namespace spec\GrumPHP\Locator;
 
 use Gitonomy\Git\Diff\Diff;
 use Gitonomy\Git\Diff\File;
+use Gitonomy\Git\WorkingCopy;
 use Gitonomy\Git\Repository;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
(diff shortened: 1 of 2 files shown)


aderuwe:
I actually like this change - I think it will make GrumPHP more accessible.

@veewee what's your take?

@mikechernev in any case, if this is to be merged, you'll need to update the specs accordingly.

veewee:
Looks good to me! This is indeed the way GrumPHP should work.
I played around with the git diff commands and I guess this commit will  also fix this issue:
https://github.com/phpro/grumphp#grumphp-does-not-work-on-first-commit

@mikechernev Can you fix the build please?
Thanks for contributing!

mikechernev:
Hey guys,

I am glad you like the change :)
I updated the specs, but since I don't have any previous experience with PhpSpec I am not sure I did it in the best possible way.
Please let me know if you know a better way to achieve this ;)

Cheers,
Mike

veewee:
That is exactly the way it should be done. Congratulations, you just received the Phpspec Ninja badge! ;)

veewee:
This commit is tagged as: https://github.com/phpro/grumphp/releases/tag/v0.3.2

mikechernev:
Awesome :)

PR #32, opened 2015-08-29
Added: Blacklist task

igormukhingmailcom:
Hi guys.

Here is new task. 
It cancel commit if blacklisted keywords will be found and prevent adding `var_dump(`, `die(` and whatever you want to your commits.

Hope this will be useful for other developers.

--
Thank you for really cool tool.

### DIFF
--- a/spec/GrumPHP/Task/BlacklistSpec.php
+++ b/spec/GrumPHP/Task/BlacklistSpec.php
@@ -70,7 +70,7 @@ class BlacklistSpec extends ObjectBehavior
 
         $process->run()->shouldBeCalled();
 
-        # Assume that blacklisted keywords was not found by `git grep` process
+        // Assume that blacklisted keywords was not found by `git grep` process
         $process->isSuccessful()->willReturn(false); 
 
         $files = new FilesCollection(array(
@@ -79,7 +79,7 @@ class BlacklistSpec extends ObjectBehavior
         $this->run($files);
     }
 
-    function it_throws_exception_if_the_process_successfull(GrumPHP $grumPHP, LocatorInterface $externalCommandLocator, ProcessBuilder $processBuilder, Process $process)
+    function it_throws_exception_if_the_process_is_successfull(GrumPHP $grumPHP, LocatorInterface $externalCommandLocator, ProcessBuilder $processBuilder, Process $process)
     {
         $this->beConstructedWith($grumPHP, array('keywords'=>array('var_dump(')), $externalCommandLocator, $processBuilder);
 
--- a/spec/GrumPHP/Task/BlacklistSpec.php
+++ b/spec/GrumPHP/Task/BlacklistSpec.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace spec\GrumPHP\Task;
+
+use GrumPHP\Collection\FilesCollection;
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Locator\LocatorInterface;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Symfony\Component\Process\Process;
+use Symfony\Component\Process\ProcessBuilder;
+use SplFileInfo;
+
+class BlacklistSpec extends ObjectBehavior
+{
+
+    function let(GrumPHP $grumPHP, LocatorInterface $externalCommandLocator, ProcessBuilder $processBuilder)
+    {
+        $this->beConstructedWith($grumPHP, array(), $externalCommandLocator, $processBuilder);
+    }
+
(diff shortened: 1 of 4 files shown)


joostfaassen:
:+1:

igormukhingmailcom:
@veewee Wow, so fast. Thank you. 
Can you please also make release?

src/GrumPHP/Task/Blacklist.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $config = $this->getConfiguration();
+        if (empty($config['keywords'])) {
+            throw new RuntimeException(
+                "Your grumphp config file misconfigued. You have empty keywords list under blacklist task."
+            );
+        }
+
+        $this->processBuilder->setArguments(array(
+            'git',
```
veewee: git is already specified at the getCommandLocation() method. I don't think this should be here?

veewee:
Hi @igormukhingmailcom,

Thanks for this great feature! 
I have added one question in the sourcecode.
Can you also provide some specs for this command?

Tomorrow I am taking a 2 weeks break. I won't be having access to the internet.
Maybe @aderuwe can tag a new release for this feature.

Thanks!

veewee:
Maybe we should also think about which keywords should be blacklisted by default.
I am thinking of leaving the default list empty so that the end-user explicitly has to determine which keywords should be blocked.

src/GrumPHP/Task/Blacklist.php
igormukhingmailcom: Yep, I thought also the same.
But if I remove `git` from `$this->processBuilder->setArguments` - it will run just `grep`. Don't know why because i'm new with Process Component. So I've left this as is...

igormukhingmailcom:
@veewee Actually code throws exception when keywords list are empty.
But I can rewrite if you want. Let me know, please.

src/GrumPHP/Task/Blacklist.php
veewee: This is because the `setArguments()` overwrites all arguments. The command is set as an attribute internally, so it will be overwritten. You could use the 'add()' method instead.

veewee:
When there are no blacklisted keywords, the test should not fail I guess. There is nothing wrong, but it would be better to disable the test in configuration when you don't want to use it.

igormukhingmailcom:
@aderuwe @veewee 
Done. 
Can you please merge and make release?

igormukhingmailcom:
@aderuwe ping :)

aderuwe:
@igormukhingmailcom I'll take a look at this later today, just got back from vacation ;)

spec/GrumPHP/Task/BlacklistSpec.php
```diff
@@ -0,0 +1,100 @@
...
+        # Assume that blacklisted keywords was not found by `git grep` process
+        $process->isSuccessful()->willReturn(false); 
+
+        $files = new FilesCollection(array(
+            new SplFileInfo('file1.php'),
+        ));
+        $this->run($files);
+    }
+
+    function it_throws_exception_if_the_process_successfull(GrumPHP $grumPHP, LocatorInterface $externalCommandLocator, ProcessBuilder $processBuilder, Process $process)
```
aderuwe: `it_throws_exception_if_the_process_is_unsuccessful`

spec/GrumPHP/Task/BlacklistSpec.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $processBuilder->setArguments(Argument::type('array'))->shouldBeCalled();
+        $processBuilder->add('-e var_dump(')->shouldBeCalled();
+        $processBuilder->add('-e die(')->shouldBeCalled();
+        $processBuilder->add('file1.php')->shouldBeCalled();
+        $processBuilder->getProcess()->willReturn($process);
+
+        $process->run()->shouldBeCalled();
+
+        # Assume that blacklisted keywords was not found by `git grep` process
```
aderuwe: Please use the // style of commenting which we've used in other places.

aderuwe:
@igormukhingmailcom if you could take a look at those two remarks, I think this PR is good to go otherwise.

igormukhingmailcom:
@aderuwe Done. Check please :)

igormukhingmailcom:
@aderuwe Can you also make release? It supposed to be 0.4 i think...

aderuwe:
@igormukhingmailcom I think we can handle this one with a patch number increase.

It's released as v0.3.3.

igormukhingmailcom:
Thank you. Have a nice day.

PR #38, opened 2015-09-14
Fixed: Use config path from composer's extra at config command

igormukhingmailcom:
Hi again.
One more think I've missed first time...
(Tag it please :))

### DIFF
--- a/src/GrumPHP/Console/Application.php
+++ b/src/GrumPHP/Console/Application.php
@@ -44,12 +44,32 @@ class Application extends SymfonyConsole
                 'config',
                 'c',
                 InputOption::VALUE_OPTIONAL,
-                '.',
-                getcwd() . DIRECTORY_SEPARATOR . self::APP_CONFIG_FILE
+                'Path to config',
+                $this->getConfigDefaultPath()
             )
         );
 
         return $definition;
+    }
+
+    /**
+     * @return string
+     */
+    protected function getConfigDefaultPath()
+    {
+        $defaultConfigName = getcwd() . DIRECTORY_SEPARATOR . self::APP_CONFIG_FILE;
+        $composerFile = 'composer.json';
+
+        if (!file_exists($composerFile)) {
+            return $defaultConfigName;
+        }
+
+        $composer = json_decode(file_get_contents($composerFile), true);
+        if (isset($composer['extra']['grumphp']['config-default-path'])) {
+            return $composer['extra']['grumphp']['config-default-path'];
+        }
+
+        return $defaultConfigName;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


veewee:
Hi @igormukhingmailcom,

Can't we use the --config=x parameter of the command for this?
If you want to create configuration on a specified location, you should use:

```
grumphp configure --config=path/to/config.yml
```

This way we don't need the commands to know the location of the composer file, config file, ... on the machine.

igormukhingmailcom:
You're right.
I've totally reworked this.
Now default config path getting right when `--config` option defining. So we don't need any code changes in another commands.

```
$ vendor/bin/grumphp 
GrumPHP version 0.1.0

Usage:
  command [options] [arguments]

Options:
  -h, --help             Display this help message
  -q, --quiet            Do not output any message
  -V, --version          Display this application version
      --ansi             Force ANSI output
      --no-ansi          Disable ANSI output
  -n, --no-interaction   Do not ask any interactive question
  -c, --config[=CONFIG]  . [default: "config/lib/grumphp.yml"] # <-------------- Default value given from composer.json's extra
  -v|vv|vvv, --verbose   Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Available commands:
  configure       
  help            Displays help for a command
  list            Lists commands
 git
  git:deinit      
  git:init        
  git:pre-commit  
```

src/GrumPHP/Console/Application.php
```diff
@@ -53,6 +53,26 @@ protected function getDefaultInputDefinition()
     }
 
     /**
+     * @return string
+     */
+    protected function getDefaultConfigName()
```
veewee: This still needs to return the absolute path for windows compatibility.

igormukhingmailcom: It seems like this value will be converted automatically to absolute when it will be used with commands: https://github.com/phpro/grumphp/blob/master/src/GrumPHP/Console/Application.php#L121
Right or I missed something? (I'm not sure)

In the other hand, relative path in command output (when you run just `grumphp` or `grumphp list`) is more readable.

Function name `getConfigDefaultPath` even better than `getDefaultConfigPath` (exact as extra key named).

WDYT?

igormukhingmailcom: Looks like you're right. I will redo this.

igormukhingmailcom:
Guys, any chance to merge/tag this today? 
I need this very hard :)

aderuwe:
I'm not too familiar with the code in question - that said, it looks ok to me.
Ping @veewee.

I'll merge and tag if @veewee doesn't make it back here today. :)

igormukhingmailcom:
Thanks. Will waiting :)

aderuwe:
Released as https://github.com/phpro/grumphp/releases/tag/v0.3.5

veewee:
It looks fine to me too!
Maybe we might change it in the future to locate the composer file in a subdirectory of the git folder etc.
But it will do for now.

igormukhingmailcom:
Thanks guys. 
Thanks to you we have cool (conventions-checker)[https://github.com/linkorb/conventions-checker) now.

Issue #33, opened 2015-09-01
No tasks runs when no files changed


igormukhingmailcom:
Hi.

When I run `php ./vendor/bin/grumphp git:pre-commit` manually (and have no changed files) - I see green man even shouldn't. So, no tasks executed.

But tasks such as 'phpunit', 'phpspec', 'behat' need to be executed.

veewee:
You want GrumPHP to validate the full codebase? Currently we only run tasks when files  have changed.
I am thinking of some continious integration commands that can run all configured tasks on the codebase.  More info at #20

igormukhingmailcom:
Tasks like phunit/behat/phpspec not executed on separate files. 
Its executed as is and any file list not passed to it as argument.
So i think that independent types of tasks need to run even if no file changed.

veewee:
If no PHP file has changed during pre-commit, why should these tasks be ran?
Normally the tests and code should be in the previous committed state and shouldn't contain any errors.

For example: a front-end designer wants to add a CSS file, and all of a sudden he gets an exceptions on the unit tests which he does not know anything about. This doesn't seem to make much sense to me.

What is your opinion @aderuwe ?

aderuwe:
@igormukhingmailcom I spoke to @veewee about this for a couple minutes this morning.
How about we introduce a grumphp:ci (or something similar) command that runs all the tasks on all of the files, while we make it so that "normal" GrumPHP usage is based off of the changed files only? (For independent tasks, as you call them, that would mean we trigger the entire suit, but only if there's changed files.)
What do you think?

igormukhingmailcom:
@aderuwe I thing that its a great idea.
I propose to name this command something like `grumphp run` (similar as `php-cs-fixer run` or `phpspec run`) or `grumphp check`, because this command will be used not only with CI.

igormukhingmailcom:
Also option `--skip=task` will be helpful I think... 

For example:

-  `grumphp run --skip=phpunit` to skip phpunit tests
-  `grumphp run --skip=phpunit,behat` to skip phpunit and behat tests

WDYT?

veewee:
sounds good! This will require some context in which the tasks needs to run.
Also: some tasks will not be able to run, like the blacklist task, since this only checks files that are being committed. So the tasks need to know in which context they can run. 
During commit, only the changed files should be added.
During run, all files should be added. (Note: wouldn't this be slow on big codebases?)

This feature will require some thinking about the architecture of the project and propably BC breaks.

[...]

PR #40, opened 2015-09-16
Refactoring for running the tasks in different contexts

veewee:
This major refactoring has some BC breaks.
It is possible to specify which task can run in which context.
A console helper is created that is responsible for running the tasks in the console commands.
This way it is easy to reuse the taskrunner in multiple console commands: You will only need to pass a new context.

This PR will make it possible to develop #33.

@aderuwe, @igormukhingmailcom : Can you please review and tell me if you like it or not?

### DIFF
--- a/spec/GrumPHP/Collection/TasksCollectionSpec.php
+++ b/spec/GrumPHP/Collection/TasksCollectionSpec.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace spec\GrumPHP\Collection;
+
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\TaskInterface;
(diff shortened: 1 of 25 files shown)


igormukhingmailcom:
@veewee Looks very cool. I like :)

igormukhingmailcom:
Can you please tag it as 2.0.0 (now or in future when it will be stable) as it has BC break as you said.

veewee:
Hello @igormukhingmailcom,
Since the project is not stable enough, I will be tagging stuff as 0.(major structural changes).(patch or small feature).
The PR is tagged as 0.4.0.
There are still some things I want to change in the codebase before tagging a stable major release.

PR #43, opened 2015-09-18
Commit message task

veewee:
This commit adds a new git hook: `commit:msg`.
You will be able to match patterns to git commit messages.
In the future additional tasks like a check on username etc. can be added. (For example: user vagrant can not commit)

What do you think @aderuwe, @igormukhingmailcom  ?
It's not a code-quality task, but it is always nice to have some control about the commit messages.
For example validating a jira issue number in the commit msg etc.


igormukhingmailcom:
This is awesome :) Love this.

PR #44, opened 2015-09-22
Update Phpcsfixer.php

TheRatG:
Bug fix - Undefined index: config-file

### DIFF
--- a/src/GrumPHP/Task/Phpcsfixer.php
+++ b/src/GrumPHP/Task/Phpcsfixer.php
@@ -71,7 +71,7 @@ class Phpcsfixer extends AbstractExternalTask
         }
 
         if ($config['config_file']) {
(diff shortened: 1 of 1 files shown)


src/GrumPHP/Task/Phpcsfixer.php
```diff
@@ -71,7 +71,7 @@ public function run(ContextInterface $context)
         }
 
         if ($config['config_file']) {
-            $this->processBuilder->add('--config-file' . $config['config-file']);
+            $this->processBuilder->add('--config-file' . $config['config_file']);
```
veewee: Shouldn't there also be an equal sign between the option and the value? Like this: `config-flle=path`

TheRatG: I've just fix it by readme.doc :smile:

veewee:
Thanks!

PR #45, opened 2015-09-25
phpunit task modifications

ecentinela:
Some tests throw exceptions with the message 'Unable to guess the Kernel directory.' In symfony.

I don't know why (I thought -c is the same as --configuration), but changing this, fixes the problem.

### DIFF
--- a/src/GrumPHP/Task/Phpunit.php
+++ b/src/GrumPHP/Task/Phpunit.php
@@ -57,7 +57,7 @@ class Phpunit extends AbstractExternalTask
         ));
 
         if ($config['config_file']) {
-            $this->processBuilder->add('-c' . $config['config_file']);
+            $this->processBuilder->add('--configuration=' . $config['config_file']);
         }
 
         $process = $this->processBuilder->getProcess();


veewee:
Thanks fo the PR. Can you take a look at the comment?

ecentinela:
I know, but when you have a big test suite, is annoying not to see any progress. It seems the process is hanged. What do you think if I create an option "show_progress" in grumphp.yml to enable it / disable it at least?

veewee:
Maybe it's a better idea not to use the output of all posible tasks, but use a custom progress indicator for GrumPHP?
For example: you configured 5 tasks but don't want to display all the output for all these tasks. 
You just want to know which task is running and how many tasks are left.

It might be a better solution to use the progressbar helper to display the progress of the tasks.
http://symfony.com/doc/current/components/console/helpers/progressbar.html

Every time a task ran, an event is triggered and the progressbar gets updated.
Does that solution works for you?

ecentinela:
It's ok for me.

Do you know how can I remove the commit from my PR?

I will create a new PR with the progress bar feature. Maybe you can give more information about where to start developing it?

Thank you.

veewee:
Hello @ecentinela,

You can remove the commit with `git rebase`.
More information can be found at: 
http://www.clock.co.uk/blog/deleting-a-git-commit

About the progress-bar. Maybe it's better to discuss this one in a new issue.
I created a small description at https://github.com/phpro/grumphp/issues/46

ecentinela:
Commit reverted. Feel free to merge this PR.

ecentinela:
Commit reverted. Feel free to merge this PR.

PR #47, opened 2015-10-18
Remove version for less documentation maintenance.

NoUseFreak:
Remove version for less documentation maintenance.

This will allow composer to determine the last stable version. 

```sh
$ composer require phpro/grumphp
Using version ^0.4.1 for phpro/grumphp
./composer.json has been created
Loading composer repositories with package information
...
```


veewee:
Thanks!

Issue #46, opened 2015-09-28
Display a progressbar while running the tasks


veewee:
When the test-suite grows, the tasks will run longer. Therefor it might be nice to display a progressbar while the tasks are running.

More information about the progressbar helper:
http://symfony.com/doc/current/components/console/helpers/progressbar.html

How should this work:
- The taskmanager should trigger some events: pre-run & post-run to specify which tasks ar running + the current context, pre-run-task to specify which task is running, post-run-task to tell the progressbar that a task has completed. 
- The console command helper should listen to those events and move the progress per task.
- The output should be something like: 'running phpspec'  Task 1/6

This feature will require some architecture changes which should be discussed before implementing.

Issue #42, opened 2015-09-18
Global grumphp or grumphp.phar (like phpunit)


TheRatG:
Hello, thanks for your library.

I suggest add new feature - run hook by grumphp.phar

Advantages:
* you don't need modify your composer
* global check your code command
* self-update packet

TheRatG:
https://github.com/box-project/box2 - create box

veewee:
That would indeed be a nice feature. Now it is possible to use `composer global require`, but you would have to change the path of the git hook manually.

igormukhingmailcom:
I can help with this. Have some experience with tasks like that :)

veewee:
Maybe it's a better idea to use composer for this?
```sh
composer global require phpro/grumphp
```

This way, GrumPHP will be installed globally in the ~/.composer/vendor/bin.
Update is as easy as=
```sh
composer global update phpro/grmphp
```

Looks easier then creating phar + self-update stuff, since everybody is using composer these days.
In the git hooks we will need to add something like:

```sh
PATH=$PATH:/full/path/to/project/vendor/bin
```
This way we can change the command to just `grumphp`. It will first search all your system include paths (in which the global composer vendor/bin directory is added). When no script is found, it will use the one from the project vendor/bin directory.

What are your thoughts on this solution? A phar looks one step too far for me, since it can easliy be done with composer.

veewee:
This one is implemented as mentioned above.
More information:
https://github.com/phpro/grumphp#global-installation

ihsanudin:
I think is better use phar as optional. Like others php tool (php-cs-fixer, deployer, and many more) has own phar file for simplifying the task.

I think many people more familiar with phar instead of global package in composer.

ihsanudin:
https://github.com/vinkla/climb/pull/28

ihsanudin:
The reason why you need phar file instead of global composer

Issue #20, opened 2015-05-26
Continious Integration CLI command


veewee:
It might be useful to use GrumPHP with CI servers like Jenkins.
An additional command needs to be created to run the tasks during Continious Integration.

Note that it should be possible to:
- Skip specific files / folders like e.g. the vendor folder.
- Add an optional threshold as a tag to the tasks services
- Add the output of specific tasks to the Jenkins server.
- ...

veewee:
This one is implemented. The threshold doesn't seem to be important at the moment.
Also tasks specific output is not available for the moment, but it can be added later.

PR #48, opened 2015-10-24
Introducing Events, Progress and a run command.

veewee:
This PR will make it possible to:

- Hook in GrumPHP through events
- Display progress when runnning tasks. This way you know WHY a commit is taking a long time. It is always enabled at the moment but could be configurable in the future.
- A basic run command that will run the commands on all the files that are registered to git.
- A new locator that can find out which files are and will be committed to git.
- Use the grumphp command system-wide

This will resolve issues: #46, #33, #20 and #42.
For know it seems enough for CI to be able to run the tasks with the run command.

Another issue should be opened to make it possible to skip tests when running the tasks.

Can you take a look at this @aderuwe, @igormukhingmailcom ? Thanks!

### DIFF
--- a/src/GrumPHP/Event/TaskFailedEvent.php
+++ b/src/GrumPHP/Event/TaskFailedEvent.php
@@ -12,7 +12,6 @@ use GrumPHP\Task\TaskInterface;
  */
 class TaskFailedEvent extends TaskEvent
 {
-
     /**
      * @var Exception
      */
--- a/src/GrumPHP/Event/TaskFailedEvent.php
+++ b/src/GrumPHP/Event/TaskFailedEvent.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace GrumPHP\Event;
+
+use Exception;
+use GrumPHP\Task\TaskInterface;
+
+/**
+ * Class TaskFailedEvent
+ *
+ * @package GrumPHP\Event
+ */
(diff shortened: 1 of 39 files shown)


src/GrumPHP/Event/TaskFailedEvent.php
```diff
@@ -0,0 +1,39 @@
...
+use GrumPHP\Task\TaskInterface;
+
+/**
+ * Class TaskFailedEvent
+ *
+ * @package GrumPHP\Event
+ */
+class TaskFailedEvent extends TaskEvent
+{
+
```
aderuwe: There's an extra empty line here.

src/GrumPHP/Task/Context/RunContext.php
```diff
@@ -0,0 +1,35 @@
...
+use GrumPHP\Collection\FilesCollection;
+
+/**
+ * Class RunContext
+ *
+ * @package GrumPHP\Task\Context
+ */
+class RunContext implements ContextInterface
+{
+
```
aderuwe: There's an extra empty line here.

aderuwe:
I'm :+1:

veewee:
I removed the newlines. However php-cs is not complaining and there are still other classes with extra newlines. Maybe the PHP-cs standard should be adjusted?

PR #56, opened 2015-11-13
Remove unused declarations

adam-paterson:
* Remove declarations
* Add newline character to `grumphp.yml` for better rendering on a terminal


veewee:
Much better this way! Thanks!

Issue #19, opened 2015-05-26
Create a wrapper for the ProcessBuilder


veewee:
Currently the Task::run commands has many of these if statements:
```php
if ($config['whatever']) { 
    $processBuilder->add(); 
}
```

It seems like a good idea to create a wrapper around the Symfony ProcessBuilder with following additional methods:

```php
// Add an optional configuration option. This will not add anything if the config is empty
$wrapper->addOptionalAttribute('--config=%s', $config['config']); 
// Add a required configuration options. Throws exception when the config is empty
$wrapper->addRequiredAttribute('--config=%s', $config['config']);
// Clone the processbuilder wrapper
$wrapper->clone()
```

PR #57, opened 2015-11-14
Make it easier to apply arguments to an external task. 

veewee:
This PR fixes issue #19.
A ProcessArgumentsCollection class is added which makes it much easier to configure the attributes of an external command in a task.

### DIFF
--- a/src/GrumPHP/Task/Behat.php
+++ b/src/GrumPHP/Task/Behat.php
@@ -50,7 +50,6 @@ class Behat extends AbstractExternalTask
     public function run(ContextInterface $context)
     {
         $files = $context->getFiles()->name('*.php');
-        // We don't care about changed files here, we want to run the entire suit every time
         if (0 === count($files)) {
             return;
         }
--- a/src/GrumPHP/Task/Behat.php
+++ b/src/GrumPHP/Task/Behat.php
@@ -2,6 +2,7 @@
 
 namespace GrumPHP\Task;
 
+use GrumPHP\Collection\ProcessArgumentsCollection;
 use GrumPHP\Exception\RuntimeException;
 use GrumPHP\Task\Context\ContextInterface;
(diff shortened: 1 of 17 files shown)


aderuwe:
I like how adding `ProcessArgumentsCollection` simplifies both the tasks and all the related specs. Big :+1: from me.

src/GrumPHP/Task/Phpspec.php
```diff
@@ -47,26 +48,20 @@ public function canRunInContext(ContextInterface $context)
     public function run(ContextInterface $context)
     {
         $files = $context->getFiles()->name('*.php');
+        // We don't care about changed files here, we want to run the entire suit every time
```
aderuwe: This comment is probably a leftover from something else? It seems that we don't run if no files are changed, looking at the next few lines.

src/GrumPHP/Task/Behat.php
```diff
@@ -49,31 +50,19 @@ public function canRunInContext(ContextInterface $context)
     public function run(ContextInterface $context)
     {
         $files = $context->getFiles()->name('*.php');
+        // We don't care about changed files here, we want to run the entire suit every time
```
aderuwe: This comment is probably a leftover from something else? It seems that we don't run if no files are changed, looking at the next few lines.

src/GrumPHP/Task/Phpunit.php
```diff
@@ -46,20 +47,17 @@ public function canRunInContext(ContextInterface $context)
     public function run(ContextInterface $context)
     {
         $files = $context->getFiles()->name('*.php');
+        // We don't care about changed files here, we want to run the entire suit every time
```
aderuwe: This comment is probably a leftover from something else? It seems that we don't run if no files are changed, looking at the next few lines.

PR #58, opened 2015-11-15
Update version constant

akira28:


veewee:
Thanks! I will try to keep the version number in sync.

PR #62, opened 2015-11-17
Load tasks from service container

adam-paterson:


adam-paterson:
it also achieves this
[SensioLabsInsight](https://insight.sensiolabs.com/projects/767dd7d5-4b2c-4181-b165-0cd3223900df)

adam-paterson:
If this is approved I'll squash the two silly mistakes above.

aderuwe:
:+1: 

@veewee and myself were discussing some potential refactoring regarding the configuration stuff touched upon here, but I think you're good for squashing the commits, we can always apply that later.

adam-paterson:
I'd love to know what the outcome of those discussions was.

I'll squash them shortly.

veewee:
As @aderuwe mentioned, the PR is ok.

Some feedback about the discussion. (But this doesn't have much to do with your PR)
We feel that the GrumPHP configuration class is being abused for the moment.
2 solutions come to mind:

1) 
Create a `TaskConfigurationResolver` which contains the logic for parsing task data. It wel look like:
```php
<?php

interface TaskConfigurationResolver
{

    /**
     * Returns the list off all configured tasks.
     * 
     * @return array
     */
    public function getConfiguredTasks();

    /**
     * Returns a list of all possible tasks.
     * 
     * @return array
     */
    public function getAllTasks();

    /**
     * Get the configuration from one specific task
     * 
     * @param string $task
     *
     * @return array
     */
    public function getTaskConfiguration($task);


    /**
     * This method will return some metadata about the task which can be configured in the tags of the service. 
     * For example: config key, order of execution, isBlocking flag, ...
     * 
     * @return TaskMetadata
     */
    public function getTaskMetadata();
}
```

2) Create a new compiler to parse all the data we want as described above and place the data in the 
parameters of the service containers.

The GrumPHP configuration class will then have shortcut functions to the resolver or the container parameters. (based on the coice we make)
The general idea is to simplify the creation of new tasks. For the moment an external task has allready 4 dependencies and a base class. This should become easier in the future.

Another thing that bothers me is that the configuration is not very transparant at the moment. Besides the documentation, you need to search the code to find out what configuration you need. It might be a good idea to use the Symfony Definition TreeBuilder to configure tasks in the future.

adam-paterson:
@veewee @aderuwe Commits squashed.

Excellent news, having a task builder is a good idea. It'll make it easier for other people to get involved in the project.

PR #64, opened 2015-11-18
Add prompt before overwritting hooks

adam-paterson:
All commands should really be tested with Behat, is this in the pipeline? I'm happy to set up some tests for this and the existing commands if you like.

### DIFF
--- a/src/GrumPHP/Console/Command/Git/InitCommand.php
+++ b/src/GrumPHP/Console/Command/Git/InitCommand.php
@@ -7,6 +7,7 @@ use GrumPHP\Console\Helper\PathsHelper;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\ProcessBuilder;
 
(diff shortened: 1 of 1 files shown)


veewee:
Hi @adam-paterson,

Currently the behat tests are not in the pipeline. Feel free to add some tests for the existing commands.
That would be great!

About the commit:
This is a great feature. One small remarks:
Maybe the question part can be moved to an IO object like in composer?

I haven't tested this yet, but this command is triggered during composer update / install. By default the hook will be overwritten, because this command does not run in interactive mode. When running the composer hook in windows, the command needs to be in interactive mode for some odd reason.
So by default, the hook will always be overwritten I guess.

adam-paterson:
@veewee Good point, however as a user I'd be irritated if a tool overwrote something I'd put into place without warning. I'll change the default to skip rather than overwrite.

On a side note, do you have a IM room somewhere I can join, got quite a few ideas for this project and wanted to run them by you.

aderuwe:
@adam-paterson I have just registered #grumphp on Freenode. I'll do some more setup on the channel later tonight, but let's use that; I'm usually on Freenode when working, and will try to hang out there in the evenings a bit too.

veewee:
Not sure if skip is better than overwrite: when the package upgrades, it is also possible that a git hook changed. So the git hook should be updated automatically.
On the other hand: when using GrumPHP, the hooks need to be the grumPHP hooks. Maybe it's better to create a copy to a backup file ?

adam-paterson:
Yeah, I'll creak backups and prompt the user to create a GrumPHP compatible task for whatever the hook did before.

aderuwe:
@adam-paterson Any update here? We'd love to have the feature. :)

adam-paterson:
@aderuwe Sorry, I've been slow on OSS projects recently. Started a new job and have been focussing on getting up to speed. I'll continue with this soon though.

aderuwe:
@adam-paterson Awesome! Good luck with the new job.

veewee:
I am closing this for now since a lot has changed in the past year. If you want this feature in the current codebase, please provide a new PR. Thanks!

PR #66, opened 2015-11-19
Unset the default value for the config option

alexgt9:
As --config value shadow the configuration used in the file selected by --config-file
and we was using always the default --config=default in the process,
the config in the file was never used


veewee:
Thanks!

alexgt9:
Thanks you for this tool :smile:

PR #69, opened 2015-11-20
Add Codeception task

adam-paterson:


veewee:
Hi @adam-paterson,

Thanks for the codeception task!
Can you change the task to use the new `ProcessArgumentsCollection`:
https://github.com/phpro/grumphp/blob/master/src/GrumPHP/Collection/ProcessArgumentsCollection.php

adam-paterson:
@veewee didn't notice that, amended. I'll squash if approved/

src/GrumPHP/Task/Codeception.php
```diff
@@ -0,0 +1,75 @@
...
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $files = $context->getFiles()->name('*.php');
+        if (0 === count($files)) {
+            return;
+        }
+
+        // We don't care about changed files here, we want to run the entire suit every time
```
aderuwe: A bit of a nit-pick, but we've been removing these comments from the tasks, they're pretty much a copy/paste relic from the past... :)

grumphp.yml
```diff
@@ -3,8 +3,4 @@ parameters:
   bin_dir: "./vendor/bin"
   git_dir: "."
   tasks:
-      phpcs:
-        standard: PSR2
-        ignore_patterns:
-          - "*Spec.php"
-      phpspec: ~
+      codeception: ~
```
aderuwe: Again, a nit-pick, but could you make sure this file ends in a new line?

aderuwe: Actually, I think this change is wrong and is causing the Travis build to fail - GrumPHP doesn't have any codeception tests, we do however want to be PSR2 compliant and run our specs. :)

aderuwe:
We should probably also add codeception to the "suggest" section of composer.json.

aderuwe:
:+1: feel free to squash the commits.

adam-paterson:
@aderuwe Done!

aderuwe:
Thanks!

PR #73, opened 2015-12-01
Make it easier to configure new tasks + Symfony 3 compatibility

veewee:
This PR will make it easier to configure tasks and fetch the configuration and metadata of a task.
Fixes #60 
Adds basic functionality for #17, #16 and #50.

PR #74, opened 2015-12-01
Make it possible to configure additional extensions.

### LINKED ISSUE #63: [Proposal] Plugins
# Context
I started building a similar library because I forgot this existed. I wanted to have a solution similar to Rocketeer which allowed people to create a directory in their project `.githooks` which contained a php config file with the tasks they wanted to do with particular git commands. The idea being companies can create their own plugins to do particular tasks. `post-commit` for example could post to a Slack channel, etc.

## Misc Tasks
It would allow people to easily create their own tasks with classes or callbacks. For example I could have a task that did a test merge with the parent branch to check for an conflicts before I push.

# Proposal
I don't want to reinvent the wheel and release a new library where 80% of the work is achieved with GrumPHP. But to make GrumPHP work how I'd like I'd need to refactor a lot of the structure. But it would allow users to specify only the plugins they required and have it all managed through composer as well as other tasks which might not have a library around them.

Any thoughts?

veewee:
This PR will make it possible to specify additional extensions. These extensions can hook into the service container. This means it is possible to create your own set of tasks and/or event listeners.

Fixes #63


README.md
```diff
@@ -523,6 +524,41 @@ Following events are triggered during execution:
 | grumphp.runner.complete | RunnerEvent       | when all tasks succeed
 
 
+## Extensions
+
+You will propably have some custom tasks or event listeners that are not included in the default GrumPHP project.
```
aderuwe: Typo: probably

grumphp.yml
```diff
@@ -7,4 +7,4 @@ parameters:
         standard: PSR2
         ignore_patterns:
           - "*Spec.php"
-      phpspec: ~
+      phpspec: ~
```
aderuwe: Shouldn't we keep the ending newline here?

aderuwe:
This needs a rebase, it looks like, after #73 was merged.

PR #77, opened 2015-12-10
Introduce the Grunt task

veewee:
This PR contains a configurable grunt task which can be used for some frontend checks before running a commit.


aderuwe:
Cool!

PR #78, opened 2015-12-10
Make it easier to configure task.

veewee:
This PR replaces the `getDefaultConfiguration()` method with a `getConfigurableOptions()` method.
By using the Symfony `OptionsResolver`, it is now easier to configure tasks in `grumphp.yml`.
When you type a typo or add a wrong configuration type, the resolver will complain and tell you what to do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "symfony/event-dispatcher": "~2.5|~3.0",
         "symfony/filesystem": "~2.4|~3.0",
         "symfony/finder": "~2.4|~3.0",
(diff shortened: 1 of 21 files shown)


aderuwe:
@veewee Some problems there with the low deps, seems that `addAllowedTypes()`'s signature changed at some point

veewee:
I bumped the version of the optionsresolver to >2.6.
In 3.0 the old way of configuring does not work anymore. I'd rather be compatible with 3.0 as with 2.3.

PR #81, opened 2015-12-14
Sort the tasks and make it easier to alter basic configuration.

veewee:
This PR splits the current configuration file into parameters, services and tasks configuration.
The tasks will be sorted to make it easier to configure the grumphp.yml file.
Also the --force flag is added to the configure command, to make it possible to recreate the grumphp.yml file from scratch.


src/GrumPHP/Configuration/Compiler/TaskCompilerPass.php
```diff
@@ -57,6 +57,10 @@ public function process(ContainerBuilder $container)
             $definition->addMethodCall('addTask', array(new Reference($id)));
         }
 
+        // Sort data:
+        sort($tasksRegistered);
+
+        // Register the data to the container:
```
aderuwe: I'm not sure that these two comments are useful.

aderuwe:
:+1:

veewee:
Rebased + fixed comments.

PR #82, opened 2015-12-14
Update version constant

aaa2000:


aderuwe:
@veewee I guess we should include this change in the release process. ;)

Thanks!

veewee:
Yes, I tend to forget that one from time to time... 
Maybe we could automate the tagging procedure.

PR #88, opened 2015-12-23
Introduce the YAML linter.

veewee:
This task improves the base linter classes + adds a YamlLint task.

### DIFF
--- a/resources/config/tasks.yml
+++ b/resources/config/tasks.yml
@@ -85,3 +85,12 @@ services:
             - '@linter.xmllint'
         tags:
             - {name: grumphp.task, config: xmllint}
+
+
+    task.yamllint:
+        class: GrumPHP\Task\YamlLint
+        arguments:
+            - '@config'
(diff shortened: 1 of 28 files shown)


resources/config/tasks.yml
```diff
@@ -85,3 +85,12 @@ services:
...
             - {name: grumphp.task, config: xmllint}
+
+
+    task.yamllint:
+        class: GrumPHP\Task\YamlLint
+        arguments:
+            - '@config'
+            - '@linter.yamllint'
+        tags:
+            - {name: grumphp.task, config: yamllint}
```
aderuwe: Nitpicking, but maybe add an ending newline here?

aderuwe:
:+1: Simplifications in LintError are cool, and having the AbstractLinterTask do the (abstracted) linting is, too.

(2 test fixtures also lack an ending newline, but I suppose it really doesn't matter there.)

PR #90, opened 2016-01-04
updated default parameter value

lykciv:
The default parameter value for `bin_dir` was wrong. I changed it to `./vendor/bin`.


veewee:
Thanks!

PR #93, opened 2016-01-23
Added security checker task

lykciv:
I have create a task which will run the SensioLabs Security Checker. This is a command line tool that checks if your application uses dependencies with known security vulnerabilities. It uses the SensioLabs Security Check Web service and the Security Advisories Database.

More information here: [https://github.com/sensiolabs/security-checker](https://github.com/sensiolabs/security-checker)

Tests and documentation are included in this pull requests.

### DIFF
--- a/src/GrumPHP/Task/SecurityChecker.php
+++ b/src/GrumPHP/Task/SecurityChecker.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace GrumPHP\Task;
+
+use GrumPHP\Exception\RuntimeException;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * SecurityChecker task
+ */
+class SecurityChecker extends AbstractExternalTask
+{
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return 'securitychecker';
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(array(
+            'lockfile' => null,
+            'format' => null,
+            'end_point' => null,
+            'timeout' => null,
+        ));
+
+        $resolver->addAllowedTypes('lockfile', array('null', 'string'));
+        $resolver->addAllowedTypes('format', array('null', 'string'));
+        $resolver->addAllowedTypes('end_point', array('null', 'string'));
+        $resolver->addAllowedTypes('timeout', array('null', 'int'));
+
+        return $resolver;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function canRunInContext(ContextInterface $context)
+    {
+        return ($context instanceof GitPreCommitContext || $context instanceof RunContext);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('security-checker');
+        $arguments->add('security:check');
+        if ($config['lockfile']) {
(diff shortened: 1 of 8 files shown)


src/GrumPHP/Task/SecurityChecker.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('security-checker');
+        $arguments->add('security:check');
+        if ($config['lockfile']) {
```
veewee: This one can be changed to a shorter version:

```
$arguments->addOptionalArgument('%s', $lockfile);
```

veewee:
Hi Vicky,

Thanks for the PR. It looks really good!
I added one little nitpick in the comments of the code.

Still 2 questions left:

Should it run on every commit? Maybe its better to run the task during the commit of composer.json or composer.lock? During CI it will still be tested. You can take a look at the composer task for reference.

Currently we are forcing the use of roave/security-advisories. This tool checks during composer install/update if the security advisories are being checked. It uses the same database ad the tool in this task. Should we use it together or is it better to remove the roave/security-advisories so that we do not push some tool to all the users of GrumPHP. In my opinition this is a better replacement for the roave package because the user can opt-in on this one. What is your opinion on this? (CC @aderuwe )

src/GrumPHP/Task/SecurityChecker.php
```diff
@@ -0,0 +1,74 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * SecurityChecker task
+ */
+class SecurityChecker extends AbstractExternalTask
+{
+    /**
+     * @return string
+     */
```
aderuwe: `{@inheritdoc}` is fine here, and in some other places below which implement `TaskInterface`.

aderuwe:
I agree on both points, this is a much nicer way of doing the same thing, but should probably only run on  commits containing composer.* changes.

(Removing the roave/security-advisories dependency is kind of a BC break (in case users don't opt-in on the task) but it should be ok in 0.x releases.)

lykciv:
I think it could be valuable if the task runs on every commit. Especially in projects which are in maintenance mode. Usually composer doesn't get updated often enough on these projects or developers forget to do so when they do have to work on it. So, when this task runs on every commit, at least developers are alerted of any security issues in their dependencies.

On the other hand, if a project is in active development, this would probably be overkill.

So, maybe this should be a configurable option? Give users the choice to run this task on every commit or only when the composer files are changed.

You could use the security checker and roave/security-advisories side by side, but this allows you to remove it as a dependency and just keep it as a suggestion. Some users might prefer to be alerted of any security issues when they require or update a dependency, so I would definitely keep the suggestion.

But I think the security checker is a better tool because it runs against a web service and always has the most recent information. The roave/security-advisories tool only runs against the version you have in your project and if you don't update composer, this won't help you.

veewee:
Just took another look at the roave security advisories. Currently they are only suggested. As @lykciv mentioned, it's better to use the cli tool, because it's always up to date even when you don't update composer. We keep the roave package in the suggestions. This way the user has the choice: run during composer update or run on every commit.

Still, I would like to have a configurable option for triggering the security advisories cli tool. The roave package breaks during composer update, but in this task we want to break during a git commit. So that are 2 different places where the security validation is being done. Maybe something like `run_always` which defaults to `false`? (Any better option name is also good :) )

The reason why I would set is to false is that most projects are also running some kind of CI tool. When CI triggers to `run` command, the composer.json file is always in git and will always check the security advisories database. This way your commit will pass, because there is nothing wrong with the commit. CI will fail because of security issues. This can be handled in a next commit / on another branch. Does that sounds good to you?

lykciv:
Sounds good to me. I've updated the code to include the `run_always` option. I've also updated tests and documentation accordingly.

veewee:
Looks great! Thanks!

veewee:
@lykciv 
This new feature is released under 0.7.2 (https://github.com/phpro/grumphp/releases/tag/v0.7.2)

PR #98, opened 2016-02-01
Allow further modifiers in regular expression for commit messages

michaelgerdemann:


veewee:
Hello @michaelgerdemann,

Thanks for your contribution. There are some issues with this PR:

- We like to use snakecase for the configurable options. This means `additional_modifiers` instead of `additionalModifiers`.
- Spec tests for this option is missing + failing tests
- Documentation for this option is missing.

Just out of curiousity: which modifier would you like to add? 
I added the `case_insensitive` for glob-style matching and a bit later the `multiline` was added.
Maybe it's better to just add the modifiers to the configured matcherin the configuration, so that we don't have to add all the modifiers. The only useful option for glob-style matchers is case_insensitive.

Sample configuration would look like:

```yaml
parameters:
    tasks:
        git_commit_message:
            matchers:
                - /someregex/i
                - /someotherregex/ix
```

This is allready possible with the current code.
(/cc @aderuwe: What is your view on this?)

michaelgerdemann:
Hello @veewee,

i need the modifier "u" for umlauts in commit messages.
I have no idea how to write spec tests, could you help me there?

src/GrumPHP/Task/Git/CommitMessage.php
```diff
@@ -91,6 +93,8 @@ public function run(ContextInterface $context)
                 $regex->addPatternModifier('m');
             }
 
+            array_map(array($regex, 'addPatternModifier'), str_split((string) $config['additional_modifiers']));
```
veewee: Currently the tests fail because when the additional modifiers are empty, `str_split` is returning an array with a empty string. You will have to add `array_filter` around the str_split method or an is-empty check around the array_map method.

veewee:
I sure can help you with that.
For this commit, I don't think there should change a lot:

`it_should_have_configurable_options()` should check for additional_modifiers.
Next all tasks should pass.
You could also add an additional test-case like:

```php
function it_runs_with_additional_modifiers(GrumPHP $grumPHP, GitCommitMsgContext $context)
{
    $grumPHP->getTaskConfiguration('git_commit_message')->willReturn([
        'matchers' => ['ümlaut'],
        'additional_identifiers' => 'u',
    ]);

    $context->getCommitMessage()->willReturn('message containing ümlaut');

    $this->run($context);
}
```

Note that I did not test the regex.

veewee:
Thanks for this new feature!

veewee:
@michaelgerdemann 
This new feature is released under 0.7.2 (https://github.com/phpro/grumphp/releases/tag/v0.7.2)

Issue #100, opened 2016-02-03
Fail when file is both in staging and not


keradus:
Hi. First of all - nice tool ! ;)

Maybe it's a matter of configuration, but please consider the following scenario:
- configure tool to use PHP CS Fixer, init git hooks
- create foo.php file with CS violations
- `git add foo.php`, `git commit -m "foo"` -> fails due to PHP CS Fixer violations
- fix violations without adding the file (so `git status` will show unfixed-files in stage and fixed one in non-stage)
- `git commit -m "foo"` will proceed, even if the file we are commiting is not valid

Please, could you help?

veewee:
Hello @keradus,

Glad you enjoy our tool!
About your problem: this has been reported before.
Please take a look at issue: https://github.com/phpro/grumphp/issues/67

This is not something that can be fixed, since phpcsfixer is running on the actual file. At the point of the last commit, the file has no CS issues and therefor you are able to commit.

keradus:
2 (alternative) hints then:
- fail PHP CS Fixer task for file if file is in both stage and unstage status
- pass content of staged version to stdin (PHP CS Fixer could work on stdin, no need to work on drive)

veewee:
The first hint is a bit strange. Both the staged as unstaged changes can succeed or fail. 
The second one will be pretty hard. php-cs-fixer supports stdin, but other tools don't.

Maybe it's a better option to create a new task that can be enabled that checks files that are both in staged as non staged status. This way you can tell GrumPHP that an error can be triggered.
If you want it to just notify when a file exists in both statuses, you can disable the future `blocking` flag.
(see: https://github.com/phpro/grumphp/issues/17)

What do you think of this solution?

(any other ideas @aderuwe ?)

keradus:
Does this behaviour occur on every task that run on files, like phpmd, composer etc?

keradus:
Oh, and then... what about:
- git stash
- run tasks
- git stash pop

?

veewee:
Yes, the behaviour applies to most tasks.
The stash idea is actually pretty nice. 

I expect that the changes only have to be stashed during pre-commit and not at commit-msg level.
Not sure if stash works as expected during pre-commit. I will have to investigate this a bit further.
If it works, it's pretty easy to implement in the actual git-hook.

Thanks for the hint :)

keradus:
Yeah, I have no idea would it work on commit hooks, but it would be extra cool ;)
If not then indeed the stage/unstage task will help as last solution.

[...]

PR #103, opened 2016-02-03
Introducting the ignore_unstaged_changes parameter.

veewee:
Introducting the ignore_unstaged_changes parameter that saved unstaged files to the stash before running the pre-commit tasks.

More information can be found in #100.

**Note that this PR will result in some BC breaks!!!**

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ parameters:
     bin_dir: "./vendor/bin"
     git_dir: "."
     stop_on_failure: false
-    ignore_unstaged_changes: false
+    ignore_unstaged_changes: true
     ascii:
         failed: grumphp-grumpy.txt
         succeeded: grumphp-happy.txt
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ parameters:
     bin_dir: "./vendor/bin"
     git_dir: "."
     stop_on_failure: false
+    ignore_unstaged_changes: false
     ascii:
         failed: grumphp-grumpy.txt
         succeeded: grumphp-happy.txt
--- a/src/GrumPHP/Event/Subscriber/StashUnstagedChangesSubscriber.php
+++ b/src/GrumPHP/Event/Subscriber/StashUnstagedChangesSubscriber.php
@@ -58,10 +58,10 @@ class StashUnstagedChangesSubscriber implements EventSubscriberInterface
     public static function getSubscribedEvents()
     {
         return array(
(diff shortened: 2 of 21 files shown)


src/GrumPHP/Event/Subscriber/StashUnstagedChangesSubscriber.php
```diff
@@ -0,0 +1,117 @@
...
+     * @return void
+     */
+    public function saveStash(RunnerEvent $e)
+    {
+        if (!$this->isStashEnabled($e->getContext())) {
+            return;
+        }
+
+        try {
+            $this->repository->run('stash', array('save', '--quiet', '--keep-index', uniqid('grumphp')));
```
keradus: save with stash name but pop without the name

veewee: It is not a name but a comment. Did not find a way to pop it with a name yet.

keradus: OK, it's not a comment by msg ;P Yep, no one-command way to do it.

veewee:
I've added some additional error handling like uncaught exceptions and internal PHP errors. Normally the stash will always be popped, even when something goes terribly wrong.
Now the only downside is that the changes are stashed as long as the task runs.

veewee:
@keradus , @aderuwe 
I made sure that the stash only runs when there are unstashed (pending) changes.
For me, this feature is complete. Do you have any additional remarks before merging?

README.md
```diff
@@ -65,6 +65,7 @@ parameters:
     bin_dir: "./vendor/bin"
     git_dir: "."
     stop_on_failure: false
+    ignore_unstaged_changes: false
```
keradus: Continuing https://github.com/phpro/grumphp/issues/100#issuecomment-179442140
Since you implement great safety guard with calling stash-pop on `register_shutdown_function` when needed, shouldn't this be true by default?
The only problem for setting this as default was making the changes disappear if sth goes wrong - but now it won't happen.

veewee: I am also thinking about setting the default value to true. 
The only downside is that changes are not available during the commit (when tests run for a long time e.g.) Maybe I should add some additional output like 'Detected unstaged changes - temporarily stash them'. This way it's clear for the user what happens.

keradus: :+1:

aderuwe:
:+1:

PR #111, opened 2016-02-06
Ability to use logger feature of gitonomy/gitlib

llaville:
Use console verbose level 1, 2, or 3

Examples:
```
grumphp run -v
[2016-02-06 13:23:40] RegisteredFiles.INFO: run command: ls-files ""  [] []
[2016-02-06 13:23:41] RegisteredFiles.DEBUG: last command (ls-files) duration: 312.00ms [] []
[2016-02-06 13:23:41] RegisteredFiles.DEBUG: last command (ls-files) return code: 0 [] []
[2016-02-06 13:23:41] RegisteredFiles.DEBUG: last command (ls-files) output: task/Phpparser.php task/sample2.php  [] []
GrumPHP is sniffing your code!
Running task 1/6: Phpcs
Running task 2/6: Phpspec
Running task 3/6: Phpunit
Running task 4/6: Composer
Running task 5/6: SecurityChecker
Running task 6/6: YamlLint
```

```
grumphp git:pre-commit -vv
[2016-02-06 13:26:11] ChangedFiles.INFO: run command: diff "-r -p -m -M --full-index --staged"  [] []
[2016-02-06 13:26:11] ChangedFiles.DEBUG: last command (diff) duration: 327.60ms [] []
[2016-02-06 13:26:11] ChangedFiles.DEBUG: last command (diff) return code: 0 [] []

and more ...
```

```
grumphp git:commit-msg -v "blah bla"
[2016-02-06 13:27:43] ChangedFiles.INFO: run command: diff "-r -p -m -M --full-index --staged"  [] []
[2016-02-06 13:27:43] ChangedFiles.DEBUG: last command (diff) duration: 327.60ms [] []
[2016-02-06 13:27:43] ChangedFiles.DEBUG: last command (diff) return code: 0 [] []

and lot more ...
```

PR #114, opened 2016-02-08
Add logger + IO classes for easy logging

veewee:
Replacement for #111.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
         "symfony/finder": "~2.4|~3.0",
         "symfony/options-resolver": "~2.6|~3.0",
         "symfony/process": "~2.4|~3.0",
-        "symfony/yaml": "~2.4|~3.0"
+        "symfony/yaml": "~2.4|~3.0",
+        "monolog/monolog": "~1.17"
     },
     "require-dev": {
         "composer/composer": "^1.0-alpha11",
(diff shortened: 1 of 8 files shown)


aderuwe:
:+1: Very elegant solution!

llaville:
@veewee Just one question ! How did you change the logger ? From null to, e.g a file stream

veewee:
This happens in the console application:
https://github.com/phpro/grumphp/pull/114/files#diff-b7ec71ee28477693b4fe918ab314d297R191

The null loger is still available but a stream logger is added to the list of handlers in verbose mode.

llaville:
I see that. I just want to know if there is a solution via yaml config file to change the logger behavior.

E.g: replace null logger by another handler. Currently all goes to the console output. when verbose mode is active

llaville:
I've tried to add in my `grumphp.yml` config file these lines  

```

services:      
  grumphp.logger:
    class: Monolog\Logger
    arguments:
      - 'GrumPHP'
      - ['@grumphp.logger.handler.streamhandler']

  grumphp.logger.handler.streamhandler:
    class: Monolog\Handler\StreamHandler
    public: false
    arguments: [ "/tmp/grump.2.log" ]
```

It works fine. But is it the best pratice ?

veewee:
Yes, that is exactly the idea.

You create the handler you want and register it to the grumphp.logger. By default is will use the nullhandler which won't log anything. If you want to write to a file or some other pllace, you can specify your own handler.

When running the CLI command in verbose mode, the output is also added to the stdout of your terminal, so that the information is displayed in the terminal screen.

llaville:
Thanks for this answer !

PR #115, opened 2016-02-09
Make it possible to specify extensions for the git blacklist task.

### LINKED ISSUE #101: Git blacklist feature should not be limited to PHP files
If I'm reading the code correctly, you can only use Git Blacklist with PHP files. It would be very useful to allow this to work with other files types as well.

E.g. Javascript files with `console.log(…)` or `debugger;` statements.

https://github.com/phpro/grumphp/blob/5b6154df71be6bb3922b50600d54ae8d671375e4/src/GrumPHP/Task/Git/Blacklist.php#L54

veewee:
Fixes #101 and specifies an easy way to filter based on extensions.

### DIFF
--- a/doc/tasks/git_blacklist.md
+++ b/doc/tasks/git_blacklist.md
@@ -12,6 +12,7 @@ parameters:
                 - "die("
                 - "var_dump("
                 - "exit;"
+            triggered_by: [php]
 ```
 
 **keywords**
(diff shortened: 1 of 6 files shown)


aderuwe:
The code looks good, as usual. ;)

One issue I have with this, is that this way (if I understand things correctly) it's not possible to specify specific keywords to blacklist *per* extension. (It might be fine to have the word 'console.log' in PHP but probably not in Javascript; that's a contrived example though I guess...)

veewee:
As mentioned in the #101 a specialized blacklist for languages like PHP will be made. Currently there is some work that needs to be done for the php parser (#108).

The idea of this blacklist is to have a general blacklist method that works for all files you want to check.
For PHP specific stuff, the source code needs to be tokenized. Otherwise you don't know the difference between `die()` and `$person->die()`. See the full discussion at #53.

aderuwe:
Righto, I must have missed that. Color me agreeing.

Issue #17, opened 2015-05-26
Add 'blocking' flag to a Task in the TaskRunner


veewee:
In a Task service it must be possible to add a "blocking" tag to the service configuration.
E.g.:

```yml
parameters:
    tasks:
        taskname:
            metadata:
                blocking: false
```

By default all tasks are blocking. A user can overwrite the task service to be non blocking.
When a task is blocking, the errors will be fatal and shown in red.
When a task is not blocking, the error will not be fatal. A yellow error message will be displayed, but the changes will be commited.

The blocking flag of a task should be read when a tasks fails.
The metadata functionallity is already available with the `getTaskMetadata()` method on the GrumPHP object.

More information: https://github.com/phpro/grumphp/blob/master/doc/tasks.md#metadata

aderuwe:
Shouldn't we plan to support this option in `grumphp.yml`?

veewee:
@aderuwe: I changed the description based on one of the previous commits.

aderuwe:
Yup, perfect and what I expected, thanks!

PR #132, opened 2016-03-14
Implements blocking task flag

aaa2000:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | 17

Try to implement #17 but there is a BC break because TaskRunner::run no longer throws FailureException

The feature is already documented in https://github.com/phpro/grumphp/blob/master/doc/tasks.md#metadata


veewee:
It looks pretty great! Thanks!

Some small questions / remarks:
- The `TaskResults` class should be written like the other classes in the `GrumPHP\Collection` namespace, just to be consequent.
- Maybe the `TaskEvents`  should also get the `TaskResults` collection instead of only the messages.
- Why did you choose codes 0 and 99. Maybe we can make them just 1,2, ...
- Maybe there should be a taskresult code for failed but non blocking. This way a possible event listener could validate the results without checking the task metadata.

I don't mind the BC break since we are still on the 0.x version.
What do you think @aderuwe ?

aaa2000:
- I will move `TaskResults` in the `GrumPHP\Collection` namespace, rename `TaskResults ` to `TaskResultCollection` and extends `ArrayCollection`
- I will replace existing parameters of `RunnerFailedEvent` by the `TaskResultCollection`but not modify others events TaskFailedEvent, RunnerEvent...
- I was inspired to https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Tester/Result/TestResult.php, I prefer keep a high code for failed task
- I will create a new taskresult code for failed but non blocking. Have you an idea about the name of the constant? `NONBLOCKING_FAILED`?

Are you okay?

veewee:
About the events. I guess following API's would be ok:

```php
RunnerEvent::__construct(TasksCollection $tasks, ContextInterface $context, TaskResultCollection $taskResults)
```
This means that both the all runner event actions will get a taskresults object. During the 'tasks.run' event, the statuscode could be `NO_TASKS` as in:
https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Tester/Result/TestResults.php


About the status codes. In Behat the tasks are returning TaskResults. Maybe that is something we can implement in a next phase. It seems better then current idea of just throwing exceptions when a task fails and return nothing when a tasks passes. This way we can also think about a `WARNINGS` status code or something.

About the constant name you proposed. I haven't come up with something better, so I guess one of `FAILED_NONBLOKCING` or `NONBLOCKING_FAILED` is ok.

aaa2000:
PR updated. 

But, it miss this
>During the 'tasks.run' event, the statuscode could be NO_TASKS

Would you rather say `During the 'grumphp.runner.run' event` ?

And what should be the behaviour of stop_on_failure with a non blocking task ? I stop only if it is a blocking task, I have to add a test on this behaviour.

Indeed, it would be better that the tasks return TaskResults.

src/GrumPHP/Runner/TaskRunner.php
```diff
@@ -113,4 +120,14 @@ public function run(ContextInterface $context)
 
         return $taskResuls;
     }
+
+    /**
+     * @param \GrumPHP\Task\TaskInterface $task
+     * @return bool
+     */
+    private function isBlockingTask(TaskInterface $task)
```
veewee: Maybe this method could be moved to the configuration class?

veewee:
Wow, this PR is starting to look pretty great!
Indeed I meant the `grumphp.runner.run` event. Currently the TaskResultCollection will pass if no task ran. By adding the NO_TASKS code, you can say that it passed but no task ran yet. Similar to the Behat class I mentioned earlier.
I think it's best that the non blocking tasks should not cause GrumPHP to fail during `stop_on_failure`. The messages are just warnings, so GrumPHP should continue running the next tasks.

When this one is merged, I will try to make some time to let the tasks return a TaskResult. We are breaking BC anyways :)

src/GrumPHP/Runner/TaskRunner.php
aaa2000: Done

aaa2000:
PR updated.

veewee:
It looks great. I will try to test it out somewhere this week. Thanks for the effort!

veewee:
I just took a deeper look at the code and noticed that the `TaskRunnerHelper` is displaying the first error it encounters and then stops displaying the errors. In previous versions all errors are displayed.

I was thinking about making the helper a bit easier by adding 2 new functions to the `TaskResultCollection`:

```php
public function filterByStatusCode() : TaskResultCollection;
public function getAllMessages() : array|string[];
```

This way you can make the `TaskRunnerHelper` a lot easier. I was thinking about something like this:
```php
public function newRun(OutputInterface $output, ContextInterface $context, $skipSuccessOutput = false)
    {
        // Make sure to add some default event listeners before running.
        $this->registerEventListeners($output, $context);

        $taskResults = $this->taskRunner->run($context);

        $warnings = $taskResults->filterByStatusCode(TaskResult::NONBLOCKING_FAILED);
        if ($warnings->count()) {
            $this->returnWarningMessages($output, $warnings->getAllMessages());
        }

        if (!$taskResults->isPassed()) {
            $failed = $taskResults->filterByStatusCode(TaskResult::FAILED);
            return $this->returnErrorMessage($output, $failed->getAllMessages());
        }

        if ($skipSuccessOutput) {
            return self::CODE_SUCCESS;
        }

        return $this->returnSuccessMessage($output);
    }
    
```


Maybe the filterByStatusCode could also be replaced by some other method. E.g blocking(), nonBlocking(), ...
What do you think about this solution?

aaa2000:
Indeed, I had not seen the regression. The solution seems to me well. 
I prefer the warnings above the errors because I first correct errors and then maybe the warnings. I don't want to scroll the terminal to see the errors.

And If the warnings is always display and the warnings are **above** errors. why not ?

```
public function run(OutputInterface $output, ContextInterface $context, $skipSuccessOutput = false)
{
        $this->registerEventListeners($output, $context);
        $taskResults = $this->taskRunner->run($context);
        $warnings = $taskResults->filterByStatusCode(TaskResult::NONBLOCKING_FAILED);
        $this->returnWarningMessages($output, $warnings->getAllMessages());

        if (!$taskResults->isPassed()) {
            $failed = $taskResults->filterByStatusCode(TaskResult::FAILED);
            return $this->returnErrorMessage($output, $failed->getAllMessages());
        }

        if ($skipSuccessOutput) {
            return self::CODE_SUCCESS;
        }

        return $this->returnSuccessMessage($output);
}
```

veewee:
That is also ok for me!

aaa2000:
Fix done but I have replaced `!$taskResults->isPassed()` by `$taskResults->isFailed()`. I didn't want change the isPassed method in order to it return true when there are `TaskResult::NONBLOCKING_FAILED`.

aaa2000:
When the grumphp.runner.failed should be trigger ? Only if a blocking task has failed ? Or a non blocking task too ?

veewee:
I think it's best to only trigger the failed event when a task has really failed. So a non-blocking task will always pass. This way it follows the same logic as the taskrunner: a task can only fail when it is blocking.

veewee:
@aaa2000: For me this one can be merged. Is their something you would like to change before merging?
@aderuwe: Can you take an extra look at this PR? It's a rather big one.

veewee:
I allready started with returning `TaskResult` objects in a task. This is still WIP but gives you an idea:
https://github.com/veewee/grumphp/commit/d4a872ad1629764b415e633ade473a85ca81d0fd

aaa2000:
For me too, it can be merged. Do you want that I squash the commits ?

I look veewee@d4a872a and I note that you use the named constructors `TaskResult::createFailed` and `TaskResult::createPassed` that I have reverted in commit 617606ef04e25e704929b61596c9888e401121d5

veewee:
For me it doesn't need to get squashed. I will be adding the named constructors again in a later version, since it improves readability.
Thanks for your work!

aaa2000:
Yes, It improves readability and I hesitated to remove its. But adding a new status code involve always to add a new method in the class and I prefered use a ternary operator in https://github.com/phpro/grumphp/pull/132/files#diff-2ad714d600b1981df304b68889ab9aefR94

Thank you for the review and for the tool !

veewee:
Thanks again for your work @aaa2000 !

Issue #145, opened 2016-05-17
Renamed: git:init during composer install uses wrong path


kvdnberg:
When (re-)installing GrumPHP using composer install or update inside vagrant, the pre-commit hook looks like this: 
```vagrant ssh --command "(exec '/vagrant/bin/grumphp' 'git:pre-commit' '--skip-success-output')"```

This gives an exception running git outside of vagrant: 
```
 [Gitonomy\Git\Exception\LogicException]
  Can't create a working copy on a bare repository
```
When reinitializing using ```bin/grumphp git:init``` inside vagrant, the line is changed to:
```vagrant ssh --command "(exec 'bin/grumphp' 'git:pre-commit' '--skip-success-output')"```

**Comment by veewee** [comment]
The `bin/grumphp` command is the correct one. When exactly did you get the `/vagrant` in your git hook?
I'm not fond of adding new parameters to the `grumphp.yml` specific for the vagrant hook.
It should be universal so that new hook presets can be added without altering the code. This is to keep the code as general as possible and independent of the file system or used tools.

That is why we decided that the line in .basrc is sufficient. You could add this file to the vagrant provision so that every user has the same .bashrc file. We put it in our boxes by default, so that's not a big problem. 

So in my oppinion the PR is not a valid fix. The issue that needs to be fixed is the hook with the `/vagrant` path and maybe a more clear section in the documentation.

**Comment by kvdnberg** [comment]
Thanks for your reply. I understand your objection to the fix.

The problem is still that it would seem that the bashrc line doesn't work for ```vagrant ssh --command```

```
$ vagrant ssh --command "pwd"
/home/vagrant
Connection to 127.0.0.1 closed.
$ vagrant ssh
vagrant@boxname:/vagrant #$ pwd
/vagrant
```

kvdnberg:
Updated the ticket to reflect the /vagrant bug during composer install/update

veewee:
Strange .. I'll have to dive deeper in the issue.
Thanks for reporting!

veewee:
I've started debugging the git:init command during both composer install and when the command runs stand-alone.
After some digging, I found out that composer is adding the absolute path to the bin dir in the PATH environemnt variable during the dispatching of events:
https://github.com/composer/composer/blob/1.1/src/Composer/EventDispatcher/EventDispatcher.php#L147-L160

Since we are using the Symfony `ExecutableFinder`:
[...]

PR #146, opened 2016-05-17
Fixes the vagrant environment by adding vagrant dir config var

kvdnberg:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes? I don't have phpspec running, but updated the tests
| Documented?   | yes
| Fixed tickets | #145 

This PR solves the issue described in #145, by adding a vagrant_dir to the config parameters, where the vagrant dir of the project can be set, which is then used in the git hooks to cd to before executing GrumPHP.

### DIFF
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -48,11 +48,14 @@ GrumPHP comes with following presets:
 - `vagrant`: All checks will run in your vagrant box.
 
 
-*Note:* When using the vagrant preset, make sure the working directory of the vagrant shell is located at your remote project path:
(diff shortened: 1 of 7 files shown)


veewee:
As mentioned in the issue #145: I am not fond of adding vagrant specific configuration in the grumphp.yml file. The code should remain as general as possible and independent of the choosen git hooks preset.

PR #155, opened 2016-06-18
Add atoum task

jhuet:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | no

[Atoum](http://atoum.org/) is unit test framework that could totally be run in a grumphp task. This is an implementation of its usage containing specs and doc.

### DIFF
--- a/doc/tasks/atoum.md
+++ b/doc/tasks/atoum.md
@@ -1,0 +1,74 @@
+# Atoum
+
+The Atoum task will run your unit tests.
+It lives under the `atoum` namespace and has the following configurable parameters:
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        atoum:
+            config_file: ~
+            bootstrap_file: ~
+            directories: ~
+            files: ~
+            namespaces: ~
+            methods: ~
+            tags: ~
+```
+
+**config_file**
+
+*Default: null*
+
+The path to your configuration file. If you name your configuration file .atoum.php, atoum will load it automatically if this file is located in the current directory. The config_file parameter is optional in this case.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/configuration_bootstraping.html#configuration-file)
+
+**bootstrap_file**
+
+*Default: null*
+
+The path to your bootstrap file if you need any.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/configuration_bootstraping.html#bootstrap-file)
+
+**directories**
+
+*Default: null*
+
+If you want to limit the execution of the unit tests to certain directories, list them here.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/running_tests.html#by-folders)
+
+**files**
+
+*Default: null*
+
+If you want to limit the execution of the unit tests to certain files, list them here.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/running_tests.html#by-files)
+
+**namespaces**
+
+*Default: null*
+
+If you want to limit the execution of the unit tests to certain namespaces, list them here.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/running_tests.html#by-namespace)
+
+**methods**
+
+*Default: null*
+
+If you want to limit the execution of the unit tests to certain methods or classes, list them here.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/running_tests.html#a-class-or-a-method)
+
+**tags**
+
+*Default: null*
+
+If you want to limit the execution of the unit tests to certain tags, list them here.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/running_tests.html#tags)
--- a/doc/tasks/atoum.md
+++ b/doc/tasks/atoum.md
@@ -42,7 +42,7 @@ The path to your bootstrap file if you need any.
 
 **directories**
 
-*Default: array()*
+*Default: []*
 
 If you want to limit the execution of the unit tests to certain directories, list them here.
 
@@ -50,7 +50,7 @@ If you want to limit the execution of the unit tests to certain directories, lis
 
 **files**
 
-*Default: array()*
+*Default: []*
 
 If you want to limit the execution of the unit tests to certain files, list them here.
 
@@ -58,7 +58,7 @@ If you want to limit the execution of the unit tests to certain files, list them
 
 **namespaces**
 
-*Default: array()*
+*Default: []*
 
 If you want to limit the execution of the unit tests to certain namespaces, list them here.
 
@@ -66,7 +66,7 @@ If you want to limit the execution of the unit tests to certain namespaces, list
 
 **methods**
 
-*Default: array()*
+*Default: []*
 
 If you want to limit the execution of the unit tests to certain methods or classes, list them here.
 
@@ -74,7 +74,7 @@ If you want to limit the execution of the unit tests to certain methods or class
 
 **tags**
 
-*Default: array()*
+*Default: []*
 
 If you want to limit the execution of the unit tests to certain tags, list them here.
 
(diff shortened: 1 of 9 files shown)


src/GrumPHP/Task/Atoum.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public function getName()
+    {
+        return 'atoum';
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
```
veewee: Shouldn't the filters be arrays instead of strings? In that case, you could set an empty array as default value instead.

src/GrumPHP/Task/Atoum.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $files = $context->getFiles()->name('*.php');
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $config = $this->getConfiguration();
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('atoum');
+        if ($config['config_file']) {
```
veewee: For this kind of if structures, there is an `addOptionalArgument` method on the `ProcessArgumentsCollection` class.

src/GrumPHP/Task/Atoum.php
```diff
@@ -0,0 +1,103 @@
...
+        if ($config['files']) {
+            $arguments->addArgumentArrayWithSeparatedValue('--files', array($config['files']));
+        }
+        if ($config['namespaces']) {
+            $arguments->addArgumentArrayWithSeparatedValue('--namespaces', array($config['namespaces']));
+        }
+        if ($config['methods']) {
+            $arguments->addArgumentArrayWithSeparatedValue('--methods', array($config['methods']));
+        }
+        if ($config['tags']) {
```
veewee: If the tags section is an array in the configuration, you can reomove these ifs and directly run the `addArgumentArrayWithSeparatedValue()` method.

veewee:
Hi @jhuet,

I am not familiar with `atoum` so I am not sure how the cli configuration works exactly. 
I've added some small remarks to the code. 

Thanks for the new task!

src/GrumPHP/Task/Atoum.php
jhuet: The thing is that the actual parameter syntax for atoum is this one : `atoum --bootstrap-file tests/units/bootstrap.php` but if i use `addOptionalArgument('--bootstrap-file %s', $config['bootstrap_file'])`, the `Process` will escape the whole argument making it be like : `'atoum' '--bootstrap-file tests/units/bootstrap.php'` instead of `'atoum' '--bootstrap-file' 'tests/units/bootstrap.php'`. I'm not sure how to do this without using `addArgumentArrayWithSeparatedValue()` which is why i kinda tricked around using it.

src/GrumPHP/Task/Atoum.php
jhuet: I have a similar problem here. For multiple arguments, the syntax is : `atoum --directories dir1 dir2`. If i use `addOptionalArgument` or like now `addArgumentArrayWithSeparatedValue` by "faking" a string, all directories are escaped in 1 parameter : `'atoum' '--directories' 'dir1 dir2'`. If i use `addArgumentArrayWithSeparatedValue` with an actual array, escaping is ok but the parameter name is duplicated : `'atoum' '--directories' 'dir1' '--directories' 'dir2'`. If you have any advice on how to do this, it'd be very much welcome :)

src/GrumPHP/Task/Atoum.php
jhuet: See my comments below.

veewee:
Hi @jhuet,

That way of adding arguments is pretty common.
Maybe you can just add 2 new methods to the `ProcessArgumentsCollection`?
I was thinking about something like:

`addOptionalSeparatedArgument()`

This method will look like `addOptionalArgument` but will add 2 arguments instead of a sprintf if the value is available.


`addSeparatedArgumentArray()`

This method will look like `addArgumentArray` but will set the key once and multiple value arguments. Note that when the array is empty, the key of the argument should not be set.

This way you can make the task more readable and improve the types of the configurable options.

jhuet:
Alright @veewee, i'll find some time to add these methods. For some reason i didn't realize this class was part of grumphp, i thought it was symfony's :)

veewee:
@jhuet Are you planning to finish this PR any time soon? Thanks!

jhuet:
@veewee, sorry i've been busy and i'm now on vacations. I'll try to finish this during the week-end of the 17th of september.

veewee:
Have a great vacation! ;)

jhuet:
Hey @veewee, i finally pushed an updated version of this task using the 2 new methods `addOptionalSeparatedArgument()` and `addSeparatedArgumentArray()` that i added in `ProcessArgumentsCollection`.

I also rebased all of this from the master branch.

src/GrumPHP/Collection/ProcessArgumentsCollection.php
```diff
@@ -65,6 +65,18 @@ public function addOptionalCommaSeparatedArgument($argument, array $values, $del
 
     /**
      * @param string $argument
+     * @param string $value
+     */
+    public function addOptionalSeparatedArgument($argument, $value = null)
```
veewee: This method was already implemented in the `addOptionalArgumentWithSeparatedValue` method:
https://github.com/jhuet/grumphp/blob/8ff4bb0ff833d00ad0ea010cc164961832173df7/src/GrumPHP/Collection/ProcessArgumentsCollection.php#L42

Can you remove the `addOptionalSeparatedArgument` method and use the one above instead??

src/GrumPHP/Collection/ProcessArgumentsCollection.php
```diff
@@ -90,6 +102,22 @@ public function addArgumentArrayWithSeparatedValue($argument, array $values)
 
     /**
      * @param string $argument
+     * @param array  $values
+     */
+    public function addSeparatedArgumentArray($argument, array $values)
```
veewee: Can you add phpspec tests for this method?

src/GrumPHP/Task/Atoum.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return ($context instanceof GitPreCommitContext || $context instanceof RunContext);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $files = $context->getFiles()->name('*.php');
```
veewee: I haven't worked with atoum myself. Is it possible to add the `$files` parameter to the process or will it always run all of its tests?

doc/tasks/atoum.md
```diff
@@ -0,0 +1,81 @@
...
+
+*Default: null*
+
+The path to your bootstrap file if you need any.
+
+[See atoum documentation](http://docs.atoum.org/en/latest/configuration_bootstraping.html#bootstrap-file)
+
+**directories**
+
+*Default: array()*
```
veewee: In other documentation files we use `[]` instead of `array()`. Can you update the documentation so it looks the same like the other tasks?

src/GrumPHP/Task/Atoum.php
jhuet: Yes, Atoum allows to run tests for specific files but `$files` would have to be the test files. AFAIK, if a file that is unit tested is given to Atoum command line, it won't know which tests to execute. So i'm not sure it would be very usefull to use.

veewee: That's what I thought. Great! :)

veewee:
Thanks for your work!

PR #156, opened 2016-06-28
Upgrade phpcsfixer task to stick to v2.0

tifabien:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes/no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | 

As the PHPCsFixer project released a first v2.0.0-alpha version, I propose this PR in order to be ready when the final v2.0 will be released.

We should wait for the final v2.0 before the merge of this PR,but I prefer to submit it in a proactive way.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "phpspec/phpspec": "~2.5",
         "phpspec/prophecy": "~1.6",
         "squizlabs/php_codesniffer": "~2.3",
-        "friendsofphp/php-cs-fixer": "~1",
+        "friendsofphp/php-cs-fixer": "~2.0",
         "phpunit/phpunit": "^4.8",
         "sensiolabs/security-checker": "^3.0"
     },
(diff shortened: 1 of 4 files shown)


veewee:
Hi @tifabien,

Thanks for your work. Let me know when 2.0 gets released so that I can rerun the CI checks.
Besides that, I notice that you only added properties. Does this mean that it works with both 2.0 and 1.0?
I think we should support both versions for a pretty long period.

Are ther eattributes in 1.0 that don't work in 2.0 anymore?
Maybe we should add the version of php-cs-fixer in the configuration part of the task documentation file?
It might also be an option to specify a php-cs-fixer version in the task configuration and validate the configuration input with the OptionsResolver through callback functions in setDefault() or setNormalizer()?

tifabien:
Hi @veewee 

You can have a look at the [UPGRADE.md](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md) guide in order to see what changed and how to migrate. Some options has been removed, other renamed,.. So if we want to support both versions this PR needs to be adapted. I'll have a look.

veewee:
This PR is closed in favour of #163.

PR #163, opened 2016-07-12
Added php cs fixer v2

Big-Shark:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | no

### DIFF
--- a/resources/config/tasks.yml
+++ b/resources/config/tasks.yml
@@ -121,6 +121,15 @@ services:
         tags:
           - {name: grumphp.task, config: phpcsfixer}
 
+    task.phpcsfixer2:
+        class: GrumPHP\Task\PhpCsFixerV2
+        arguments:
+          - '@config'
+          - '@process_builder'
+          - '@formatter.php_cs_fixer'
+        tags:
+          - {name: grumphp.task, config: phpcsfixer2}
+
     task.phpspec:
         class: GrumPHP\Task\Phpspec
         arguments:
--- a/spec/GrumPHP/Task/PhpCsFixerV2Spec.php
+++ b/spec/GrumPHP/Task/PhpCsFixerV2Spec.php
@@ -1,0 +1,157 @@
+<?php
+
+namespace spec\GrumPHP\Task;
+
+use GrumPHP\Collection\FilesCollection;
+use GrumPHP\Collection\ProcessArgumentsCollection;
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Formatter\PhpCsFixerFormatter;
+use GrumPHP\Process\ProcessBuilder;
+use GrumPHP\Runner\TaskResult;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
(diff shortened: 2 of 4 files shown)


Big-Shark:
```
tasks:
        phpcsfixer2:
            allow-risky: ~
            cache-file: ~
            config: ~
            rules: ~
            using-cache: ~
            path-mode: ~
            verbose: true
```

Big-Shark:
@veewee Hi, please check this, and write your decision.

veewee:
@Big-Shark: 
Thanks for the PR! Did you also take a look at #156? 
This PR needs some documentation and an updated composer.json file. 
We are also enforcing snake case on the task options. This means that e.g. `allow-risky` becomes `allow_risky`.
I am also thinking about removing the inheritance of the php-cs-fixer 1 task.
This way, we are not blocked by the first task if the second version gets an update.
Maybe the shared code can be placed in another class. 

@tifabien: 
can you also take a look at this task? Maybe it's better to make 2 separate tasks, one for php-cs-fixer 1 and one for php-cs-fixer 2. What is your oppinion on this?

tifabien:
@veewee ,
I just had a quick look at this PR. I'm ok with your comments regarding snake case on task options, docs addition, composer.json updates... 
If you want to make 2 separated tasks for phpcsfixer I'm also OK. 
Btw your suggestion in the PR [#156](https://github.com/phpro/grumphp/pull/156#issuecomment-229383344) seems also interesting. This configuration check depending the version config value is in my opinion a good idea. This PR is a more straightforward solution.
Now it's up to you to decide whether you want one specific class per version or one Class with multiple version support :)

veewee:
I think a second task is indeed a better solution. Let's work on this PR in favour of #156.

Big-Shark:
@veewee Updated code style, and created a trait, check please.

Big-Shark:
@veewee 5.3? Really?

veewee:
Currently: Yes really :) 
We should drop support for 5.3, 5.4 and even 5.5 but there are still some legacy apps that run on vagrant boxes with php 5.3, so I would like it to be as BC as possible with old PHP versions.

Besides to that, I really don't like traits. 
It seems better to create a class like `PhpCsFixerRunner` which has it's own dependencies (formatter, processbuilder, ...) and inject this service into the php-cs-fixer tasks. 

This way, the code is easy testable and it's context is fixed.

veewee:
@Big-Shark Are you planning to finish this PR any time soon? Thanks!

Big-Shark:
@veewee Recently, I have little time, and I would be happy if you could refactor this PR. Thanks.

veewee:
Thanks for your feedback! I'll try to make some time for this.

veewee:
@tifabien, @Big-Shark 
I've merged a first version of the phpcsfixer 2 task.
It's open for improvements.

Big-Shark:
@veewee thx

tifabien:
@veewee ,
I'll try the new v0.9.2 + Cs fixer 2.0 and let you know if I find something :)

Issue #166, opened 2016-07-29
phpunit task group option


Kanti:
| Q               | A
| --------------- | ---
| Branch          | /
| Bug?            | no
| New feature?    | no
| Question?       | yes
| Documentation?  | no
| Related tickets | /

**About:**
Could we get an option for the phpunit task that could specify the groups the would like to run?
it would be very helpful.

**something like this:**
```yaml
# grumphp.yml
    phpunit:
      config_file: tests/phpunit.xml
      groups: [small, fast]
      metadata:
        priority: 200
```

veewee:
Great idea! Feel free to implement.

PR #167, opened 2016-07-31
[FEATURE] add functionality to phpunit task --group

Kanti:
| Q             | A
| ------------- | ---
| Branch        | /
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #166 

**About:**
Could we get an option for the phpunit task that could specify the groups the would like to run?
it would be very helpful.

**something like this:**
```yaml
# grumphp.yml
    phpunit:
      config_file: tests/phpunit.xml
      groups: [small, fast]
      metadata:
        priority: 200
```

### DIFF
--- a/src/GrumPHP/Task/Phpunit.php
+++ b/src/GrumPHP/Task/Phpunit.php
@@ -29,9 +29,11 @@ class Phpunit extends AbstractExternalTask
         $resolver = new OptionsResolver();
         $resolver->setDefaults(array(
             'config_file' => null,
+            'group' => null,
         ));
 
         $resolver->addAllowedTypes('config_file', array('null', 'string'));
+        $resolver->addAllowedTypes('group', array('null', 'array'));
 
         return $resolver;
     }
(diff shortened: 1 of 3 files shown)


src/GrumPHP/Task/Phpunit.php
```diff
@@ -29,9 +29,11 @@ public function getConfigurableOptions()
         $resolver = new OptionsResolver();
         $resolver->setDefaults(array(
             'config_file' => null,
+            'group' => null,
         ));
 
         $resolver->addAllowedTypes('config_file', array('null', 'string'));
+        $resolver->addAllowedTypes('group', array('null', 'array'));
```
veewee: Isn't it better to set the default value to an empty array and only allow array types?

veewee:
Thanks for the PR @Kanti!
I've added some small remarks. Can you take a look at them?
Also the spec file for the phpunit test should be altered before I can merge this change.

Kanti:
Do I need to change more in the PhpunitSpec.php file?

veewee:
Nope, that's all. Can you also update the default values in the documentation?

veewee:
Thanks!

PR #172, opened 2016-08-16
Escape arguments we're stripping out

Mezzle:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 

On windows, the arguments are escaped with a `"` rather than a `'`.  This fix uses the argument escaper to escape the arguments we're stripping out correctly, so that it detects and strips them out correctly.


veewee:
Thanks for the PR!

PR #174, opened 2016-08-30
[Tasks] added PHPMD to tasks

Devolicious:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | 

This PR implements the usage for PHPMD! Standard it will use the rulesets ```cleancode, codesize,naming```. I also implemented the triggered_by options with the default being ```php```. As format I hardcoded it to ```text``` because it provides a clean overview when using the pre-commit hook. Also, it does not generate output when there are no issues which means that the process has exit 0 code and will result in success.


veewee:
Thanks for the PR! 

There is still a little work that needs to be done before we can merge this code:

It looks like there is still a little styling issue of which the CI is complaining.

The composer file will need an extra entry which suggests the installation of the phpmd tool.

Can you also add some documentation on how to configure this command in the doc folder?
Files that need to be altered/created:
- doc/tasks/phpmd.md: A file on how to use the command, formatted as the other task documentations.
- doc/tasks.md: A link to the task configuration and an entry in the example tasks code.
- readme.md: An entry in the example tasks code and the composer suggestion.

Devolicious:
@veewee: done :)

veewee:
Thanks for your work!

Devolicious:
Thanks for the quick response! :)

Issue #165, opened 2016-07-29
Add php linter task


sc0rp10:
| Q               | A
| --------------- | ---
| Branch          | master for features and deprecations
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

It's be very useful if you add https://github.com/JakubOnderka/PHP-Parallel-Lint task for fast parallel files linting

veewee:
It looks like a great addition. Can it also take changed files as input parameter?
Feel free to implement.

sc0rp10:
```
sc0rp10 @ work:/tmp $ ./vendor/bin/parallel-lint foo.php bar.php
PHP 7.0.9 | 10 parallel jobs
.                                                            2/2 (100 %)


Checked 2 files in 0 seconds
No syntax error found
```

Yes, It is.

veewee:
Added in PR #182 .

PR #182, opened 2016-09-12
Add a parallel php lint task

bertptrs:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | 165

This pull request adds a task to lint all php files to GrumPHP. It uses the parallel php linter as suggested in #165.

Please let me know if I missed some crucial documentation or made some weird errors.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "sensiolabs/security-checker": "Lets GrumPHP be sure that there are no known security issues.",
         "codeception/codeception": "Lets GrumPHP run your project's full stack tests",
         "phpmd/phpmd": "Lets GrumPHP sort out the mess in your code",
-		"jakub-onderka/php-parallel-lint": "Lets GrumPHP quickly lint your entire code base."
+        "jakub-onderka/php-parallel-lint": "Lets GrumPHP quickly lint your entire code base."
     },
     "authors": [
         {
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "squizlabs/php_codesniffer": "~2.3",
         "friendsofphp/php-cs-fixer": "~1|~2",
         "phpunit/phpunit": "^4.8",
-        "sensiolabs/security-checker": "^3.0"
+        "sensiolabs/security-checker": "^3.0",
+        "jakub-onderka/php-parallel-lint": "^0.9.2"
     },
     "suggest": {
         "behat/behat": "Lets GrumPHP validate your project features.",
@@ -41,7 +42,8 @@
         "roave/security-advisories": "Lets GrumPHP be sure that there are no known security issues.",
         "sensiolabs/security-checker": "Lets GrumPHP be sure that there are no known security issues.",
         "codeception/codeception": "Lets GrumPHP run your project's full stack tests",
-        "phpmd/phpmd": "Lets GrumPHP sort out the mess in your code"
+        "phpmd/phpmd": "Lets GrumPHP sort out the mess in your code",
+		"jakub-onderka/php-parallel-lint": "Lets GrumPHP quickly lint your entire code base."
     },
     "authors": [
         {
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -24,6 +24,7 @@ parameters:
         phpcsfixer: ~
         phpcsfixer2: ~
         phpmd: ~
+        phplint: ~
         phpspec: ~
         phpunit: ~
         robo: ~
@@ -54,6 +55,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 - [PHP-CS-Fixer](tasks/php_cs_fixer.md)
 - [PHP-CS-Fixer 2](tasks/php_cs_fixer2.md)
 - [PhpMd](tasks/phpmd.md)
+- [PHPLint](tasks/phplint.md)
 - [Phpspec](tasks/phpspec.md)
 - [Phpunit](tasks/phpunit.md)
 - [Robo](tasks/robo.md)
@@ -64,7 +66,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 
 ## Metadata
 
-Every task has a pre-defined `metadata` key on which application specific options can be configured. 
+Every task has a pre-defined `metadata` key on which application specific options can be configured.
 For example:
 
 ```yaml
@@ -89,7 +91,7 @@ The higher the priority, the sooner the task will be executed.
 
 *Default: true*
 
-This option can be used to make a failing task non-blocking. 
+This option can be used to make a failing task non-blocking.
 By default all tasks will be marked as blocking.
 When a task is non-blocking, the errors will be displayed but the tests will pass.
 
--- a/doc/tasks/phplint.md
+++ b/doc/tasks/phplint.md
@@ -9,7 +9,7 @@ parameters:
         phplint:
             exclude: []
             jobs: ~
-            triggered_by: ['php']
+            triggered_by: ['php', 'phtml', 'php3', 'php4', 'php5']
 ```
 **exclude**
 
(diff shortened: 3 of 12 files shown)


composer.json
```diff
@@ -41,7 +42,8 @@
         "roave/security-advisories": "Lets GrumPHP be sure that there are no known security issues.",
         "sensiolabs/security-checker": "Lets GrumPHP be sure that there are no known security issues.",
         "codeception/codeception": "Lets GrumPHP run your project's full stack tests",
-        "phpmd/phpmd": "Lets GrumPHP sort out the mess in your code"
+        "phpmd/phpmd": "Lets GrumPHP sort out the mess in your code",
+		"jakub-onderka/php-parallel-lint": "Lets GrumPHP quickly lint your entire code base."
```
veewee: It looks like you've added too many spaces.

doc/tasks.md
```diff
@@ -24,6 +24,7 @@ parameters:
         phpcsfixer: ~
         phpcsfixer2: ~
         phpmd: ~
+        phplint: ~
```
veewee: Can you also add this task in README.md?

doc/tasks/phplint.md
```diff
@@ -0,0 +1,35 @@
...
+    tasks:
+        phplint:
+            to_check: [ '.' ]
+            exclude: []
+            jobs: ~
+```
+
+**to_check**
+
+*Default: project root*
```
veewee: In the other tasks, the default config is pasted here. You can clearify that it is loading the whole project in the text below.

doc/tasks/phplint.md
```diff
@@ -0,0 +1,35 @@
...
+
+The directories to check. By default, it checks the entire repository,
+but you can fine-tune this as needed.
+
+**exclude**
+
+*Default: array()*
```
veewee: In the other tasks we are using [] to specify that an array is used.

grumphp.yml.dist
```diff
@@ -20,3 +20,6 @@ parameters:
       yamllint:
         ignore_patterns:
           - "#test/(.*).yml#"
+      phplint:
+        exclude:
+          - vendor
```
veewee: Maybe it's better to run the task on the context files and filter on extensions (php, phtml, ...).
This way you won't need to exclude all the folders you specified in the .gitignore file.
When you only change 1 file, you won't have to validate ALL the files in the project.

So I guess it's better to drop the `to_check` parameter and use the context files instead. You can exclude some patterns by filtering the context files.

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,74 @@
...
+use GrumPHP\Runner\TaskResultInterface;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * PHP parallel lint task.
+ */
+class PHPLint extends AbstractExternalTask
```
veewee: Can you also add a phpspec test for this class?

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
+
+        $args = $this->processBuilder->createArgumentsForCommand('parallel-lint');
+        $args->add('--no-colors');
+        $args->addOptionalArgument('-j', $config['jobs']);
```
veewee: You are not setting the amount of jobs with this command. The command should be something like `-j %s`.

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,74 @@
...
+        $args->addArgumentArrayWithSeparatedValue('--exclude', $config['exclude']);
+        $args->addArgumentArray('%s', $config['to_check']);
```
veewee: As mentioned above, lets use `$context->getFiles()` instead of the to_check argument.

veewee:
Thanks for your PR. It looks pretty good, but I've added some remarks.
We also need some phpspec tests for this new task to make sure it works exactly as it should.

bertptrs:
I think that covers all your suggestions. Note that this is the first time I've used phpspec I'm not 100% sure of what I did. I mostly copied the tests from PHPMD as they are rather similar.

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return $context instanceof RunContext || $context instanceof GitPreCommitContext;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $fileNames = array_map(function (\SplFileInfo $file) {
```
veewee: You don't have to use the array_map method here., just set
```
$files = $context->getFiles()->extensions(array('php'))
```

You could add an additional option `triggered_by` which defaults to ['php'] so that it can be configurable.

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,79 @@
...
+        $config = $this->getConfiguration();
+
+        $args = $this->processBuilder->createArgumentsForCommand('parallel-lint');
+        $args->add('--no-colors');
+        if (!empty($config['jobs'])) {
+            $args->add('-j');
+            $args->add($config['jobs']);
+        }
+        $args->addArgumentArrayWithSeparatedValue('--exclude', $config['exclude']);
+        $args->addArgumentArray('%s', $fileNames);
```
veewee: You can use:
```
$arguments->addFiles($files);
```

This will add the pathname as an argument so that you wont have to use the array_map() method.

src/GrumPHP/Task/PHPLint.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        $fileNames = array_map(function (\SplFileInfo $file) {
+            return $file->getPathname();
+        }, iterator_to_array($context->getFiles()->extensions(array('php'))));
+
+        $config = $this->getConfiguration();
+
+        $args = $this->processBuilder->createArgumentsForCommand('parallel-lint');
+        $args->add('--no-colors');
+        if (!empty($config['jobs'])) {
```
veewee: Is there a reason why you did remove the `$args->addOptionalArgument()` method?
It could look like this as far as I can tell from the documentation of the lint programm:

```
$args->addOptionalArgument('-j %s', $config['jobs'])
```

veewee:
Hi, the phpspec task looks good. Thanks for your fast response!
I've added some small improvements to the code in the task.

src/GrumPHP/Task/PHPLint.php
bertptrs: The addOptionalArgument creates one argument, effectively sending the argument as `"-j 10"` rather than 2 arguments. The program requires it to be two seperate arguments.

bertptrs:
Those should be implemented as well.

veewee:
Can you check this checkbox:
https://github.com/blog/2247-improving-collaboration-with-forks

I've made same small changes I want to push into the branch.

veewee:
I've made some small changes:

- The  `-e`  option is removed since the extensions filter will do this. I've also added the default extension filters to the `triggered_by` option.
- I've changed the name of the task class to make it universal over the tasks.
- I've added a `addOptionalArgumentWithSeparatedValue()` so that we can get rid of the if statement.

Now it looks good to me. Can you verify one last time before I merge?

Thanks for your effort!

bertptrs:
Looks fine to me. Go ahead.

PR #186, opened 2016-09-20
[FEATURE] Git blacklist formatter

Kanti:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | -
| Fixed tickets | -

Nicer Git Blacklist Task message for very long lines.
Example:
image

### DIFF
--- a/src/GrumPHP/Formatter/GitBlacklistFormatter.php
+++ b/src/GrumPHP/Formatter/GitBlacklistFormatter.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace GrumPHP\Formatter;
+
+use Symfony\Component\Process\Process;
+
+/**
+ * Class GitBlacklistFormatter
+ *
+ * @package GrumPHP\Formatter
+ */
+class GitBlacklistFormatter implements ProcessFormatterInterface
+{
+    /**
+     * @param Process $process
+     *
+     * @return string
+     */
+    public function format(Process $process)
+    {
+        $output = $process->getOutput();
+        if (!$output) {
+            return $process->getErrorOutput();
+        }
+
+        return $this->formatOutput($output);
+    }
+
+    /**
+     * @param string $output
+     * @return string
+     */
+    protected function formatOutput($output)
+    {
+        $result = "\033[m";
+        foreach (array_filter(explode(PHP_EOL, $output)) as $lineNumber => $line) {
+            if (preg_match('/^[0-9]+/', $line)) {
+                $result .= $this->trimOutputLine($line, $lineNumber) . PHP_EOL;
+            } else {
+                $result .= $line . PHP_EOL;
+            }
+        }
+        return $result;
+    }
+
+    protected function trimOutputLine($line, $lineNumber)
+    {
+        $wordColor = "\033[1;31";
+        $resetColor = "\033[m";
+        $colonColor = "\033[36m";
+        $coloredColon = $colonColor . ':' . $resetColor;
+
+        $before = 20;
+        $after = 20;
+
+        if (strlen($line) >= 50) {
+            $lastPos = 0;
+            $positionsFirst = [];
+            $positionsSecond = [];
+            $parts = [];
+
+            while (($lastPos = mb_strpos($line, $wordColor, $lastPos)) !== false) {
+                $positionsFirst[] = $lastPos;
+                $lastPos = $lastPos + mb_strlen($wordColor);
+            }
+            $lastPos = 0;
+            while (($lastPos = mb_strpos($line, $resetColor, $lastPos)) !== false) {
+                $positionsSecond[] = $lastPos;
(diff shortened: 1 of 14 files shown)


veewee:
Thanks for the PR @Kanti! At first sight, it looks like a great formatter.
This formatter requires ANSI escape codes but there are users on e.g. who use terminals which don't support those colors. Can you check `IOInterface::isDecorated()` to enable / disable colored output?

I'll try to dive in the code somewhere this week.

Kanti:
I have found something strange. 
I have added This:
`````php
var_dump(get_class($container->get('grumphp.io')));
````
in this line https://github.com/phpro/grumphp/blob/master/src/GrumPHP/Console/Application.php#L178

There I should get ``string(20) "GrumPHP\IO\ConsoleIO"`` printed.
But instead I get ``string(17) "GrumPHP\IO\NullIO"`` printed.

Is this a Bug? Could you fix it?

**Comment by veewee** [comment]
That indeed is something strange. I'll have to dive into that one.

**Comment by veewee** [comment]
Hi @Kanti,

Can you tell me if this is still an issue if you mark the task as a lazy service?

```yml
services:
    formatter.git_blacklist:
        class: GrumPHP\Formatter\GitBlacklistFormatter
        lazy: true
        arguments:
            - '@grumphp.io'
```

Currently it is possible that the `TaskRunner` is initialized before the `configureIO` method is being called. This is indeed a problem which we need to fix. In the meantime, the `lazy` option can be a quick fix.

src/GrumPHP/Formatter/GitBlacklistFormatter.php
```diff
@@ -0,0 +1,106 @@
...
+            } else {
+                $result .= $line . PHP_EOL;
+            }
+        }
+        return trim($result);
+    }
+
+    private function trimOutputLine($line, $lineNumber)
+    {
+        $wordColor = "\033[1;31";
```
veewee: Maybe these colors should be class constants.

src/GrumPHP/Formatter/GitBlacklistFormatter.php
```diff
@@ -0,0 +1,106 @@
...
+            if (preg_match('/^[0-9]+/', $line)) {
+                $result .= $this->trimOutputLine($line, $lineNumber) . PHP_EOL;
+            } else {
+                $result .= $line . PHP_EOL;
+            }
+        }
+        return trim($result);
+    }
+
+    private function trimOutputLine($line, $lineNumber)
```
veewee: Missing method docblocks.

src/GrumPHP/Formatter/GitBlacklistFormatter.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $wordColor = "\033[1;31";
+        $resetColor = "\033[m";
+        $colonColor = "\033[36m";
+        $coloredColon = $colonColor . ':' . $resetColor;
+
+        $before = 20;
+        $after = 20;
+
+        if (strlen($line) >= 80) {
```
veewee: It's more readable if you invert the if statement and add an early return at this point.
You might want to split the logic in this if into a new private method.

src/GrumPHP/Formatter/GitBlacklistFormatter.php
```diff
@@ -0,0 +1,106 @@
...
+            $lastPos = 0;
+            $positionsFirst = array();
+            $positionsSecond = array();
+            $parts = array();
+
+            while (($lastPos = mb_strpos($line, $wordColor, $lastPos)) !== false) {
```
veewee: There are a lot of `while` iterations going on in here. Can you add some comments above these iterations about what those iterations are doing? Are there infinite loops possible?

src/GrumPHP/Task/Git/Blacklist.php
```diff
@@ -16,6 +20,22 @@
 class Blacklist extends AbstractExternalTask
 {
     /**
+     * @var IOInterface
+     */
+    private $IO;
+
+    public function __construct(
```
veewee: Missing docblocks

src/GrumPHP/Formatter/GitBlacklistFormatter.php
Kanti: Are there infinite loops possible?
=> no, $lastPos will get bigger on every iteration.

Kanti:
Hi @veewee can you look into the bug in the next days?

veewee:
Hi @Kanti,

I'll try to pick it up somewhere this week. Sorry for the latency :)

veewee:
Hi @Kanti,

I've added late input / output binding to the console IO.
Can you verify if this works for you? It seems to work here.

Kanti:
all good from my side. 😃

PR #189, opened 2016-09-21
Fix PhpMd task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

[PHPMD expects multiple source files to be comma separated](https://github.com/phpmd/phpmd#using-multiple-source-files-and-folders). Currently the files are separated with spaces which makes PHPMD think that the second file is the report format it should use which results in an error (since the name of the second file is not a valid report format).

```shell
Running task  9/11: PhpMd
Aborted ...

Can't find the custom report class: path/to/the/second/File.php
```

### DIFF
--- a/spec/GrumPHP/Collection/ProcessArgumentsCollectionSpec.php
+++ b/spec/GrumPHP/Collection/ProcessArgumentsCollectionSpec.php
@@ -103,4 +103,15 @@ class ProcessArgumentsCollectionSpec extends ObjectBehavior
             'file2.txt',
         ));
     }
+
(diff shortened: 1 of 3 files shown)


veewee:
Hi @jyggen,

Thanks for noticing!
Can you fix the CI error and add a spec test for this new function?

jyggen:
@veewee: Done :+1:

veewee:
Thanks!

Issue #192, opened 2016-09-28
Add phpcpd task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

More info https://github.com/sebastianbergmann/phpcpd

jyggen:
I've started on a PR for this, will hopefully submit it later today :+1:

PR #195, opened 2016-10-02
Add PhpCpd task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #192

# New Task Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ To make GrumPHP even more awesome, it will suggest installing some extra package
 - phpspec/phpspec : ~2.1
 - phpunit/phpunit : ~4.5
 - roave/security-advisories : dev-master@dev
+- sebastian/phpcpd : ~2.0
 - squizlabs/php_codesniffer : ~2.3
 - codeception/codeception : ~2.1
 - sensiolabs/security-checker : ~3.0
@@ -96,6 +97,7 @@ parameters:
         gulp: ~
         jsonlint: ~
         phing: ~
+        phpcpd: ~
         phpcs: ~
         phpcsfixer: ~
         phpcsfixer2: ~
@@ -146,7 +148,6 @@ This package has been tested with following git clients:
 
 Following tasks are still on the roadmap:
 
-- phpcpd
 - phpdcd
 - twig lint
 - symfony validation
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,7 @@
         "malukenho/kawaii-gherkin": "Lets GrumPHP lint your Gherkin files.",
         "phpspec/phpspec": "Lets GrumPHP spec your code.",
         "phpunit/phpunit": "Lets GrumPHP run your unit tests.",
+        "sebastian/phpcpd": "Lets GrumPHP find duplicated code.",
         "squizlabs/php_codesniffer": "Lets GrumPHP sniff on your code.",
         "roave/security-advisories": "Lets GrumPHP be sure that there are no known security issues.",
         "sensiolabs/security-checker": "Lets GrumPHP be sure that there are no known security issues.",
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -22,6 +22,7 @@ parameters:
         jsonlint: ~
         make: ~
         phing: ~
+        phpcpd: ~
         phpcs: ~
         phpcsfixer: ~
         phpcsfixer2: ~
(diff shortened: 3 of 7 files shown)


veewee:
Hi @jyggen,

It looks pretty good! Can you also add the `exclude` option?
Normally we wouldn't like to run this tools on spec or test files etc.

Thanks!

jyggen:
I tried adding that option at first but the problem I'm facing with both `--exclude` and `--names-exclude` is that they only seem to work on sub-directories and files within the analyzed location, and since GrumPHP gives `phpcpd` a list of exact file locations those options seem to be simply ignored. 

Running it on a specific file:
```shell
$ ./vendor/bin/phpcpd --exclude=spec --exclude=vendor spec/GrumPHP/Task/PhpCsFixerSpec.php

phpcpd 2.0.4 by Sebastian Bergmann.

Found 1 exact clones with 20 duplicated lines in 1 files:

  -	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:87-107
 	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:113-133

12.90% duplicated lines out of 155 total lines of code.

Time: 30 ms, Memory: 4.00MB
```

Running it on the root folder:
```shell
$ ./vendor/bin/phpcpd --exclude=spec --exclude=vendor .
phpcpd 2.0.4 by Sebastian Bergmann.

0.00% duplicated lines out of 8668 total lines of code.

Time: 79 ms, Memory: 8.00MB
```

However; I've pushed an implementation of an `exclude` option using `FilesCollection` (with some minor optimizations) instead of what's built into `phpcpd`. Does it seem like a good alternative or should I simply leave out an `exclude` option for now?

veewee:
Does the tool only detect copy-pasted code between the files you add as a parameter to the command? So for example: If I commit `class1.php` which contains duplicate code from `class2.php`, it won't detect the duplicate code since it is only checking `class1.php`?
Maybe it's better to let the tool run on the complete codebase instead of only on the committed files?

jyggen:
Correct. It only detect copy-pasted code between files added as parameters, so it's probably better to let the tool run on the complete codebase. Any pointers on how to make it do that?

```shell
$ ./vendor/bin/phpcpd spec/GrumPHP/Task/PhpCsFixerSpec.php
phpcpd 2.0.4 by Sebastian Bergmann.

Found 1 exact clones with 20 duplicated lines in 1 files:

  -	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:87-107
 	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:113-133

12.90% duplicated lines out of 155 total lines of code.

Time: 31 ms, Memory: 4.00MB
```

```shell
$ ./vendor/bin/phpcpd spec/GrumPHP/Task/
phpcpd 2.0.4 by Sebastian Bergmann.

Found 3 exact clones with 120 duplicated lines in 2 files:

  -	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:87-107
 	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:113-133

  -	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:55-87
 	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerV2Spec.php:57-89

  -	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerSpec.php:86-154
 	/path/to/grumphp/spec/GrumPHP/Task/PhpCsFixerV2Spec.php:88-156

3.61% duplicated lines out of 3327 total lines of code.

Time: 80 ms, Memory: 6.00MB
```

veewee:
Maybe it's best to let the tool run in the current working directory by specifying the source `.` or `getcwd()`?
This way, the tool will run on the full codebase and you can specify which directories / files you want to exclude for the phpcpd task.
Another option is to run it in the `$this->grumPHP->getGitDir();` folder, but this might be overkill in projects with a nested structure.

jyggen:
@veewee: Done.  I added `vendor` as the default exclude since I doubt anyone will ever want to find copy-pasted code in there.

veewee:
Hi @jyggen,

Now that I see the code, I was thinking in making it configurable.
Instead of using the `getcwd()` method, we could read it from `$config['directory']` or something like that. By default the value is set to `.` which means it will load the current directory.
This way it is easy to configure it to scan e.g. the `src` folder, but exclude specific folders like e.g. `src/Resources`. What do you think about this feature?

jyggen:
@veewee: Sounds good to me. Implemented :)

veewee:
Super! :) I've merged it.

PR #197, opened 2016-10-03
Add an option to always execute PHPUnit

kix:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | none

This adds an option to always run PHPUnit tests, even if no PHP files were changed. Useful by itself, but also makes sure there's always something to retrieve coverage from.

Currently, for a freshly cloned/installed project, I won't get any coverage at all, it would just break the build.

### DIFF
--- a/doc/tasks/phpunit.md
+++ b/doc/tasks/phpunit.md
@@ -10,7 +10,7 @@ parameters:
         phpunit:
             config_file: ~
             group: []
-            always_execute: ~
+            always_execute: false
 ```
 
 **config_file**
--- a/doc/tasks/phpunit.md
+++ b/doc/tasks/phpunit.md
@@ -10,6 +10,7 @@ parameters:
         phpunit:
             config_file: ~
             group: []
+            always_execute: ~
 ```
 
 **config_file**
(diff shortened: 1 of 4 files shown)


kix:
Whoops, just found the specs, will fix those.

kix:
I've flipped the default, now it defaults `always_execute` to `false`, and the specs have been updated accordingly.

doc/tasks/phpunit.md
```diff
@@ -10,6 +10,7 @@ parameters:
         phpunit:
             config_file: ~
             group: []
+            always_execute: ~
```
veewee: Can you also change this option to the default value `false`?

veewee:
Thanks, it's looking good!
I've added a small remark.

kix:
Sure, will update ASAP

veewee:
I've changed it. Thanks for the PR!

PR #198, opened 2016-10-03
Fix triggered_by config for PhpCpd task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets |

Currently (thanks to me) the `triggered_by` option isn't really doing anything so let's (ab?)use phpcpd's own `--names` parameter :)

### DIFF
--- a/src/GrumPHP/Task/PhpCpd.php
+++ b/src/GrumPHP/Task/PhpCpd.php
@@ -69,10 +69,14 @@ class PhpCpd extends AbstractExternalTask
         }
 
         $arguments = $this->processBuilder->createArgumentsForCommand('phpcpd');
+        $extensions = array_map(function ($extension) {
+            return sprintf('*.%s', $extension);
+        }, $config['triggered_by']);
 
         $arguments->addArgumentArray('--exclude=%s', $config['exclude']);
         $arguments->addRequiredArgument('--min-lines=%u', $config['min_lines']);
         $arguments->addRequiredArgument('--min-tokens=%u', $config['min_tokens']);
+        $arguments->addOptionalCommaSeparatedArgument('--names=%s', $extensions);
         $arguments->addOptionalArgument('--fuzzy', $config['fuzzy']);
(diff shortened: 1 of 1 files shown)


veewee:
@jyggen, the way it was, is how other tasks are solving the problem as well.
Imagine you commit only an XML file, then you don't want the phpcpd task to run since it is irrelevant to the commited context.
In my opinion there is no need for using the `--names` option if the filtering is done with the `extensions()` filter. Since it is running on git commands, it's probably faster then the filesystem.

If you have another opinion on the matter, feel free to enlighten me :)

jyggen:
That's fair, and I agree that it probably should be kept to avoid running it when no relevant files have been changed.

I do believe `--names` is still necessary though for cases where people want to detect duplicated code in things that aren't `.php` files (like Drupal's `.inc` and `.module`), and since `$files` isn't used by the task (since we changed to scanning the entire source tree based on `$config['directory']`) it needs to be provided to the task in some other way, hence why `--names` is used.

veewee:
+1 Let's re-add the extensions() check and leave the `--names` option as is!

jyggen:
@veewee: Fixed!

src/GrumPHP/Task/PhpCpd.php
```diff
@@ -69,10 +69,16 @@ public function run(ContextInterface $context)
         }
 
         $arguments = $this->processBuilder->createArgumentsForCommand('phpcpd');
+        $extensions = array_reduce($config['triggered_by'], function (array $carry, $extension) {
```
veewee: Can you use `array_map()` instead of `array_reduce()`?

jyggen: Done!

veewee:
Great!

PR #199, opened 2016-10-03
Add TwigLint task

jyggen:
| Q | A |
| --- | --- |
| Branch | master |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Documented? | yes |
| Fixed tickets |  |

Lint Twig templates!

*Awaiting: asm89/twig-lint#31*

# New Task Checklist:

### DIFF
--- a/src/GrumPHP/Task/TwigLint.php
+++ b/src/GrumPHP/Task/TwigLint.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace GrumPHP\Task;
+
+use GrumPHP\Exception\RuntimeException;
+use GrumPHP\Runner\TaskResult;
+use GrumPHP\Linter\Twig\TwigLinter;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Class TwigLint
+ *
+ * @package GrumPHP\Task
+ */
+class TwigLint extends AbstractLinterTask
+{
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return 'twiglint';
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = parent::getConfigurableOptions();
+        $resolver->setDefaults(array(
+            'triggered_by' => array('twig'),
+        ));
+
+        $resolver->addAllowedTypes('triggered_by', array('array'));
+
+        return $resolver;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function canRunInContext(ContextInterface $context)
+    {
+        return ($context instanceof GitPreCommitContext || $context instanceof RunContext);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
+        $files = $context->getFiles()->extensions($config['triggered_by']);
+
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        try {
+            $lintErrors = $this->lint($files);
+        } catch (RuntimeException $e) {
+            return TaskResult::createFailed($this, $context, $e->getMessage());
+        }
+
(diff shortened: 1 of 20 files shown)


src/GrumPHP/Task/TwigLint.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return $context instanceof RunContext || $context instanceof GitPreCommitContext;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
```
veewee: It's best to filter on the .twig extension or make the extension configurable.

src/GrumPHP/Task/TwigLint.php
```diff
@@ -0,0 +1,74 @@
...
+        $arguments = $this->processBuilder->createArgumentsForCommand('twig-lint');
```
veewee: Does this linter handle environment specific errors like e.g. undefined functions etc?

jyggen:
I changed the approach a bit to use GrumPHP's linter system instead (CI builds fail for some weird reason though, and the error output isn't really helping so not sure what's going on).

As for custom/undefined filters/functions/tags I don't think that's supported by the linter since those are registered by the user _somewhere_ in their code. I'll check whether or not the linter crashes on those though. Maybe one option if we want to support this would be to make it configurable, but that requires the user to manually keep a list of filters/functions/tags in `grumphp.yml` up-to-date.

veewee:
@jyggen,

You get more feedback about the exception when you run:

```
./vendor/bin/phpspec run -v
```

You should see some deprecation warnings like:

> Deprecated: The Twig_TokenParserBroker class is deprecated since version 1.12 and will be removed in 2.0. in /..../grumphp/vendor/twig/twig/lib/Twig/TokenParserBroker.php on line 36

Besides that, I get 2 fatal exceptions:

> GrumPHP\Exception\RuntimeException: The stashed changes could not be applied. Please run `git stash pop` manually!More info: Exception in /.../vendor/phpspec/prophecy/src/Prophecy/Promise/ThrowPromise.php:79

And 

> PHP Fatal error:  Uncaught Exception in /.../vendor/phpspec/prophecy/src/Prophecy/Promise/ThrowPromise.php:79

I'll take a deeper look in the code somewhere this week.

jyggen:
I updated the implementation to use Twig directly as well as adding some PHPUnit tests for the linter itself.

Something that's missing now though, which the linter package took care of, is handling undefined functions or tests. I guess there are three roads we can go down;

1) Get the package maintainer to use less deprecated code and go back to that.
2) Implement something similar ourselves.
3) Force the user to define all custom things in their `grumphp.yml`.

veewee:
@jyggen,
About the 3 options:

1) This seems like the prefered option for me. I don't like to reinvent te wheel and would love to use a 3rd party for this. Maybe there are still other packages that perform twig linting or some options inside twig itself?
2)  This is also possible, but that would be something we'll need to maintain ourselves.
3) Rather not: If you are using e.g. 3rd party Symfony Bundles, you don't really know which extensions are added by these bundles. Another idea would be to create a file that loads the environment from e.g. symfony, but that seems like an overkill to me at this point. The good thing about this would be that the linter is exactly the same as during runtime.

jyggen:
Yeah, I agree with you. I'll open a issue over at [asm89/twig-lint](https://github.com/asm89/twig-lint) about the deprecated notices.

veewee:
Allright, keep me posted on the issue :)

jyggen:
Will do! Here's the issue for reference: https://github.com/asm89/twig-lint/issues/31

veewee:
It doesn't look like the package is very active at the moment.

jyggen:
Not really, no :(

veewee:
We've moved from PSR-0 to PSR-4. Can you rebase your changes against current master?
It looks like the maintainer of the twig lint is waiting for a PR to fix the open issue. Is this something you are willing to create or do we keep on waiting until somebody else does?

jyggen:
Yeah, I saw that. I'll try to send a PR in the near future :)

welcoMattic:
Any update on this PR? ;)

veewee:
@welcoMattic, We cannot use this task if https://github.com/asm89/twig-lint/issues/31 isn't implemented first.

Issue #50, opened 2015-10-26
Make it possible to skip tests.


veewee:
For example:

``` sh
grumphp run --skip=phpunit
grumphp run --skip=phpunit,behat
```

veewee:
It seems like a better idea to specify task suites in the grumphp configuration.
This way, it is possible a different CI test suite then the pre-commit test-suite.

I was thinking about the following logic:

`git:pre-commit` uses the git-pre-commit suite if it exists. If it does not exist it used all specified tasks.
`git:commit-msg uses` the git-commit-essage suite if it exists. If it does not exist it used all specified tasks.
`run` uses all specified tasks. When a `--testsuite=[suite]` is specified, the suite is used or an exception is thrown when it does not exist.

When this is implemented, it would be possible to:

- Run fast tasks during commit only
- Run slower tasks during CI
- Run GrumPHP for functional / acceptance tests
- Run parts of the tests manually by specifying a testsuite.

Any thoughts? 
/cc @aderuwe

aderuwe:
+1

PR #201, opened 2016-10-10
Fix phpcpd in case-sensitive environments

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | -
| Fixed tickets | -

Seems like I sneaked in a typo in the `phpcpd` configuration which makes GrumPHP crash and burn in case-sensitive environments. Sorry about that! Maybe it would be a good idea to cover the configurations with tests somehow so CI will catch similar errors in the future?

### DIFF
--- a/resources/config/tasks.yml
+++ b/resources/config/tasks.yml
@@ -122,7 +122,7 @@ services:
             - {name: grumphp.task, config: phing}
 
     task.phpcpd:
-        class: GrumPHP\Task\Phpcpd
+        class: GrumPHP\Task\PhpCpd
         arguments:
           - '@config'
           - '@process_builder'


veewee:
Thanks!

We were thinking about adding some behat or codeception CLI tests to the project for testing if every task works. These tasks will probably get pretty slow because we'll have to create a file strucutre for every task etc. This means we'll also have to implement #50 to make sure the tests only run on CI. We'll also have to put a lot of time in creating all the happy and sad paths for every task. In other words: it is possible but it will consume a lot of time to get the testing right which I don't have at the moment :)

Feel free to help out where you can.

PR #203, opened 2016-10-15
Add Doctrine ORM task

jyggen:
This task makes it possible to validate mappings for Doctrine ORM. By default it'll only validate the mapping and the entity classes, but it's also possible to have it check that the database is in sync. Skipping it by default is personal preference, but I have no problem changing that :)

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | -

### DIFF
--- a/doc/tasks/doctrine_orm.md
+++ b/doc/tasks/doctrine_orm.md
@@ -1,0 +1,32 @@
+# Doctrine ORM
+
+The Doctrine ORM task will validate that your Doctrine mapping files and check if the mapping is in sync with the database.
+It lives under the `doctrine` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        doctrine_orm:
+            skip_mapping: false
+            skip_sync: false
+            triggered_by: ['php', 'xml', 'yml']
+```
+
+**skip_mapping**
+
+*Default: false*
+
+With this parameter you can skip the mapping validation check.
+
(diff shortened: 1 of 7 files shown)


veewee:
Hi @jyggen,

This PR looks great! One remark: can you rename the task to `doctrine_orm`, `doctrine_orm_schema` or something like that?
Other Doctrine modules like ODM also have a schema validation command.

Thanks!

jyggen:
@veewee: Absolutely. Done! :+1:

doc/tasks/doctrine_orm.md
```diff
@@ -0,0 +1,32 @@
...
+            triggered_by: ['php', 'xml', 'yml']
+```
+
+**skip_mapping**
+
+*Default: false*
+
+With this parameter you can skip the mapping validation check.
+
+**skip_sync**
```
veewee: Small question: Why is this sync skipping enabled by default? I prefer to use the defaults of the actual commandline tool.

jyggen: Changed it, having the defaults of the CLI tool makes sense. As I wrote in the PR it's just a personal preference of mine, don't really care if the DB is in sync when I commit and on CI the DB is always migrated anyway :)

veewee:
Thanks!

Issue #179, opened 2016-09-06
Aborting on Phpcs task


jtgrimes:
| Q               | A
| --------------- | ---
| Branch          | 0.9.1
| Bug?            | yes
| New feature?    | no
| Question?       | yes
| Documentation?  | no
| Related tickets | comma-separated list of related tickets

**My configuration**
```yaml
# grumphp.yml
parameters:
    bin_dir: "./vendor/bin"
    git_dir: "."
    hooks_dir: ~
    hooks_preset: local
    stop_on_failure: true
    ignore_unstaged_changes: false
    process_timeout: 240
    ascii:
        failed: ~
        succeeded: ~
    tasks:
#        composer: ~
#        git_blacklist: ~
#        git_commit_message: ~
        phpcs:
            standard: PSR2
            show_warnings: true
            tab_width: ~
            ignore_patterns:
              - tests/
              - database/
              - features/
              - autoload.php
              - .blade.php
              - resources/lang
            sniffs: []
            triggered_by: [php]
        phpunit: ~
#        behat:
#            config: ~
#            format: progress
#            stop_on_failure: false
    extensions: []
```

**Steps to reproduce:**
```sh
vendor\bin\grumphp run
```
Running on Windows, works fine on the same machine on other repos so it's not a PHP configuration issue

**Result:**
```
GrumPHP is sniffing your code!
Running task 1/2: Phpcs
Aborted ...
```

**Question**
Is there any way to have grumphp log what it's doing so that I don't have to step through code to figure out more about why this isn't working?

veewee:
You've configured the `stop_on_failure` property which will stop running tasks once one of the tasks fail. So in this case, Phpcs will fail and the state will be aborted. Normally you should see the output of the phpcs task underneeth the aborted step.

You can also run php with the `-vvv` option. This one will show you some more output. 
Once the tasks start running, there isn't a lot of additional output for the moment.

jtgrimes:
Thanks for a kind answer to a dumb question.

If I could add a --debug flag which would echo `$process->getCommandLine()` in the task classes, would that be a welcome PR, or unnecessary?

veewee:
It sure is welcome. I think it's easiest to add the `IOInterface` as a dependency to the `ProcessBuilder` class.
In the `buildProcess()` you can check if the `isVeryVerbose()` flag is set and `write()` the `getCommandLine()`. 
[...]

PR #206, opened 2016-10-18
Write command to output stream when -vvv flag is set

jtgrimes:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
Documented?   | no
| Fixed tickets | #179

Per a discussion on issue #179, this PR outputs a copy of each task's command when the -vvv flag is used. I didn't update any docs, since I couldn't see where the -vvv flag is mentioned anywhere.

### DIFF
--- a/src/GrumPHP/Process/ProcessBuilder.php
+++ b/src/GrumPHP/Process/ProcessBuilder.php
@@ -4,6 +4,7 @@ namespace GrumPHP\Process;
 
 use GrumPHP\Collection\ProcessArgumentsCollection;
 use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\IO\IOInterface;
 use GrumPHP\Locator\ExternalCommand;
 use Symfony\Component\Process\Process;
 use \Symfony\Component\Process\ProcessBuilder as SymfonyProcessBuilder;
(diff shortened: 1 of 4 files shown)


src/GrumPHP/Process/ProcessBuilder.php
```diff
@@ -72,4 +80,14 @@ private function getCommandLocation($command)
...
         return $this->externalCommandLocator->locate($command);
     }
+
+    /**
+     * @param Process $process
+     */
+    private function writeVerboseOutput($process)
+    {
+        if ($this->io->isVeryVerbose()) {
+            $this->io->write(PHP_EOL.'Command: '.$process->getCommandLine(), true);
```
veewee: Why did you add PHP_EOL? The last parameter of the method is already adding a newline.

jtgrimes: The last arguments adds a newline at the end of the line, but the symfony ProgressBar class does't add a newline at the end of the task, so without it, we end up with ugly output like in the attached image.

grumphp

Another option would be to add the newline to the `$progressFormat` property of `Event\Subscirber\ProgressSubscriber` - in that case, we'd have a blank line between every task, which wasn't something I wanted to do.

jtgrimes:
The spec should run correctly now - I develop on Windows and forgot that the command line arguments are escaped differently on Linux. The test should be platform independent now.

veewee:
Thanks!

PR #210, opened 2016-10-20
Run PhpCsFixer tasks asynchronously

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | -

PhpCsFixer takes quite some time when running it on changed files, and this is of course because GrumPHP needs to run one process per file. Symfony's Process component supports running processes asynchronously, so I thought why not give it a spin!

The following benchmark is performed on 23 files with PhpCsFixerV2 as the only task. The former is GrumPHP 0.9.6 and the latter is with this PR. From 10s to 2s is pretty sweet!

```shell
$ time ./vendor/bin/grumphp run
Running task 1/1: PhpCsFixerV2

real    0m10.478s
user    0m1.734s
sys     0m8.297s
```

```shell
$ time ./vendor/bin/grumphp run
Running task 1/1: PhpCsFixerV2

real    0m2.350s
user    0m1.563s
sys     0m7.078s
```

The `array_chunk()` and extra `foreach` loop is there because I don't like the idea of potentially kicking off hundreds of `php-cs-fixer` processes at the same time. Right now the size of each chuck is hardcoded to ten, maybe something worth making configurable?

### DIFF
--- a/spec/GrumPHP/Process/AsyncProcessRunnerSpec.php
+++ b/spec/GrumPHP/Process/AsyncProcessRunnerSpec.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace spec\GrumPHP\Process;
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Process\AsyncProcessRunner;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Prophet;
+
+/**
+ * @mixin AsyncProcessRunner
+ */
+class AsyncProcessRunnerSpec extends ObjectBehavior
+{
+    public function let(GrumPHP $grumPHP) {
+        $this->beConstructedWith($grumPHP);
+
+        $grumPHP->getProcessAsyncWaitTime()->willReturn(0);
+        $grumPHP->getProcessAsyncLimit()->willReturn(5);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('GrumPHP\Process\AsyncProcessRunner');
+    }
+
+    function it_should_be_able_to_run_processes()
+    {
+        $prophet = new Prophet();
+        $processes = array();
+
+        for ($i = 0; $i < 20; $i++) {
+            $process = $prophet->prophesize('Symfony\Component\Process\Process');
+
+            $process->started = false;
+            $process->terminated = false;
+
+            $process->start()->will(function () use ($process) {
+                $process->started = true;
+            })->shouldBeCalledTimes(1);
+
+            $process->isTerminated()->will(function () use ($process) {
+                if (!$process->terminated) {
+                    $process->terminated = true;
+                    return false;
+                }
+
+                return true;
+            })->shouldBeCalledTimes(2);
+
+            // The number of times isStarted() is called starts at 3
+            // and increases by 2 after each chunk of five processes.
+            $process->isStarted()->will(function () use ($process) {
+                return $process->started;
+            })->shouldBeCalledTimes(floor($i / 5) * 2 + 3);
+
+            $processes[] = $process->reveal();
+        }
+
+        $this->run($processes);
+        $prophet->checkPredictions();
+    }
+}
--- a/src/GrumPHP/Configuration/GrumPHP.php
+++ b/src/GrumPHP/Configuration/GrumPHP.php
@@ -73,6 +73,22 @@ class GrumPHP
     }
 
     /**
+     * @return int
+     */
+    public function getProcessAsyncLimit()
+    {
+        return (int) $this->container->getParameter('process_async_limit');
+    }
+
+    /**
+     * @return int
+     */
+    public function getProcessAsyncWaitTime()
+    {
+        return (int) $this->container->getParameter('process_async_wait');
+    }
+
+    /**
      * @return float|null
      */
     public function getProcessTimeout()
--- a/src/GrumPHP/Process/AsyncProcessRunner.php
+++ b/src/GrumPHP/Process/AsyncProcessRunner.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace GrumPHP\Process;
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Runner\TaskResult;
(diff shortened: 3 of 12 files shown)


veewee:
Great idea!

Can you move the logic of the chunked process runner into a separate service named `ParallelProcessRunner` or something like that?
This way we can reuse the logic in other tasks when we need to.
Lets also try to keep 1 level of indentation in the classes. 

Currently the processes are running in chunks of 10 which means that it waits untill every task is finished before it starts it's next chunk. It would be really  nice if it would start a new task once an old task finished with a maximum of 10 tasks. Would that be possible?

jyggen:
I've pushed a refactor that uses a new `AsyncProcessRunner`. Let me know what you think.

src/GrumPHP/Process/AsyncProcessRunner.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @param Process[] $processes
+     */
+    public function run(array $processes)
+    {
+        $this->processes = $processes;
+        $this->running = 0;
+
+        while ($this->watchProcesses()) {
+            usleep(1000);
```
veewee: Maybe this setting could also be configurable?

src/GrumPHP/Process/AsyncProcessRunner.php
```diff
@@ -0,0 +1,91 @@
...
+        $shouldRunAgain = false;
+
+        foreach ($this->processes as $key => $process) {
+            $isTerminated = $this->handleProcess($process);
+
+            if ($isTerminated) {
+                unset($this->processes[$key]);
+            }
+
+            $shouldRunAgain = max($shouldRunAgain, !$isTerminated);
```
veewee: why are you using the `max()` method on booleans? This is just a regular `or` isn't it?

src/GrumPHP/Process/AsyncProcessRunner.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * @return bool
+     */
+    private function handleProcess(Process $process)
+    {
+        $isTerminated = false;
+
+        if (!$process->isStarted() && $this->running < $this->config->getProcessAsyncLimit()) {
```
veewee: Try to avoid else / elseif's.
Just invert the if and return early.
We're trying to folow the [object calisthenics](http://www.slideshare.net/guilhermeblanco/object-calisthenics-applied-to-php) rules.

src/GrumPHP/Process/AsyncProcessRunner.php
```diff
@@ -0,0 +1,91 @@
...
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Runner\TaskResult;
+use Symfony\Component\Process\Process;
+
+/**
+ * Class AsyncProcessRunner
+ *
+ * @package GrumPHP\Process
+ */
+class AsyncProcessRunner
```
veewee: Can you add some spec tests to make sure this class works as expected?

src/GrumPHP/Process/AsyncProcessRunner.php
```diff
@@ -0,0 +1,91 @@
...
+            $isTerminated = $this->handleProcess($process);
+
+            if ($isTerminated) {
+                unset($this->processes[$key]);
+            }
+
+            $shouldRunAgain = max($shouldRunAgain, !$isTerminated);
+        }
+
+        return $shouldRunAgain;
```
veewee: Can't the logic for `$shouldRunAgain` be simplified in `count($this->processes) !== 0`

jyggen:
Refactored it a bit and added some tests :+1:

src/GrumPHP/Configuration/GrumPHP.php
```diff
@@ -73,6 +73,22 @@ public function ignoreUnstagedChanges()
     }
 
     /**
+     * @return int
+     */
+    public function getProcessAsyncLimit()
```
veewee: Can you ass a spec for this method?

src/GrumPHP/Configuration/GrumPHP.php
```diff
@@ -73,6 +73,22 @@ public function ignoreUnstagedChanges()
...
+    {
+        return (int) $this->container->getParameter('process_async_limit');
+    }
+
+    /**
+     * @return int
+     */
+    public function getProcessAsyncWaitTime()
```
veewee: Can you ass a spec for this method?

spec/GrumPHP/Process/AsyncProcessRunnerSpec.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @mixin AsyncProcessRunner
+ */
+class AsyncProcessRunnerSpec extends ObjectBehavior
+{
+    public function let(GrumPHP $grumPHP) {
+        $this->beConstructedWith($grumPHP);
+
+        $grumPHP->getProcessAsyncWaitTime()->willReturn(1000);
```
veewee: Maybe you can set this value to 0 to make sure the tests aren't slowed down by the `usleep()` call?

spec/GrumPHP/Process/AsyncProcessRunnerSpec.php
```diff
@@ -0,0 +1,48 @@
...
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Process\AsyncProcessRunner;
+use PhpSpec\ObjectBehavior;
+use Symfony\Component\Process\Process;
+
+/**
+ * @mixin AsyncProcessRunner
+ */
+class AsyncProcessRunnerSpec extends ObjectBehavior
```
veewee: Maybe it's a good idea to test the async process limit part as well?

jyggen:
Done and done!

veewee:
Thanks for your work!

Issue #208, opened 2016-10-20
NPM scripts task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

Run a specific script from the nodejs package.json file with npm.

https://docs.npmjs.com/misc/scripts

Asuza:
What script is supposed to be run via npm?

veewee:
@Asuza,
It should be configurable like the [grunt](https://github.com/phpro/grumphp/blob/master/doc/tasks/grunt.md) task.

It looks like @Kanti picket it up already :)

PR #211, opened 2016-10-21
[FEATURE] add npm_script Task 

Kanti:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #208

# New Task Checklist:

### DIFF
--- a/src/GrumPHP/Task/NpmScript.php
+++ b/src/GrumPHP/Task/NpmScript.php
@@ -63,7 +63,7 @@ class NpmScript extends AbstractExternalTask
         $arguments->addRequiredArgument('%s', $config['script']);
 
         $process = $this->processBuilder->buildProcess($arguments);
-        $process->setWorkingDirectory($config['working_directory']);
+        $process->setWorkingDirectory(realpath($config['working_directory']));
         $process->run();
 
         if (!$process->isSuccessful()) {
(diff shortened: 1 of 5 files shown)


Kanti:
Please Review! 😅

veewee:
Thanks @Kanti! The code looks good.

2 small questions:

- Maybe it's better to set the defaults for the `triggered_by` option since it mostly is triggered by js / css changes? These are the defaults for Grunt: [js, jsx, coffee, ts, less, sass, scss]
- Is it possible to specify the location of the `package.json` file for people with an exotic file structure? Can't find how to do it in the documentation, but it could also be done by specifying a working directory on the process.

Kanti:
added changes Please Review 😅

src/GrumPHP/Task/NpmScript.php
```diff
@@ -0,0 +1,75 @@
...
+        $files = $context->getFiles()->extensions($config['triggered_by']);
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('npm');
+        $arguments->addRequiredArgument('%s', $config['script']);
+
+        $process = $this->processBuilder->buildProcess($arguments);
+        $process->setWorkingDirectory($config['working_directory']);
```
veewee: Will this use the correct directory if you add relative directories? Maybe we should make it relative to the getcwd() ?

veewee:
Thanks!

Issue #212, opened 2016-10-21
Composer scripts task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

Run a specific script from the composer.json file with composer.

https://getcomposer.org/doc/articles/scripts.md#running-scripts-manually

The script to run should be configurable, just like the [grunt](https://github.com/phpro/grumphp/blob/master/doc/tasks/grunt.md) task.

Kanti:
I will pick this one.

PR #213, opened 2016-10-21
[FEATURE] add composer_script task

Kanti:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #212 

# New Task Checklist:

### DIFF
--- a/src/GrumPHP/Task/ComposerScript.php
+++ b/src/GrumPHP/Task/ComposerScript.php
@@ -62,7 +62,7 @@ class ComposerScript extends AbstractExternalTask
         $arguments = $this->processBuilder->createArgumentsForCommand('composer');
         $arguments->add('run-script');
         $arguments->addRequiredArgument('%s', $config['script']);
-        $arguments->addOptionalArgument('-d %s', $config['working_directory']);
+        $arguments->addOptionalArgument('--working-dir=%s', $config['working_directory']);
 
         $process = $this->processBuilder->buildProcess($arguments);
         $process->run();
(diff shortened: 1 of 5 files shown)


Kanti:
Please Review! 😅

veewee:
Also thanks for this PR!
I've got the same comments as on the npm script task:

- Can you set default triggered_by extensions for the composer script? I was thinking about [php, phtml] (Note: Currently the check for run on every extension is not added?)
- Can you specify an alternative composer.json location just like in the composer task?

Kanti:
added changes Please Review 😅

src/GrumPHP/Task/ComposerScript.php
```diff
@@ -62,6 +64,7 @@ public function run(ContextInterface $context)
         $arguments->addRequiredArgument('%s', $config['script']);
 
         $process = $this->processBuilder->buildProcess($arguments);
+        $process->setWorkingDirectory($config['working_directory']);
```
veewee: The run-script command actually has a working dir option.
Maybe it's better to use that one if the working_directory is set and default it to null?

```
  -d, --working-dir=WORKING-DIR  If specified, use the given directory as working directory.
```

Kanti:
now composer_script uses the -d option from composer.

veewee:
Thanks!

PR #224, opened 2016-11-08
Prefer configured bin directoy over global installations

newLoki:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | -
| Fixed tickets | -

To be able to work on different project with potentially different versions of external tools the configured bin dir is checked first and therefore a project installation of a tool (e.g. per composer) is used if available and the global installed one is just a fallback.
The Symfony `ExecutableFinder` checks for the $PATH first and second the configured bin directory, which causes that a global installed tool was preferred over a project specific version (see https://github.com/symfony/process/blob/master/ExecutableFinder.php#L69).

IMO this was a bug, but I'm happy to hear other arguments too.

Note: I was not sure if this is a change which requires a test as there is just a very basic one for the binary location. If so please give me a short note and I will be happy to try to provide one.

### DIFF
--- a/src/GrumPHP/Locator/ExternalCommand.php
+++ b/src/GrumPHP/Locator/ExternalCommand.php
@@ -44,8 +44,14 @@ class ExternalCommand
      */
     public function locate($command, $forceUnix = false)
     {
-        // Search executable:
-        $executable = $this->executableFinder->find($command, null, [$this->binDir]);
+        
+        // Search executable (prefer local installation over global):
+        if (is_executable($this->binDir . '/' . $command)) {
+            $executable = $this->binDir . '/' . $command;
+        } else {
+            $executable = $this->executableFinder->find($command, null, [$this->binDir]);
+        }
+
         if (!$executable) {
             throw new RuntimeException(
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for the PR. I'll take a look at it later this week.

veewee:
@newLoki, 

This issue has been reported and fixed before in: https://github.com/phpro/grumphp/pull/148 .

If this isn't working for you, can you take a look what is going wrong there? If you aren't able to fix the actual issue, you can always describe me how to reproduce the error.

The reason why we choose not to write logic like the one you committed, is because it is not OS safe. The ExecutableFinder from Symfony creates OS safe-paths and adds extensions like .bat, .exe, ... 
Therefor we choose to change the path variable instead during the initialization of the command helpers.

newLoki:
@veewee Sorry for missing to respond.
IMO than it must be an issue with our project structure I have to find.

But thanks for your support and the clarification.
I will just close the pull request.

Issue #223, opened 2016-11-07
Add PHP version checker task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

Since this tool is striving for good code quality, it is a logical step that the code cannot run on deprecated PHP versions.
A new task should be added that checks current PHP version against a configurable list of PHP versions and there End-Of-Life date.

A list of supported versions end their latest security update date can be found here:
http://php.net/supported-versions.php

PR #227, opened 2016-11-17
phpversion

janvernieuwe:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #223

Checks if the php version is currently supported.
Also allows to manually set a higher version than only the supported ones in the configuration. 

# New Task Checklist:

Add feature to add a php version check


README.md
```diff
@@ -120,6 +120,7 @@ parameters:
         shell: ~
         xmllint: ~
         yamllint: ~
+        phpversion: ~
```
veewee: The tasks are sorted alphabetically.

doc/tasks.md
```diff
@@ -40,6 +40,7 @@ parameters:
         shell: ~
         xmllint: ~
         yamllint: ~
+        phpversion: ~
```
veewee: The tasks are sorted alphabetically.

doc/tasks.md
```diff
@@ -79,6 +80,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 - [Shell](tasks/shell.md)
 - [XmlLint](tasks/xmllint.md)
 - [YamlLint](tasks/yamllint.md)
+- [PhpVersion](tasks/phpversion.md)
```
veewee: The tasks are sorted alphabetically.

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * PhpVersion constructor.
+     * @param GrumPHP $config
+     * @param array $versions
+     */
+    public function __construct(GrumPHP $config, array $versions)
+    {
+        $this->versions = $versions;
+        $options = $config->getTaskConfiguration($this->getName());
```
veewee: You can take a look at other tasks how to set configurable options. This should be done by returning an OptionsResolver in the `getConfigurableOptions()` method. The `getConfiguration()` method used this OptionsResolver to validate the configured options.

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * This methods specifies if a task can run in a specific context.
+     *
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    public function canRunInContext(ContextInterface $context)
+    {
+        return true;
```
veewee: By setting this to true, it will run twice during a commit: once for the pre-commit and once for the commit-msg.

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * @param ContextInterface $context
+     *
+     * @return TaskResultInterface
+     */
+    public function run(ContextInterface $context)
```
veewee: This code isn't testable.
Maybe it's better to move it to a separate class and test all the different cases in that class?
Will it work for people who are using nightly builds and pre-releases?

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+        $versionIsSupported = false;
+        $now = new \DateTime();
+        foreach ($this->versions as $number => $eol) {
+            $eol = new \DateTime($eol);
+            if ($now < $eol && version_compare(PHP_VERSION, $number) >= 0) {
+                $versionIsSupported = true;
+            }
+        }
+        if (!$versionIsSupported) {
+            return TaskResult::createFailed($this, $context, sprintf('PHP version %s is end of life', PHP_VERSION));
```
veewee: Maybe you could add more information about which date the support ended or a link to the supported versions page?

resources/config/tasks.yml
```diff
@@ -281,3 +281,13 @@ services:
             - '@linter.yamllint'
         tags:
             - {name: grumphp.task, config: yamllint}
+
+    task.phpversion:
+        class: GrumPHP\Task\PhpVersion
+        arguments:
+            - '@config'
+            -
+              5.6: '2016-12-31 23:59:59'
```
veewee: Two questions:

- Should we add old versions for a complete overview?
- Should we use the active support or security support as an end-date? Security support is still support ... :)

doc/tasks/phpversion.md
```diff
@@ -0,0 +1,18 @@
...
+    tasks:
+        phpversion:
+            project: 7.0
+```
+
+**project**
+
+*Default: null*
+
+Allows to set a higher version than only currently supported version for your project.
```
veewee: Can you rephrase this description? I didn't get it at first.

spec/GrumPHP/Task/PhpVersionSpec.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    function it_should_run_in_run_context(RunContext $context)
+    {
+        $this->canRunInContext($context)->shouldReturn(true);
+    }
+
+    function it_runs_the_suite(PhpVersion $version, ContextInterface $context)
+    {
+        $version->isInstalled()->willReturn(true);
```
veewee: There is no isInstalled() method on the version class. Also: there is a type mismatch between the let() method and this method, so this means this function isn't working as expected.

spec/GrumPHP/Task/PhpVersionSpec.php
```diff
@@ -0,0 +1,57 @@
...
+        $result = $this->run($context);
+        $result->shouldBeAnInstanceOf('GrumPHP\Runner\TaskResultInterface');
+        $result->isPassed()->shouldBe(true);
+    }
+}
```
veewee: Can you also add a failing test and a test for the project setting?

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * PhpVersion constructor.
+     * @param GrumPHP $grumPHP
+     * @param \GrumPHP\Util\PhpVersion $util
+     */
+    public function __construct(GrumPHP $grumPHP, \GrumPHP\Util\PhpVersion $util)
+    {
+        $this->grumPHP = $grumPHP;
+        $this->util = $util;
```
veewee: maybe the util property isn't a good name?

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,115 @@
...
+        if (!$this->util->isSupportedVersion(PHP_VERSION)) {
+            return TaskResult::createFailed(
+                $this,
+                $context,
+                sprintf('PHP version %s is unsupported', PHP_VERSION)
+            );
+        }
+
+        // Check the project version if defined
+        if (array_key_exists('project', $config) !== null) {
```
veewee: There is no need for `array_key_exists` here. The `OptionsResolver` will make sure it exists since you default it to null.

src/GrumPHP/Util/PhpVersion.php
```diff
@@ -0,0 +1,56 @@
...
+     * @see https://secure.php.net/supported-versions.php for a list of currently supported versions
+     */
+    public function isSupportedVersion($currentVersion)
+    {
+        $versionIsSupported = false;
+        $now = new \DateTime();
+        foreach ($this->versions as $number => $eol) {
+            $eol = new \DateTime($eol);
+            if ($now < $eol && version_compare($currentVersion, $number) >= 0) {
+                $versionIsSupported = true;
```
veewee: You could return it early here which will make the code a bit shorter and easier to follow.
Maybe you can also use $now <= $eol so that you can set the end date to the correct date with hours 00:00:00?

test/src/GrumPHPTest/Util/PhpVersionTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace GrumPHPTest\Util;
+
+use GrumPHP\Util\PhpVersion;
+
+/**
+ * Class PhpVersionTest
+ * @package GrumPHPTest\Util
+ */
+class PhpVersionTest extends \PHPUnit_Framework_TestCase
```
veewee: Is there a reason why u added a phpunit test instead of a phpspec test?
This logic is perfectly testable with phpspec. I only want to use phpunit for more advanced integrations between multiple components. Otherwise it isn't clear when to use phpspec or phpunit.

spec/GrumPHP/Task/PhpVersionSpec.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    function let(GrumPHP $grumPHP, PhpVersionSpec $version)
+    {
+        $grumPHP->getTaskConfiguration('phpversion')->willReturn([]);
+        $this->beConstructedWith([]);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('GrumPHP\Task\PhpVersion');
```
veewee: Please use the `PhpVersion:class` syntax here.

spec/GrumPHP/Task/PhpVersionSpec.php
```diff
@@ -0,0 +1,57 @@
...
+
+    function it_should_have_a_name()
+    {
+        $this->getName()->shouldBe('phpversion');
+    }
+
+    function it_should_have_configurable_options()
+    {
+        $options = $this->getConfigurableOptions();
+        $options->shouldBeAnInstanceOf('Symfony\Component\OptionsResolver\OptionsResolver');
```
veewee: Please use the `OptionsResolver:class` syntax here.

spec/GrumPHP/Task/PhpVersionSpec.php
```diff
@@ -0,0 +1,57 @@
...
+    function it_should_run_in_run_context(RunContext $context)
+    {
+        $this->canRunInContext($context)->shouldReturn(true);
+    }
+
+    function it_runs_the_suite(PhpVersion $version, ContextInterface $context)
+    {
+        $version->isInstalled()->willReturn(true);
+        $result = $this->run($context);
+        $result->shouldBeAnInstanceOf('GrumPHP\Runner\TaskResultInterface');
```
veewee: Please use the `TaskResultInterface:class` syntax here.

test/src/GrumPHPTest/Util/PhpVersionTest.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    public function testValidVersion()
+    {
+        $now = new \DateTime();
+        $inOneYear = $now->add(new \DateInterval('P1Y'));
```
veewee: Please import the DateInterval class here.

veewee:
PHP 7.1 is also available now. It's supported untill 1 Dec 2019:
http://php.net/supported-versions.php

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * This methods specifies if a task can run in a specific context.
+     *
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    public function canRunInContext(ContextInterface $context)
+    {
+        return $context instanceof RunContext;
```
veewee: Can you also add the PreCommitContext? This way it is checked both during commit and in run modus.

src/GrumPHP/Task/PhpVersion.php
```diff
@@ -0,0 +1,114 @@
...
+     * @var GrumPHP
+     */
+    private $grumPHP;
+
+    /**
+     * PhpVersion constructor.
+     * @param GrumPHP $grumPHP
+     * @param \GrumPHP\Util\PhpVersion $phpVersionUtility
+     */
+    public function __construct(GrumPHP $grumPHP, \GrumPHP\Util\PhpVersion $phpVersionUtility)
```
veewee: Maybe you could use an alias so that you don't need the FQCN here?

Issue #152, opened 2016-06-07
Run grumphp in a docker container


jverdeyen:
| Q | A |
| --- | --- |
| Branch | master |
| Bug? | no |
| New feature? | no |
| Question? | yes |
| Documentation? | no |

In short:
I'm developing php applications in docker containers. It's a lightweight vagrant box with a docker server.
The OS X docker client connects with it as it was native.

This means:
- Everything related with git is on OS X cli.
- Everything related with composer and bin executables are on the PHP container.

I would like to extend grumphp in the way that the pre-commit hook command executes in the docker container.

Example to run grumphp using docker and docker-compose from OS X cli:

``` bash
docker exec -it `docker-compose ps -q php` bash -c "cd /var/www/app && ./vendor/bin/grumphp run";
```

I can make something like the vagrant pre-commit settings:
https://github.com/phpro/grumphp/blob/master/resources/hooks/vagrant/pre-commit

But I'll need some extra parameters, which container (`php`) and which docroot (`/var/www/app`).
The git diff can be done on the OS X cli but the run should happen on the container.

Can I add a `docker` parameter, with all settings?
Hooks can be installed local as default.

I'm currently using it by hardcoding this in the `pre-commit` hook.

```
# Run GrumPHP
#(cd "./" && echo "${DIFF}" | exec '/vendor/phpro/grumphp/bin/grumphp' 'git:pre-commit' '--skip-success-output')
(cd "./" && echo "${DIFF}" | docker exec -i `docker-compose ps -q php` bash -c "cd /var/www/app && ./vendor/bin/grumphp git:pre-commit --ansi --skip-success-output")
```

A gist:
https://gist.github.com/jverdeyen/9c0b247bb8b34c7fae514315db36a23e

I would like to hear feedback about this :)

Thanks!

veewee:
Hello @jverdeyen,

A docker preset sure sounds look a nice addition!
In your gist: how does the docker container get access to the commit message file? I don't think this will work but haven't tried it.

With the vagrant preset files, we tried to avoid setting additional parameters in the yaml file. We did this to make sure that the configuration was preset independent. The disadvantage of this approuch is that a home directory needs to be set in the vagrant box.

Maybe it's a better idea to provide something like a `git_hooks_variables` parameter which can be used to fill replacements. For example:

```yaml
parameters:
    git_hook_variables:
         DOCKER_CONTAINER: php
         DOCKER_HOME_DIRECTORY: /app/something
```

[...]

PR #232, opened 2016-11-27
Docker

pecavalheiro:
| Q             | A
| ------------- | ---
| Branch        | docker
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | [#152](https://github.com/phpro/grumphp/issues/152)

A base Docker image for running GrumPHP inside a container. Not sure what else is needed, beyond the basic functionalities. Any suggestions?

The image pre installs GrumPHP and its dependencies, and runs with current dir files (mapped inside the container). I also added the image to DockerHub, for easier testing.

Running:

```
docker run -ti -v $(pwd):/app pecavalheiro/grumphp run
```

Or you can build the image yourself

```
docker build -t grumphp .
docker run -ti -v $(pwd):/app grumphp run

```

Default command is `help`, accepts all grumphp's default commands/params.

Also, I added the Mcrypt extension, needed by some packages. Is there a list of all needed extensions, so I can add them to the Dockerfile?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM php:5.6-cli
+
+WORKDIR /tmp
+
+RUN apt-get update && apt-get install git zip php5-mcrypt -y && \
+    php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" && \
+    php composer-setup.php && \
+    php -r "unlink('composer-setup.php');" && \
+    mv composer.phar /usr/local/bin/composer && \
+    composer global require phpro/grumphp && \
+    composer global update phpro/grumphp && \
+    echo 'date.timezone = America/Sao_Paulo' > /usr/local/etc/php/conf.d/php.ini && \
+    echo 'memory_limit = -1' >> /usr/local/etc/php/conf.d/php.ini && \
+    echo 'extension=/usr/lib/php5/20131226/mcrypt.so' >> /usr/local/etc/php/conf.d/php.ini && \
+    apt-get clean && apt-get autoremove -y
+
+WORKDIR /app
+
+ENTRYPOINT ["/root/.composer/vendor/bin/grumphp"]
+CMD ["help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN apt-get update && apt-get install git zip php5-mcrypt -y && \
     mv composer.phar /usr/local/bin/composer && \
     composer global require phpro/grumphp && \
     composer global update phpro/grumphp && \
-    echo 'date.timezone = America/Sao_Paulo' > /usr/local/etc/php/conf.d/php.ini && \
+    echo 'date.timezone = UTC' > /usr/local/etc/php/conf.d/php.ini && \
     echo 'memory_limit = -1' >> /usr/local/etc/php/conf.d/php.ini && \
     echo 'extension=/usr/lib/php5/20131226/mcrypt.so' >> /usr/local/etc/php/conf.d/php.ini && \
     apt-get clean && apt-get autoremove -y


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM php:5.6-cli
```
veewee: Shouldn't we force the latest 7 version in the container?

veewee:
Hi @pecavalheiro,

Thanks for your PR. 

This PR currently only contains a Dockerfile to run grumphp in global mode.
We'll also need some presets to run the pre-commit and comit-msg command in the container.
You can take a look at the vagrant version here:

- https://github.com/phpro/grumphp/blob/master/resources/hooks/vagrant/pre-commit
- https://github.com/phpro/grumphp/blob/master/resources/hooks/vagrant/commit-msg
- https://github.com/phpro/grumphp/blob/master/doc/parameters.md  At the parameter: hooks_preset

Besides that, it should be possible to configure environment variables for the git hooks with sensible defaults:

```
parameters:
    git_hook_variables:
         DOCKER_CONTAINER: php
         DOCKER_CONTAINER_VERSION: php
         DOCKER_VOLUME_DIRECTORY: /app/something
```

This way, the users are free to use any container in any version they like. 
Maybe we could also implement these variables in a later stage, since we also want to use them for vagrant.

Since not everybody knows how to use or extend the container, we'll also need a documentation page about the container.

Currently we don't have a list of needed extensions and/or packages.

pecavalheiro:
Hi @veewee,

Ok, I'll take a look at those hooks, see what I can do :smile: 

About the custom parameters

```
parameters:
    git_hook_variables:
         DOCKER_CONTAINER: php
         DOCKER_CONTAINER_VERSION: 7
         DOCKER_VOLUME_DIRECTORY: /app/something
```

I think that the only way to allow the user to set his own container/php version is to dynamically generate a Dockerfile, which then should be built in the user's environment. This works, but does not sounds right for me, because you can't predict incompatibilities between containers and package versions. You solve a problem, but create another one. 
Maybe a better ideia is having a set of default containers, for each needed setup/version with a specifig tag, like most repos in DockerHub ([PHP DockerHub](https://hub.docker.com/_/php/)). What do you think?

veewee:
That doesn't sound like something I would like to maintain in this repository :)
The way I thought about it, was that users could specify the docker container they want to use and the grumphp command runs inside that container. Based on the environment variables they filled in, they could specify what they want to run in any container they would like to run it in.
But maybe we could start out with a simple version and work further on that one once it is finished?

pecavalheiro:
I agree that having a Dockerfile in this repository is hard to maintain. It is possible to allow the user to define his own container, where grumphp will run, but then he (the user) will be 100% responsible for building a working container (this means installing php, all necessary extensions, installing grumphp and defining all paths), so we can just call it from an external hook. To make this easiar, we can provide a base Dockerfile, like this PR, so the user can change it according to its needs. How about that?

veewee:
@pecavalheiro: That sounds good. We do need a starting point, so I have no problem adding a basic Docker file to this package as a starting point.

Bilge:
I do not think this belongs to this repository. If you want to maintain a Docker container for a specific library, create a new repository, don't push it to the library's repository because it is outside the scope of that library.

veewee:
@Bilge, I Agree: this is not something I would like to maintain. 
I do think that we should provide some basic git hook presets to work with docker.
We could also add some Dockerfiles in the documentation or host it in a separate repository.

I am running grumphp on bitbucket pipelines which are docker based. Non of the projects has the same container which makes it difficult to provide a sensible preset.

veewee:
I am closing this for now since we won't be maintaining a Dockerfile in this repository. If you want to add configurable git hooks for Dokcer, feel free to create a PR.

Issue #250, opened 2016-12-25
Impossible to run custom NPM scripts


Anahkiasen:
| Q               | A
| --------------- | ---
| Branch          | 0.11.0 
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

**My configuration**
```yaml
parameters:
    git_dir: .
    bin_dir: ./vendor/bin
    tasks:
        npm_script:
            script: run lint
            triggered_by: [js]
            working_directory: "./"
```

**Steps to reproduce:**

Run Grumphp

**Result:**

In NPM, you can add custom scripts to the `package.json` file. Custom being any scripts that are not core scripts (npm build, npm test, etc). To run those, you call `npm run {name of script}`.

Issue is currently Grumphp runs `npm 'run lint'` instead of `npm run lint`, which fails.

veewee:
Thanks @Anahkiasen for reporting this issue.
The script parameter is indeed handled as one argument.
Possible solutions would be:

- Create an additional `is_custom` or `is_run_task` parameter which will add the run before the `script` parameter in the CLI.
- Create an additional `arguments` parameter in which you can specify custom cli arguments.
- ... ?

Can you tell me which you like best?

Anahkiasen:
Could Grumphp recognize if the custom script starts with `"run "`, it's always either going to or not going to anyway no?

veewee:
That is also a possibility. However, I prefer to keep the configuration explicit. This way, the parameters can't be abused and won't contain any unpredictable logic. Maybe the boolean option is the best choice. As stated in the NPM docs, it is also possible to add custom attributes to a custom command. We could use an arguments parameter for that case.

veewee:
This issue is fixed in #251. Can you check if this is working for you?

PR #251, opened 2016-12-27
npm run is implemented in the npm_script task

bvanderstee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | https://github.com/phpro/grumphp/issues/250

Added an extra parameter 'is_run_task' in the npm_script task to make it possible to run custom npm scripts. eg. npm run lint.

### DIFF
--- a/src/GrumPHP/Task/NpmScript.php
+++ b/src/GrumPHP/Task/NpmScript.php
@@ -62,7 +62,7 @@ class NpmScript extends AbstractExternalTask
         }
 
         $arguments = $this->processBuilder->createArgumentsForCommand('npm');
-        $arguments->addOptionalArgument('run', $config['is_run_config']);
+        $arguments->addOptionalArgument('run', $config['is_run_task']);
         $arguments->addRequiredArgument('%s', $config['script']);
 
         $process = $this->processBuilder->buildProcess($arguments);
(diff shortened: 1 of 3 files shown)


src/GrumPHP/Task/NpmScript.php
```diff
@@ -59,7 +61,12 @@ public function run(ContextInterface $context)
             return TaskResult::createSkipped($this, $context);
         }
 
-        $arguments = $this->processBuilder->createArgumentsForCommand('npm');
+        $command = 'npm';
+        if (true === $config['is_run_task']) {
```
veewee: This will result in `'npm run'` on the CLI instead of `'npm' 'run'`. Can you use `$arguments->addOptionalArgument()` instead?

veewee:
Thanks for the PR. I've added one small remark.

src/GrumPHP/Task/NpmScript.php
bvanderstee: I've made the changes and added them to this PR.

veewee: @bvanderstee 

The `addOptionalArgument()` method is created to decrease the amount of if statements in the tasks. A better use-case in which you don't need the if-statement looks like this:

``
$arguments->addOptionalArgument('run',  $config['is_run_task']);
```

bvanderstee: Thanks for the tip, and PR is updated.

veewee:
Thanks @bvanderstee !

Issue #110, opened 2016-02-05
Allow php-cs-fixer to autofix


tacone:
Hello, I would like to propose to make `--dry-run` optional. I understand auto-fixing may not suit everyone's needs, but I find it terribly handy.

Before finding about grumphp I even put together a [simple package](https://github.com/tacone/psr-police) to accomplish that. I would be so happy to dump my homegrown solution for something like grumphp, but I know I'd miss the auto-fixing too much.

What do you think? It seems something very easy to accomplish, let me know if you can do it or you would like a pull request.

veewee:
Hi @tacone,

Thanks for propising this feature.

Personally I am not a big fan of manipulating code automatically right before a commit.
This way you don't have control about what you commit. Imagine that the command goes wrong and your files are totally messed up.

I also have 2 other concerns: 

- You will have to add the file again after the manipulations, but what about files with both staged and unstaged changes? This may result in unexpected behaviour.
- What about the priority of the task? If this task runs and changes files, this means that the tests that ran previously to the fix command could succeed, but after the automatic fix could fail. 

What are your views on the topics above?

tacone:
Hello @veewee, thank you for the thoughtful reply. You are right in what you say, and I agree with you that the dry run is a sensible choice and should be kept as default.

Regarding the concerns, it is possible even if unlikely that php-cs-fixer would mess up something. It is very unlikely that will change the actual behavior of the code or the result of the other tests.

My take on the whole matter is very simple: if GrumPHP will not let you commit without fixing, then you are going to fix anyways, automatically or not.

I agree with you that manually fixing gives you more control.

veewee:
Hi @tacone, 

I don't have a problem with the automatic fixing, but with the automatic `git add $file` of the code that is changed during validation. 
Maybe the files could be marked as changed in a central service. When all tasks ran, the question could be asked to the user: The phpcsfixer changed 6 files. Do you want to add them to the commit?
Maybe with an option to view the diff of the unstaged files at that point.

At that point the phpcsfixer task should not throw any exceptions, but it is recommended that the ouput of the task is shown in a non-blocking way. (See #17)

Not sure that when you ask a question in GUI tools, they will also prompt the question the the user. So this is something we will need to take a look at.

tacone:
Yes, I understood that.

The issues you raise are worth considering. I guess that the best short term solution to adjust for my workflow is just creating a custom task and live with it.

Thank you for now :)

ceesvanegmond:
[...]

PR #253, opened 2016-12-28
Add phpdoc support

nicolasleborgne:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | 

Added support for Phpdoc. It allowed to generate Phpdoc in pre-commit but also in run context.
When running in pre-commit context, generated doc is added to staging area right before commit.

# New Task Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ parameters:
         phpcs: ~
         phpcsfixer: ~
         phpcsfixer2: ~
+        phpdoc: ~
         phplint: ~
         phpmd: ~
         phpparser: ~
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -32,6 +32,7 @@ parameters:
         phpcs: ~
         phpcsfixer: ~
         phpcsfixer2: ~
+        phpdoc: ~
         phplint: ~
         phpmd: ~
         phpparser: ~
@@ -74,6 +75,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 - [Phpcs](tasks/phpcs.md)
 - [PHP-CS-Fixer](tasks/php_cs_fixer.md)
 - [PHP-CS-Fixer 2](tasks/php_cs_fixer2.md)
+- [Phpdoc](tasks/phpdoc.md)
 - [PHPLint](tasks/phplint.md)
 - [PhpMd](tasks/phpmd.md)
 - [PhpParser](tasks/phpparser.md)
--- a/doc/tasks/phpdoc.md
+++ b/doc/tasks/phpdoc.md
@@ -1,0 +1,191 @@
+# Phpdoc
+
+Phpdoc is the world standard auto-documentation tool for PHP. Written in PHP, phpdoc can be used directly from the command-line. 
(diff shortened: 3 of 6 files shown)


nicolasleborgne:
Hi, sorry for starting submitting this PR with some defects ... 

Now the code is working good, but CI tells to not use elseif or else statement, is this forbidden ? Does that mean I have to use only simple if statement ?

veewee:
Hi @nicolasleborgne,

Thanks for the PR! 

 We will not merge this one in since it is changing files before a commit.
This doesn't seem like something you would want to do since you don't have any control about what you are committing. We had a similar discussion in this issue: https://github.com/phpro/grumphp/issues/110

Luckily, your code doesn't have to go to waste. You can create your own GrumPHP extension or try to add it to an existing extension library like: https://github.com/wearejust/grumphp-extra-tasks

One way we could get this task in, is by checking if there are non committed files in the phpdoc directory. When there are changed and non-staged files, it just fails with a message like: 'It looks like your docs are not up-to-date'.

Feel free to ask any questions about this decision!

nicolasleborgne:
Hi @veewee, thanks for your reply, I understand your position.

I think I will try to add this version to the grumphp-extra-tasks repo you have mentioned.

I am also interested by your thought. I like the idea to tell the developer that docs are not up-to-date and fails the commit. But if we implement this, the developer will check the non-staged changes, staged the doc, and retry his commit. But as the first time, the commit will trigger the phpdoc generation, and we will have some new non-staged changes in the phpdoc folder, and commit will fail again, so the developer will be blocked.

Maybe we can manage that by triggering phpdoc only if we have in the commit :
* some php file in relation with the `directory` or `filename` phpdoc parameters
* no files from the phpdoc `target_folder`

What you think about that ?

veewee:
Hi @nicolasleborgne,

Ideal, there would be a dry-run modus for the phpdoc command which results in a list of doc files that will be altered. I am not too familiar with phpdoc, but I dont think that one exists.

Most of current tasks are simple proxies to the commandline tools. Logic like the one you described in previous post is possible, but I would like to keep the tasks as straight forward as possible.

nicolasleborgne:
Hi @veewee,

Yes, it would be nice. For now, this mode does not exist in phpdoc.

I completely understand that. Currently I keep this task as custom task in my projects and maybe I will try to add it to https://github.com/wearejust/grumphp-extra-tasks.

Anyway, thanks for your help :)

veewee:
Ok. Thanks for the PR anyways! Good luck on getting it merged in at the extra-tasks.
I'll close this one for now.

vyshkant:
AFAIU, this is not about "check for missing phpdoc-comments"? Is there a task to check if all methods and properties have a phpdoc comment?

nicolasleborgne:
As far as i know there is no grumphp task to do this. This task will just generate phpdocs. If that interests you anyway, you can find it on grumphp-extra-tasks.

PR #269, opened 2017-01-14
Removed class name duplicated into comments and package annotations.

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | n/a

Class names duplicated into class docblocks and namespaces duplicated in `@package` annotations are pure technical debt, providing no added value but producing maintenance burden to ensure they remain correct.

I presume nobody manually entered these comments so you probably want to look into reconfiguring your editor templates to prevent perpetuating this problem.

### DIFF
--- a/.phpspec/specification.tpl
+++ b/.phpspec/specification.tpl
@@ -6,9 +6,6 @@ use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 use %subject%;
 
-/**
- * Class %name%
- */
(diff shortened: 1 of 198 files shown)


veewee:
Thanks for the PR! I agree that it is unmaintainable technical debt. Let's get rid of it :)

PR #279, opened 2017-02-06
Add 'diff' option to phpcsfixer2 task

jhuet:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | no

PHP CS Fixer 2 has an option called `--diff` to visualize the diff of all the code violations it found. I added an option to enable the output of this diff when GrumPHP is executed. By default this option is disabled.

One note: i'm not sure why the [`--format=json`](https://github.com/phpro/grumphp/compare/master...jhuet:feature/phpcsfixer2-option-diff?expand=1#diff-5734ec63a7bc6aa3851c2701bfa89c70L66) was forced but it didn't allow the diff output to appear so i removed it.

### DIFF
--- a/doc/tasks/php_cs_fixer2.md
+++ b/doc/tasks/php_cs_fixer2.md
@@ -15,6 +15,7 @@ parameters:
             using_cache: true
             path_mode: ~
             verbose: true
+            diff: false
 ```
 
 
@@ -22,8 +23,8 @@ parameters:
 
 *Default: false*
 
-The allow_risky option allows you to set whether riskys fixer may run. 
-Risky fixer is a fixer, which could change code behaviour. 
+The allow_risky option allows you to set whether riskys fixer may run.
+Risky fixer is a fixer, which could change code behaviour.
 By default no risky fixers are run.
 
 
@@ -55,9 +56,9 @@ The rules option lets you choose the exact fixers to apply.
 
 *Default: true*
 
-The caching mechanism is enabled by default. 
-This will speed up further runs by fixing only files that were modified since the last run. 
-The tool will fix all files if the tool version has changed or the list of fixers has changed. 
+The caching mechanism is enabled by default.
+This will speed up further runs by fixing only files that were modified since the last run.
+The tool will fix all files if the tool version has changed or the list of fixers has changed.
 Cache is supported only for tool downloaded as phar file or installed via composer.
 
 
@@ -73,3 +74,9 @@ Specify path mode (can be override or intersection).
 *Default: true*
 
 Show applied fixers.
+
+**diff**
+
+*Default: false*
+
+Show the full diff that will be applied.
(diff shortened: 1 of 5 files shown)


veewee:
Thanks for the PR @jhuet,

The `--json` option was added for the custom `\GrumPHP\Formatter\PhpCsFixerFormatter` and can't be removed. Does it still work when the `--diff` option is added?
Can you add a test for this in the formatter?
Also: you are missing a test for the new option.

jhuet:
Alright, i didn't realize there was a formatter involved to handle the json.

Also, I couldn't find the test i was missing for the new option, which one are you thinking of ?

As for the move to PSR-4 i'll do it last, Composer wouldn't easily let me reconfigure the autoload so i couldn't bother doing it right now :)

veewee:
The new option should also be tested in `\spec\GrumPHP\Task\PhpCsFixerV2Spec::it_should_have_configurable_options()`. 
Can you show me an example of what the formatter looks like with and without the diff option?

spec/GrumPHP/Task/PhpCsFixerV2Spec.php
```diff
@@ -54,6 +54,7 @@ function it_should_have_configurable_options()
         $options->getDefinedOptions()->shouldContain('using_cache');
         $options->getDefinedOptions()->shouldContain('path_mode');
         $options->getDefinedOptions()->shouldContain('verbose');
+        $options->getDefinedOptions()->shouldContain('diff');
```
jhuet: @veewee this test ?

jhuet:
Here's an example with just 1 diff, the output might be really huge if many diffs are found : 

    Running task 1/1: PhpCsFixerV2
    1) /var/www/project/src/Project.php (elseif)

    --- Original
    +++ New
    @@ @@
                         $values[$serviceName] = $serviceConfig;
    -                } else if (isset($serviceConfig['factoryClassName'])) {
    +                } elseif (isset($serviceConfig['factoryClassName'])) {
    
    You can fix all errors by running following commands:
    '/var/www/project/vendor/bin/php-cs-fixer' '--allow-risky=no' '--using-cache=yes' '--verbose' '--diff' 'fix' 'src/Project.php'

spec/GrumPHP/Task/PhpCsFixerV2Spec.php
veewee: Doh! Yes indeed. I've looked over that one :)

veewee:
That's a valid point. But on the other hand: the diff option is disabled by default.
So I don't think the length of the message is really an issue.

For me it's good to go once you rebase against master.

veewee:
Thanks for the PR!

jhuet:
You're welcome :)

Issue #286, opened 2017-02-10
Add option to hide circumvention tip


Bilge:
| Q               | A
| --------------- | ---
| Branch          | n/a
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | n/a

When GrumPHP tasks fail, it outputs a line similar to the following.

>To skip commit checks, add -n or --no-verify flag to commit command

To discourage users from taking this approach I would appreciate a configuration parameter to disable this notice.

Issue #275, opened 2017-01-15
Phpcs and PhpLint tasks fail on Windows 7


Bilge:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | n/a

`Phpcs` and `PhpLint` tasks are skipped on Windows 7 due to "the windows cmd input limit". As I understand, this error message is generated by GrumPHP so it is this project's prerogative to devise a suitable workaround for running these tasks on Windows.

**My configuration**
Default.

**Steps to reproduce:**
1. `bin/grumphp run`

**Result:**
```
GrumPHP is sniffing your code!
Running task 1/7: PhpParser
Running task 2/7: Phpcs
Oh no, we hit the windows cmd input limit!
Skipping task ...

Running task 3/7: Phpspec
Running task 4/7: Phpunit
Running task 5/7: Composer
Running task 6/7: PhpLint
Oh no, we hit the windows cmd input limit!
Skipping task ...

Running task 7/7: YamlLint
             âââââââââââââââââââââââââ
           âââââââââââââââââââââââââââââ
         ââââââââââââââââââââââââââââââââ
        ââââââââââââââââââââââââââââââââââ
        ââââââââââââââââââââââââââââââââââ
  ââââââââââââââââââââââââââââââââââââââââ
 âââââââââââââââââââââââââââââââââââââââââ
 âââââââââââââââââââââââââââââââââââââââââ
   âââââââââââââââââââââââââââââââââââââââ
     ââââââââââââââââââââââââââââââââââââââââââ
      ââââââââââââââ                 ââââââââââ
     ââââââââ                         âââââââ
      âââââ      âââ          ââ       âââââ
        âââ                            âââ
         ââ           ââ ââ           ââââ
        âââ       ââââââââââââ       ââââ
         âââ ââââââââââââââââââââââââââââ
        âââââââââââââââââââââââââââââââââ
          ââââââââââââââââââââââââââââââ
         ââââââââââââââââââââââââââââââ
          âââââââââââââââââââââââââââââ
           ââ âââââââââââââââââââââââââ
            â âââââââââââââââââââââ ââ
                ââââââââââââââââââ
                 âââ  ââââââ  âââ
                        âââ   ââ
       _    _ _                         _ _
      / \  | | |   __ _  ___   ___   __| | |
     / _ \ | | |  / _` |/ _ \ / _ \ / _` | |
[...]

Issue #261, opened 2017-01-11
What is the easiest way to bypass GrumPHP?


Bilge:
| Q               | A
| --------------- | ---
| Question?       | yes

Although we would want most people to respect GrumPHP most of the time, exceptions must apply. What is the easiest way to temporarily bypass GrumPHP if it is not desired?

jyggen:
Use the `--no-verify` (`-n`) flag when you commit which bypasses the pre-commit and commit-msg hooks.

veewee:
As @jyggen mentioned, you can bypass GrumPHP by using the `-n` flag in git.
Another option is to remove the git hooks. There is also the command `git:deinit`.

the feedback on this issue will be added to the documentation. In the meantime, I created #265 to keep track on this change. I'll close this issue for now.

PR #289, opened 2017-02-11
Added hide_circumvention_tip parameter.

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #286

Adds a parameter to configure whether to show the Git commit hook circumvention tip when a task fails.

## Issues

One of the phpspec tests currently fails but I don't know phpspec so I don't understand why. Since I do not know phpspec I haven't written any new tests to cover this new feature; help wanted!

>----  broken examples
>
>        GrumPHP\Runner\TaskRunnerContext
>  40  ! knows to skip the success message
        no skipSuccessOutput([array:0]) matcher found for
        [obj:GrumPHP\Runner\TaskRunnerContext].

## Comments

I had difficulty even just testing this new feature manually. Although I can normally install GrumPHP, it does not work for the GrumPHP project itself for some reason.

>$ bin/grumphp git:init
>
>  [GrumPHP\Exception\RuntimeException]
  The executable for "grumphp" could not be found.

Furthermore GrumPHP does not have a feature to specify which files to check, so without being able to install the Git hook, it checks every single file which either takes a long time or fails completely due to #275. 

However, while coding I accidentally tripped a callisthenic check from the PhpParser task by writing an `else` statement, which allowed me to verify this feature does indeed seem to be working.

### DIFF
--- a/src/Console/Helper/TaskRunnerHelper.php
+++ b/src/Console/Helper/TaskRunnerHelper.php
@@ -41,7 +41,7 @@ class TaskRunnerHelper extends Helper
     /**
      * @return PathsHelper
      */
-    private function paths()
+    private function getPathsHelper()
     {
         return $this->getHelperSet()->get(PathsHelper::HELPER_NAME);
     }
@@ -65,23 +65,30 @@ class TaskRunnerHelper extends Helper
         }
 
         $taskResults = $this->taskRunner->run($context);
-
         $warnings = $taskResults->filterByResultCode(TaskResult::NONBLOCKING_FAILED);
+
         if ($taskResults->isFailed()) {
             $failed = $taskResults->filterByResultCode(TaskResult::FAILED);
-            return $this->returnErrorMessages($output, $failed->getAllMessages(), $warnings->getAllMessages());
+            $this->printErrorMessages($output, $failed->getAllMessages(), $warnings->getAllMessages());
+
+            $context->shouldHideCircumventionTip() || $this->printCircumventionTip($output);
+
+            return self::CODE_ERROR;
         }
 
-        if ($context->isSkipSuccessOutput()) {
-            $this->returnWarningMessages($output, $warnings->getAllMessages());
-            return self::CODE_SUCCESS;
+        if ($context->shouldSkipSuccessOutput()) {
+            $this->printWarningMessages($output, $warnings->getAllMessages());
         }
 
-        return $this->returnSuccessMessage($output, $warnings->getAllMessages());
+        if (!$context->shouldSkipSuccessOutput()) {
+            $this->printSuccessMessage($output, $warnings->getAllMessages());
+        }
+
+        return self::CODE_SUCCESS;
     }
 
     /**
-     * @param OutputInterface  $output
+     * @param OutputInterface $output
      */
     private function registerEventListeners(OutputInterface $output)
     {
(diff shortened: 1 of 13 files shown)


src/Console/Helper/TaskRunnerHelper.php
```diff
@@ -65,23 +65,30 @@ public function run(OutputInterface $output, TaskRunnerContext $context)
...
 
-        if ($context->isSkipSuccessOutput()) {
-            $this->returnWarningMessages($output, $warnings->getAllMessages());
-            return self::CODE_SUCCESS;
+        if ($context->shouldSkipSuccessOutput()) {
+            $this->printWarningMessages($output, $warnings->getAllMessages());
         }
 
-        return $this->returnSuccessMessage($output, $warnings->getAllMessages());
+        if (!$context->shouldSkipSuccessOutput()) {
```
Bilge: I wanted to write this as an `else` statement, but the PhpParser task wouldn't let me!

src/Console/Helper/TaskRunnerHelper.php
```diff
@@ -65,23 +65,30 @@ public function run(OutputInterface $output, TaskRunnerContext $context)
...
+
+            $context->shouldHideCircumventionTip() || $this->printCircumventionTip($output);
+
+            return self::CODE_ERROR;
         }
 
-        if ($context->isSkipSuccessOutput()) {
-            $this->returnWarningMessages($output, $warnings->getAllMessages());
-            return self::CODE_SUCCESS;
+        if ($context->shouldSkipSuccessOutput()) {
```
Bilge: I added this guard to maintain the same logic, but reading it does not make sense to me. Surely the `shouldSkipSuccessOutput` flag should only control whether or not we `printSuccessMessage`. Why does it determine whether we print warnings?

veewee:
I was thinking: maybe we should hide the tip by default. But then again, nobody would probably enable it. Maybe we should just remove the tip? It doesn't provide any additional value to the project.

veewee:
Strange: it should just register the git hook during installation. It might have something to do with #275. I am not using Windows myself, so I'll have to dive into that one.

Bilge:
@veewee I think it does provide value to the project because I asked the very same question (#261) when I was starting out. However, I would not have gotten as far as a failing run because I wanted to know this before I start using it, therefore it should be in the documentation.

However, even if it were clearly present in the documentation I do not think that necessarily precludes it from being printed at the end of a failed test run. This is still useful for those who miss it in the documentation or forget about it until they have a failure. For me, personally, I want to disable this notification to discourage my team for bypassing the checks, but that is my specific use-case, so I think an option is appropriate. 

Nevertheless, if it is your decision to drop the line entirely I would find that satisfactory, too.

veewee:
For us, the git hooks are a first line of defense against uwanted code. Since everybody can bypass it, we always set GrumPHP up in combination with CI tools. That way, we know if we can merge a specific branch in or not. It is always possible that somebody wants to push some code at the end of the day which isn't fully formatted or tested; That is OK, as long as it is fixed by the time it has to be merged in.

That is why we leave the option at the end of the error message. Even though the error message will discourage some people, it gives that people more flexibility on how to use the tool. As you mentioned before: you also wanted to know how to bypass it.

So maybe we'll just leave it as is and close this one for now?

Bilge:
>So maybe we'll just leave it as is and close this one for now?

Why would that ever be a good solution? As mentioned, I'm agreeable to either removing the tip completely or adding an option to hide it. I am certainly not agreeable to leaving it in without any options to hide it!

veewee:
Hi @Bilge,

I took the liberty of changing some parts of the code. Since this is just a simple configuration flag, I did not want to litter e.g. the TaskRunnerContext. By directly injecting the config object into the helper we create as little overhead as possible.

src/Console/Helper/TaskRunnerHelper.php
```diff
@@ -97,55 +98,63 @@ private function registerEventListeners(OutputInterface $output)
...
     /**
      * @param OutputInterface $output
-     * @param array $errorMessages
-     * @param array $warnings
+     * @param array           $errorMessages
+     *
+     * @return int
      */
-    private function printErrorMessages(OutputInterface $output, array $errorMessages, array $warnings)
+    private function returnErrorMessages(OutputInterface $output, array $errorMessages, array $warnings)
```
Bilge: Do you really think this is a better name? This method does not return error messages, it returns an error code. I believe *printErrorMessages* is still the correct name for this method, and moreover, the decision about the code should be left to the caller.

src/Console/Helper/TaskRunnerHelper.php
```diff
@@ -97,55 +98,63 @@ private function registerEventListeners(OutputInterface $output)
...
             $output->writeln('<fg=red>' . $errorMessage . '</fg=red>');
         }
+
+        if (!$this->config->hideCircumventionTip()) {
+            $output->writeln(
+                '<fg=yellow>To skip commit checks, add -n or --no-verify flag to commit command</fg=yellow>'
+            );
+        }
+
+        return self::CODE_ERROR;
```
Bilge: I think it is a better design to let the caller manage the return code since this method's return value is invariant otherwise.

veewee:
@Bilge, I've just left the old naming for now. Sure print sounds better, but it was added to make the run a bit more compact. I agree that it could be optimized. Feel free to PR.

PR #290, opened 2017-02-12
Remove task field shadowing

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | n/a

Removes field shadowing from tasks and replaces them with `@property` annotations to facilitate code completion for the derived type of the inherited field.

### DIFF
--- a/src/Task/JsonLint.php
+++ b/src/Task/JsonLint.php
@@ -10,13 +10,11 @@ use GrumPHP\Task\Context\GitPreCommitContext;
 use GrumPHP\Task\Context\RunContext;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
+/**
+ * @property JsonLinter $linter
+ */
 class JsonLint extends AbstractLinterTask
 {
-    /**
-     * @var JsonLinter
-     */
-    protected $linter;
-
     /**
      * @return string
      */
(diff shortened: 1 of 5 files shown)


veewee:
@Bilge, I've just tested this and I do not get any property-specific code completion in Phpstorm. It only recognizes the interface. What IDE are you using?

Bilge:
I use PhpStorm 2016.3.2 (latest). Perhaps it is because I also have the [*PHP Annotations* plugin](https://github.com/Haehnchen/idea-php-annotation-plugin) installed. I recommend you install this plugin anyway.

veewee:
Strange. I've got the plugin installed like forever and I am also using the same PhpStorm version. I do not get the autocompletion for the implementation specific API.
Can you verify that e.g. in the `YamlLinter` you get auto completion for `setExceptionOnInvalidType()`?

Bilge:
Precisely how sure do you need me to be, exactly?

sure im sure

veewee:
That was the sureness I was looking for :)
I would love to get this autocomletion working on my machine before merging it in.

Bilge:
I would love to help but it's difficult to diagnose why it is not working for you, remotely. According to [this post](https://intellij-support.jetbrains.com/hc/en-us/community/posts/207067895-Declare-class-property-with-phpdoc) the feature has been available natively since 2011.

veewee:
All of a sudden, it works! Thanks for the PR

Issue #288, opened 2017-02-10
PHP-CS-Fixer rule parameter support


Bilge:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | FriendsOfPHP/PHP-CS-Fixer#2423

PHP-CS-Fixer just tagged release [2.10](https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.1.0) in which FriendsOfPHP/PHP-CS-Fixer#2494 was merged adding support for JSON rule configurations. This is significant because this is the only way to specify configuration parameters for rules; otherwise you can only specify the rule itself and must live with its defaults, which may not be what you want.

GrumPHP will need to include specific support for this feature to produce the correctly formatted JSON string for passing to the rule parameter.

This is very important because it precludes the need to write a configuration file just to specify rule parameters. Writing a PHPCS configuration is incredibly inconvenient not just because you must duplicate all fixer settings in that configuration, but also because that configuration seems to require specifying the paths it should search because it does not automatically respect the paths passed on the command line.

Bilge:
Closing because #292 is merged.

PR #292, opened 2017-02-12
Added JSON rules support for PHP-CS-Fixer v2.1.

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #288

Adds JSON rule support for PHP-CS-Fixer v2.1. To take advantage of this, the rules should be specified as a map instead of a list in the YAML configuration. However, to maintain backwards compatibility, list style is still supported, in which case rules will be comma-delimited as before.

## Issues
- [x] ~~Since I do not understand phpspec there are no tests. Please help by adding some! Feel free to commit directly to my branch.~~ Tests is dunned.
- [x] ~~There may be an issue where rules are specified as a map and any of the keys or values contains the single quote character (`'`) because it may escape the shell argument quoting. Single quotes may need to be escaped.~~ ~~I added a commit to address this issue by escaping single quotes. However, all that code may be unnecessary if the Symfony console layer handles escaping for us. Investigation needed.~~ Custom escaping is not needed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,7 @@
         "composer/composer": "^1.0",
         "doctrine/collections": "~1.2",
         "gitonomy/gitlib": "~1.0",
+        "phptype/array": "^1",
         "monolog/monolog": "~1.17",
         "ocramius/proxy-manager": "~0.4|~1.0|~2.0",
         "seld/jsonlint": "~1.1",
@@ -24,7 +25,7 @@
         "symfony/yaml": "~2.7|~3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "~1|~2",
+        "friendsofphp/php-cs-fixer": "^1|^2.1",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
         "nikic/php-parser": "~2.1",
         "phpspec/phpspec": "^3.2.2",
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
         "composer/composer": "^1.0",
         "doctrine/collections": "~1.2",
         "gitonomy/gitlib": "~1.0",
-        "phptype/array": "^1",
         "monolog/monolog": "~1.17",
         "ocramius/proxy-manager": "~0.4|~1.0|~2.0",
         "seld/jsonlint": "~1.1",
@@ -25,7 +24,7 @@
         "symfony/yaml": "~2.7|~3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^1|^2.1",
+        "friendsofphp/php-cs-fixer": "~1|~2",
         "jakub-onderka/php-parallel-lint": "^0.9.2",
         "nikic/php-parser": "~2.1",
         "phpspec/phpspec": "^3.2.2",
--- a/src/Task/PhpCsFixerV2.php
+++ b/src/Task/PhpCsFixerV2.php
@@ -73,12 +73,12 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
         $arguments->addOptionalArgument('--cache-file=%s', $config['cache_file']);
         $arguments->addOptionalArgument('--config=%s', $config['config']);
 
-        $rules = $config['rules'];
-        // JSON-encode rules if specified as a map.
-        if (array_values($rules) !== $rules) {
-            $arguments->addOptionalArgument('--rules=%s', json_encode($rules));
-        } else {
-            $arguments->addOptionalCommaSeparatedArgument('--rules=%s', $rules);
+        if ($rules = $config['rules']) {
+            $arguments->add(sprintf(
+                '--rules=%s',
+                // Comma-delimit rules if specified as a list; otherwise JSON-encode.
+                array_values($rules) === $rules ? implode(',', $rules) : json_encode($rules)
+            ));
         }
 
         $arguments->addOptionalArgument('--using-cache=%s', $config['using_cache'] ? 'yes' : 'no');
--- a/src/Task/PhpCsFixerV2.php
+++ b/src/Task/PhpCsFixerV2.php
@@ -75,7 +75,7 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
         $arguments->addOptionalArgument('--config=%s', $config['config']);
 
         if (is_array($rules = $config['rules']) && ArrayType::isMap($rules)) {
-            $arguments->addOptionalArgument('--rules=\'%s\'', json_encode($rules));
+            $arguments->addOptionalArgument('--rules=\'%s\'', str_replace('\'', '\'\\\'\'', json_encode($rules)));
         } else {
             $arguments->addOptionalCommaSeparatedArgument('--rules=%s', $rules);
         }
--- a/src/Task/PhpCsFixerV2.php
+++ b/src/Task/PhpCsFixerV2.php
@@ -5,7 +5,6 @@ namespace GrumPHP\Task;
 use GrumPHP\Runner\TaskResult;
 use GrumPHP\Task\Context\ContextInterface;
 use GrumPHP\Task\Context\RunContext;
-use ScriptFUSION\Type\ArrayType;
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -9,6 +9,7 @@
         "composer/composer": "^1.0",
         "doctrine/collections": "~1.2",
         "gitonomy/gitlib": "~1.0",
+        "phptype/array": "^1",
```
veewee: Is this additional dependency really necessary?

composer.json
```diff
@@ -24,7 +25,7 @@
         "symfony/yaml": "~2.7|~3.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "~1|~2",
+        "friendsofphp/php-cs-fixer": "^1|^2.1",
```
veewee: Does this mean that it won't work for 2.0?

src/Task/PhpCsFixerV2.php
```diff
@@ -72,7 +73,13 @@ public function run(ContextInterface $context)
         $arguments->addOptionalArgument('--allow-risky=%s', $config['allow_risky'] ? 'yes' : 'no');
         $arguments->addOptionalArgument('--cache-file=%s', $config['cache_file']);
         $arguments->addOptionalArgument('--config=%s', $config['config']);
-        $arguments->addOptionalCommaSeparatedArgument('--rules=%s', $config['rules']);
+
+        if (is_array($rules = $config['rules']) && ArrayType::isMap($rules)) {
```
veewee: `rules` is always an array as enforced by the options resolver.
Can you show me some example on how to configure the rules parameter?

composer.json
Bilge: Ah, this is an error. At first I thought I would not be able to support the old list-style rule configuration, but then I wrote the array helper library (`phptype/array`) to specifically support backwards-compatibility, so this change can be reverted! Nice spot :+1:

composer.json
Bilge: Yes. I wrote it specifically for this PR! It provides the `ArrayType::isMap` used later in this PR.

src/Task/PhpCsFixerV2.php
Bilge: Perhaps we can remove the array check in that case. Since this is a cutting edge feature they have not provided any documentation yet, so the best I can do is link the [PR that introduced the change](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2494), which I also linked from the issue requesting this feature (#288).

composer.json
jyggen: Installing `phptype/array` (and through that `scriptfusion/static-class`) for one line of code seems a bit overkill to me. Using `array_values($rules) !== $rules` directly is way easier :)

Bilge: The trade-off is loss of clarity and single-responsibility. However, if it is preferred, I can make the change you suggest and substitute that code directly in lieu of this dependency.

src/Task/PhpCsFixerV2.php
```diff
@@ -72,7 +72,15 @@ public function run(ContextInterface $context)
         $arguments->addOptionalArgument('--allow-risky=%s', $config['allow_risky'] ? 'yes' : 'no');
         $arguments->addOptionalArgument('--cache-file=%s', $config['cache_file']);
         $arguments->addOptionalArgument('--config=%s', $config['config']);
-        $arguments->addOptionalCommaSeparatedArgument('--rules=%s', $config['rules']);
+
+        $rules = $config['rules'];
+        // JSON-encode rules if specified as a map.
+        if (array_values($rules) !== $rules) {
+            $arguments->addOptionalArgument('--rules=%s', json_encode($rules));
+        } else {
```
Bilge: Object callisthenics, or whateverthefuck, complains about this. I am not used to being forbidden from writing `else` statements. How "else" should I write this?

veewee: I think the easiest thing to do is something like this:

```php
        if (count($rules)) {
            $arguments->add(sprintf(
                '--rules=%s', 
                array_values($rules) === $rules ? implode(',', $rules) : json_encode($rules)
            ));
        }
```

The current implementation of the json part will always add an empty array. Even though there are no rules configured. Since this is a pretty custom argument, it is OK not to use the `ProcessArgumentsCollection`.

Bilge: Why is a ternary statement better than an else statement?

veewee: It's all about the mental map of the code. In one look, you can tell that if you got rules, you want to add them as an argument. Independently of the value, you know that you want to add an additional parameter. There is one single path, one single responsibility of the if statement.
Agree: for the actual value, you'll need the ternary statement. But that is just one line of logic.

However, If you use the else statement. You'll have to check 2 paths for wath the code is doing + you must find out what the 2 argument functions are doing exactly. In your case, you'll also have to add an additional if or ternary statement to check if the array is empty sinct the json function will always create the string '[]';

Bilge: I can't say I'm in complete agreement. As a consequence of "calisthenics" we've duplicated the comma-delimiting code. I'd take the mental depression of having to comprehend an `else` statement over outright code duplication in any of my projects, but this is your project, and I respect your decision.

Bilge:
Tests are passing and docs are written. Only thing left is to add some new tests for this new feature.

veewee:
Hi @Bilge,

In other tasks we are currently not validating all these different possible options.
There is also no easy way to validate if the code is working since the logic is located in a bigger public function.

This is the easiest I could get in phpspec for the normal rules:

```php
    function it_handles_a_list_of_rules(
        GrumPHP $grumPHP,
        ProcessBuilder $processBuilder,
        Process $process,
        RunContext $context,
        PhpCsFixerFormatter $formatter
    )
    {
        $formatter->resetCounter()->shouldBeCalled();
        $grumPHP->getTaskConfiguration('phpcsfixer2')->willReturn([
            'config' => '.php_cs',
            'rules' => ['rule1', 'rule2'
        ]]);
        $context->getFiles()->willReturn(new FilesCollection([new SplFileInfo('file2.php', '.', 'file2.php')]));

        $processBuilder->createArgumentsForCommand('php-cs-fixer')->willReturn(new ProcessArgumentsCollection());
        $processBuilder->buildProcess(Argument::that(function (ProcessArgumentsCollection $args) {
            return $args->contains('--rules=rule1,rule2');
        }))->willReturn($process);

        $process->run()->shouldBeCalled();
        $process->isSuccessful()->willReturn(true);

        $this->run($context)->isPassed()->shouldBe(true);
    }
```

You'll also want to add a similar test for the json encoded data.

src/Task/PhpCsFixerV2.php
veewee: Just to get some closure on the subject: this is how I see it: Adding the `implode()` method is not duplicating code.

The actual problem you want to solve is: When there are rules configured, I want to add them to the phpcsfixer cli tool.
A sub problem arises: I want to format the value in a different way based on the type: map or vector.

You want to solve the actual problem by solving the sub-problem in which you can use 2 different methods that are already existing.
Since there is no method available on the arguments collection that formats the string as a comma separated value for maps and json ecoded for vectors, you'll have to call 2 different methods on the arguments collection. This means that there are 2 possible paths in the task.

If you look at the problem from the "actual problem" point of view, you only want one path: When there are rules, I want to add an argument to the collection.
Since there is no method that can format based on map or vector, you'll have to write a new argument formatting implementation. This implementation is so custom, that there is no need for an extra method in the arguments collection. So you are not duplicating code, but you are writing a new implementation for the formatting problem. Sure, internally it uses a duplication of implode(), but that is only one part of the new formatting implementation.

Bilge:
@veewee You can commit directly to my branch, but I guess since you pasted the code you don't mind if I commit it myself? I will see if I can base a JSON test on your sample.

veewee:
Great! Thanks for the work!

PR #298, opened 2017-02-15
Changed issue template "branch" to "version"

Bilge:
*Branch* doesn't make sense for issues. I have replaced it with the version because we might want to know which version of the program the user is using instead.


veewee:
You are right. Thanks!

Issue #294, opened 2017-02-14
GrumPHP relative config path doesn't work


Bilge:
| Q               | A
| --------------- | ---
| Branch          | v0.11.1  
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | n/a

Given that my composer.json contains the following

```json
"extra": {
      "grumphp": {
          "config-default-path": "config/grumphp.yml"
      }
},
```

When I run `git commit -a` I see the following error.

>$ git commit -a
>
>Fatal error: Uncaught exception 'Symfony\Component\Config\Exception\FileLocatorFileNotFoundException' with message 'The file "config/grumphp.yml" does not exist (in: D:\Users\f\g\h\api\vendor\phpro\grumphp\src\GrumPHP\Configuration/../../../resources/config).' in D:\Users\f\g\h\api\vendor\symfony\config\FileLocator.php on line 71
>
>Symfony\Component\Config\Exception\FileLocatorFileNotFoundException: The file "config/grumphp.yml" does not exist (in: D:\Users\f\g\h\api\vendor\phpro\grumphp\src\GrumPHP\Configuration/../../../resources/config). in D:\Users\f\g\h\api\vendor\symfony\config\FileLocator.php on line 71
>
>Call Stack:
    0.0005     236784   1. {main}() D:\Users\f\g\h\api\vendor\phpro\grumphp\bin\grumphp:0
    0.0190    2792784   2. GrumPHP\Console\Application->__construct() D:\Users\f\g\h\api\vendor\phpro\grumphp\bin\grumphp:30
    0.0200    2978760   3. GrumPHP\Console\Application->getContainer() D:\Users\f\g\h\api\vendor\phpro\grumphp\src\GrumPHP\Console\Application.php:57
    0.2620    5946456   4. GrumPHP\Configuration\ContainerFactory::buildFromConfiguration() D:\Users\f\g\h\api\vendor\phpro\grumphp\src\GrumPHP\Console\Application.php:158
    0.4965    7964888   5. Symfony\Component\DependencyInjection\Loader\YamlFileLoader->load() D:\Users\f\g\h\api\vendor\phpro\grumphp\src\GrumPHP\Configuration\ContainerFactory.php:53
    0.4965    7964936   6. Symfony\Component\Config\FileLocator->locate() D:\Users\f\g\h\api\vendor\symfony\dependency-injection\Loader\YamlFileLoader.php:67
>
[...]

PR #299, opened 2017-02-15
Removed GrumPHP config parameter from git hooks

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #294

There is no need to pre-bake the config path into the Git hook because GrumPHP reads the config path out of `composer.json` at runtime. Furthermore, baking the relative path actually causes GrumPHP to fail to find the config.

### DIFF
--- a/src/Console/Command/Git/InitCommand.php
+++ b/src/Console/Command/Git/InitCommand.php
@@ -4,7 +4,6 @@ namespace GrumPHP\Console\Command\Git;
 
 use GrumPHP\Configuration\GrumPHP;
 use GrumPHP\Console\Helper\PathsHelper;
-use GrumPHP\Exception\FileNotFoundException;
 use GrumPHP\Util\Filesystem;
 use RuntimeException;
 use SplFileInfo;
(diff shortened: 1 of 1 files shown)


veewee:
As I mentioned before: this is not the solution to the problem you are facing.

Bilge:
It certainly is *a* solution to the problem. Empirically, I have verified it works and cannot find any problems. If this does not work, what exactly doesn't work? Certainly we know that whatever it is, tests are no more aware of it than I am.

Bilge:
@veewee I have forked GrumPHP for my organization, but once this is merged, we can switch back to mainline GrumPHP. I'm sure you hadn't forgotten, but just in case, please don't forget about this one!

veewee:
Hi @Bilge,

I am working om some fixes regarding the installation on exotic paths. Hopefully more information in some hours ;)

veewee:
@Bilge, I am having some serious difficulties in getting this ticket right.
```
.
├── .git
├── grumphp.yml
├── somefile
└── src
    ├── composer.json
    ├── composer.lock
    └── vendor
```

Currently this is the hardest 'exotic' structure to take care of. The problem is getting al paths relative and get the command working in both the root as the src folder + getting the hooks right.
I'll have to investige how I can get all the cases right. But dropping realpath isn't an option for the moment.

Bilge:
Thanks for the update. This is the only outstanding issue preventing us from switching back to the original GrumPHP distribution.

Issue #297, opened 2017-02-15
testsuite option for phpunit task


cdekok:
it would be nice if there was a testsuite option to only run a certain suite in phpunit.
My usecase would be to skip integration tests on every commit.

veewee:
Thanks for reporting @cdekok,

That is indeed a missing option. I'dd accept a PR for that.

PR #303, opened 2017-02-20
Phpunit suite

cdekok:
| Q             | A
| ------------- | ---
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #297 

Add a testsuite option to the phpunit task

### DIFF
--- a/doc/tasks/phpunit.md
+++ b/doc/tasks/phpunit.md
@@ -9,6 +9,7 @@ parameters:
     tasks:
         phpunit:
             config_file: ~
+            suite: ~
             group: []
             always_execute: false
 ```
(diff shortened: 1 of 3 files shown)


cdekok:
it looks like appveyor is failing on something called `Chocolatey` seems to be their problem.

veewee:
Thanks for the PR! Yes, I've also noticed it. I'll try to take a look at it at the end of the week. The code looks good, one small remark: why not use `testsuite` as property name? This way it has the same naming as in phpunit.

cdekok:
You are right there was not really a reason only i liked it better with group ;) but better to keep the same name convention as the phpunit parameters

Issue #248, opened 2016-12-21
Add Phpstan task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | yes
| Related tickets | 

https://medium.com/@ondrejmirtes/phpstan-2939cd0ad0e3#.kurelsdfc
https://github.com/phpstan/phpstan

> PHPStan focuses on finding errors in your code without actually running it. It catches whole classes of bugs even before you write tests for the code.
> PHPStan moves PHP closer to compiled languages in the sense that the correctness of each line of the code can be checked before you run the actual line.

PR #306, opened 2017-02-20
Add PHPStan task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #248

PHPStan task. When in `GitPreCommitContext` it'll filter the result and only report errors for files that are actually being commited. 

Currently it parses the results table with regexp (since PHPStan don't support different output formats), which have worked flawlessly for me for a while but probably is a bit too hacky to be in an "official" GrumPHP task. 

Luckily there's a PR in progress (https://github.com/phpstan/phpstan/issues/132), so either we wait for it to be merged and rewrite some code or just remove the filtering all together if we rather always report all files :)

# New Task Checklist:

### DIFF
--- a/doc/tasks/phpstan.md
+++ b/doc/tasks/phpstan.md
@@ -1,0 +1,39 @@
+# PHPStan
+
+The PHPStan task focuses on finding errors in your code without actually running it. It catches whole classes of bugs even before you write tests for the code.
+It lives under the `phpstan` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        phpstan:
+            autoload_file: ~
+            configuration: ~
+            level: 0
+            triggered_by: ['php']
+```
+
+**autoload_file**
+
+*Default: ~*
+
+With this parameter you can specify the path your project's additional autoload file path.
+
+**configuration**
+
+*Default: ~*
+
+With this parameter you can specify the path your project's configuration file.
+
+**level**
+
+*Default: 0*
+
+With this parameter you can set the level of rule options - the higher the stricter.
+
+**triggered_by**
+
+*Default: [php]*
+
+This is a list of extensions to be sniffed.
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ To make GrumPHP even more awesome, it will suggest installing some extra package
 - phing/phing : ~2.0
 - sstalle/php7cc : ~1.1
 - phpspec/phpspec : ~2.1
+- phpstan/phpstan : ^0.6
 - phpunit/phpunit : ~4.5
 - roave/security-advisories : dev-master@dev
 - sebastian/phpcpd : ~2.0
@@ -121,8 +122,9 @@ parameters:
         phpmd: ~
         phpparser: ~
         phpspec: ~
+        phpstan: ~
         phpunit: ~
-        phpversion: ~        
+        phpversion: ~
         robo: ~
         securitychecker: ~
         shell: ~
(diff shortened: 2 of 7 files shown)


jyggen:
Not sure I can fix the current build error without making the passing environments fail instead. Catch 22 :)

```
FILE: /home/travis/build/phpro/grumphp/src/Task/PhpStan.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 116 | ERROR | [x] Expected 0 spaces before opening parenthesis; 1
     |       |     found
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

jyggen:
Took the liberty of bumping `squizlabs/php_codesniffer` to `~2.4`.

veewee:
@jyggen, Thanks for the PR!
If I understand it correctly, there is no way of passing which files should be analyzed? Therefor you added a custom filter to the output? 
Do you have any idea in which timeframe the formatter will be added to phpstan?

Ideally, we should just be able to pass the changed files to the tool. That way we do not need any filtering. If they are planning to add support for both a list of files and a formatter in short term, I think it is better to wait for those changes in the main package. As the current project version states, it is still a work in process.

If they are not planning to support is, it might be better to move the formatting code to a custom formatter. That way you can add some edge test-cases.

Can you give me some additional information on the subject? Thanks!

jyggen:
Yes, you can give PHPStan specific files as well.

I thought sending in specific files to PHPStan would prevent it from blooming into its full potential since, without analysing the files with classes and whatnot that your changed file depends on, it won't be able to detect some issues (and _might_ complain that the class it depends on isn't even defined). However, this behaviour does not affect 3rd-party dependencies available through Composer's autoloader since PHPStan automatically analyses them - which makes me wonder if maybe PHPStan actually can autoload your referenced application classes as well. I need to verify this, because if that's the case it would solve a lot.

The reason why I think we need to send all files to PHPStan is because it's possible to get the behaviour we want (parse everything but only analyse some) by using a configuration file which have more advanced options than what's available as CLI arguments and options.

I'll look into this further during the weekend :)

Fuco1:
@jyggen Actually, PHPStan will load all files that are required for analysis and it makes no difference for that one particular file if you specify all the rest or not.  So you can safely just pass it one file.

If you use standard autoloading (e.g. via composer) there's nothing to do, otherwise you can add a bootstrap script to `phpstan.neon` which will set up your own custom autoloaders.

Also, this PR https://github.com/phpstan/phpstan/pull/132 was resolved so you can do whatever you planned to do with it :)

jyggen:
@Fuco1 Great, thanks! :)

veewee:
Hi @jyggen,

If i understand the discussion correctly, the filterResult part can be dropped and the files should be passed as arguments during pre-commit context?

jyggen:
Yep! I'll try to finish up this PR later today.

jyggen:
@veewee Should be ready to go now! :+1:

doc/tasks/phpstan.md
```diff
@@ -0,0 +1,39 @@
...
+
+**configuration**
+
+*Default: ~*
+
+With this parameter you can specify the path your project's configuration file.
+
+**level**
+
+*Default: 0*
```
veewee: The default is set to 5 in the task. Which one is correct?

jyggen: Ah, my bad. 0 is correct. Should be fixed now.

veewee:
Looks good! Thanks!

Issue #169, opened 2016-08-01
How can I run shell command?


sc0rp10:
| Q | A |
| --- | --- |
| Branch | master |
| Bug? | yes |
| Question? | yes |

I need to run some console command on pre-commit stage. e.g. `./bin/console translation:check-catalogue --env=dev -vvv`
Now I unable to do it: I got `/bin/sh: ./bin/console translation:check-catalogue --env=dev: No such file or directory` error.
I can create shell file with single line for this, but it is not comfortable.
How can I do it?

veewee:
Currently that is not possible.
The process builder is translating this to:

```sh
'sh' './bin/console translation:check-catalogue --env=dev'
```

As you can see, there are single quotes around all the input. This means that it is handled as one argument.
The code of the Shell task could be improved to make sure that multiple arguments.

bvanderstee:
I'll be working on this issue

PR #314, opened 2017-02-24
Feature/#169

bvanderstee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #169

This PR will make it possible to run shell commands and scripts via the shell task

### DIFF
--- a/doc/tasks/shell.md
+++ b/doc/tasks/shell.md
@@ -1,6 +1,6 @@
 # Shell
 
-The Shell task will run your automated shell scripts.
+The Shell task will run your automated shell scripts / commands.
 It lives under the `shell` namespace and has following configurable parameters:
 
 ```yaml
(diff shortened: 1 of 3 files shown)


src/Task/Shell.php
```diff
@@ -75,12 +80,12 @@ public function run(ContextInterface $context)
     }
 
     /**
-     * @param $script
+     * @param array $scriptArguments
      */
-    private function runShell($script)
+    private function runShell(array $scriptArguments = array())
```
veewee: You can drop the default value, it will always be an array here.

doc/tasks/shell.md
```diff
@@ -18,7 +18,12 @@ parameters:
 
 This options specifies the paths to your shell scripts.
 You can specify one or more scripts. 
-All scripts need to succeed for the task to complete.
+
+You also can specify one or more shell commands:
+- In a string ``"./bin/command arg1 arg2 ... arg-n"``
+- In an array ``["./bin/command", "arg1", "arg2", ["arg-n"]]``
+
+All scripts / shell commands need to succeed for the task to complete.
```
veewee: Can you add a real configuration example for both underneath?

src/Task/Shell.php
bvanderstee: Thanks for the tip, default value has been removed.

doc/tasks/shell.md
bvanderstee: Added configuration examples for all possible configurations.

PR #319, opened 2017-02-27
Add various commit message tasks

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

Various tasks related to commit messages I created a while back that I thought might interest others. 

- `git_capitalized_subject` ensures commit message subject lines start with a capital letter.
- `git_empty_message` checks that the commit message is not empty.
- `git_single_line_subject` ensures commit message subject lines are followed by a blank line.
- `git_text_width` ensures the number of columns the subject and commit message lines occupy is under the preferred limits.
- `git_trailing_period` ensures commit message subject lines do not have a trailing period.

Both the tasks and the specs are more or less straight PHP ports of some [Overcommit](https://github.com/brigade/overcommit) hooks, not sure what the best way to give them attribution is?

The follow code to get the commit message as an array of lines is shared between most of the tasks, might be interesting to add as a method to `GitCommitMsgContext`?

```php
$commitMessage = str_replace("\r", '', $commitMessage);
$lines = explode("\n", $commitMessage);
```

# New Task Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,13 +105,19 @@ parameters:
         gherkin: ~
         git_blacklist: ~
         git_branch_name: ~
+        git_capitalized_subject: ~
         git_commit_message: ~
+        git_conflict: ~
+        git_empty_message: ~
+        git_single_line_subject: ~
+        git_text_width: ~
+        git_trailing_period: ~
         grunt: ~
         gulp: ~
         jsonlint: ~
         kahlan: ~
         npm_script: ~
-        phan: ~        
+        phan: ~
         phing: ~
         php7cc: ~
         phpcpd: ~
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -71,10 +71,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 - [Deptrac](tasks/deptrac.md)
 - [Gherkin](tasks/gherkin.md)
 - [Git blacklist](tasks/git_blacklist.md)
-<<<<<<< HEAD
 - [Git branch name](tasks/git_branch_name.md)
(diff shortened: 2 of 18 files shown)


veewee:
Thanks for the PR. I'll try to digg into the new tasks this week.

veewee:
Hi @jyggen, I did not have the time to look at the PR in depth yet.
There is one thing on my mind regarding all these tasks:
Maybe it's a better idea to add these checks as some kind of defaults / presets to the `Git\CommitMessage`? Most of the code is similar and it might make the configuration easier.

Something like this:

```yml
# grumphp.yml
parameters:
    tasks:
        git_commit_message:
            matchers:
                - /JIRA-([0-9]*)/
            options:            
                capitalized_subject: true
                allow_empty_message: false
                allow_single_line_subject: false
                allow_subject_trailing_period: false
                max_text_width_body: 72
                max_text_width_subject: 60

            # Maybe it's better to get rid of these and force people to add the modifiers to the matcher itself?
            case_insensitive: true
            multiline: true
            additional_modifiers: ''
```

What is your opinion on this statement?

jyggen:
@veewee Yeah, that could work. I'll give it a try later this week :)

sc0rp10:
Hi, guys!
In our company is there a standard (based on popular "git flow" model): commit message should start with the current branch name.
eg for branch `FEAT-1234` commit messages should look like `FEAT-1234 do some work`.
How do you think, is there a case for an additional option for `CommitMessage` task?

veewee:
Fixed in #369

PR #322, opened 2017-03-03
Removed file scanning code from GitCommitMsgContext

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no?
| Deprecations? | no
| Documented?   | no
| Fixed tickets | n/a

The commit msg hook does not need a list of files for any purpose. Why is it there? Seemingly because `ContextInterface` requires it. That is a very unfortunate interface. Perhaps we should also remove the method from the interface?

I had to remove this because it was impossible to run `git:commit-msg` from the command line because it was hanging [here](https://github.com/phpro/grumphp/blob/db227510aa702105f47a9b8ac5e7dfd9947b0306/src/Console/Command/Git/CommitMsgCommand.php#L77) whilst trying to scan for files or some such nonsense. Imagine my reaction when I find the reason the application is hung is because it's doing something it doesn't even need to do that doesn't serve any purpose. This is it. My reaction is to remove that useless code.

### DIFF
--- a/spec/Task/Context/GitCommitMsgContextSpec.php
+++ b/spec/Task/Context/GitCommitMsgContextSpec.php
@@ -2,21 +2,15 @@
 
 namespace spec\GrumPHP\Task\Context;
 
-use GrumPHP\Collection\FilesCollection;
 use GrumPHP\Task\Context\ContextInterface;
 use GrumPHP\Task\Context\GitCommitMsgContext;
 use PhpSpec\ObjectBehavior;
 
 class GitCommitMsgContextSpec extends ObjectBehavior
 {
-    /**
-     * @var string
-     */
-    protected $tempFile;
-
-    function let(FilesCollection $files)
+    function let()
     {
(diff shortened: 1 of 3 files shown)


veewee:
Can you tell me in which context the `git:commit-msg` was hanging? That never happened to me before... It's not because you don't use the files during the hook, that nobody has created a custom task that does use the files during that hook. I can imagine applying some rules like: if you changed composer dependencies, add x to your commit message.

Bilge:
>Can you tell me in which context the `git:commit-msg` was hanging?

Running it from the command like, e.g. `bin/grumphp git:commit-msg myfile`.

veewee:
I am not convinced that the files should be removed. We should pinpoint the exception instead, but at the moment I was not able to reproduce the error. Feel free to add some information about how I can reproduce it.

veewee:
Closing this one for now. Feel free to file another issue explaining the exception so that we can handle it.

Bilge:
There was no exception, it just hangs.

PR #323, opened 2017-03-03
Added support for a commit message rule names displayed in failure output

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | n/a

When a commit message failed we used to see:

>The commit message does not match the rule: /JIRA-\d+/

Now we see:

>Rule not matched: "Must contain JIRA issue number" /JIRA-\d+/

Seems good 👍

### DIFF
--- a/doc/tasks/git_commit_message.md
+++ b/doc/tasks/git_commit_message.md
@@ -1,7 +1,6 @@
 # Git commit message
 
-The git comit message can be used in combination with the git hook `git:commit-msg`.
-It can be used to enforce patterns in a commit message.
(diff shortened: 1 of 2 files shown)


veewee:
What about existing implementations? They will display:
```
Rule not matched: "0" /something/"
```

In that case it might be better not to display the ruleName?

Bilge:
I think it is fine as is. That message means rule index zero was matched, which is correct. If the user does not like this they should set a more appropriate name.

PR #324, opened 2017-03-03
Added original commit message to CommitMessage task

Bilge:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | n/a

GrumPHP is quite rude when he swallows your commit message. It's still available, if you know where to look, but it would be much more polite to echo it back to the user when this happens so they can just copy/paste it back into their editor for amending.

### DIFF
--- a/src/Task/Git/CommitMessage.php
+++ b/src/Task/Git/CommitMessage.php
@@ -96,7 +96,18 @@ class CommitMessage implements TaskInterface
         }
 
         if (count($exceptions)) {
-            return TaskResult::createFailed($this, $context, implode(PHP_EOL, $exceptions));
+            return TaskResult::createFailed(
+                $this,
+                $context,
+                implode(PHP_EOL, $exceptions) . <<<KAWAii
+</>
+
+Original commit message follows
+_______________________________
+
+$commitMessage
+KAWAii
+            );
         }
 
         return TaskResult::createPassed($this, $context);


src/Task/Git/CommitMessage.php
```diff
@@ -96,7 +96,18 @@ public function run(ContextInterface $context)
         }
 
         if (count($exceptions)) {
-            return TaskResult::createFailed($this, $context, implode(PHP_EOL, $exceptions));
+            return TaskResult::createFailed(
+                $this,
+                $context,
+                implode(PHP_EOL, $exceptions) . <<<KAWAii
+</>
```
veewee: Why is this one added? I think it's more clear to just keep the color like the other tasks do.

src/Task/Git/CommitMessage.php
```diff
@@ -96,7 +96,18 @@ public function run(ContextInterface $context)
         }
 
         if (count($exceptions)) {
-            return TaskResult::createFailed($this, $context, implode(PHP_EOL, $exceptions));
+            return TaskResult::createFailed(
+                $this,
+                $context,
+                implode(PHP_EOL, $exceptions) . <<<KAWAii
```
veewee: What does KaWaii even mean?! Maybe you could give the heredoc a more generic name?
Normally this kind of formatting is added to a formatter. That way you can change it to whatever formatting you like. However, the current formatters implementation is based on the processes. Maybe a `ContextFormatterInterface` should also be added for cases like this.

src/Task/Git/CommitMessage.php
Bilge: This makes the commit message visually distinguishable from the error output since it is not really part of the error. Furthermore, the overuse of red becomes hard to read at some point.

src/Task/Git/CommitMessage.php
Bilge: I'm not sure what you mean or how to do this.

src/Task/Git/CommitMessage.php
veewee: The overuse of random colors also becomes hard to read. That is why the messages have specific colors. Green = ok, Red = not ok, Yellow = warning.
I think it's best to keep it consequent for now. If you do want the colors, you could still it to a custom formatter for your project to make it 100% Kawaii.

src/Task/Git/CommitMessage.php
veewee: Other tasks have a `\GrumPHP\Formatter\ProcessFormatterInterface` as dependency in their constructor. Since this task does not run as a proccess, we might need to add a new interface or just a formatter without an interface that can be added to the constructor of the task. This formatter can then be used to style the exceptions. It is interchangeable in the configuration so that you can add your own formatter if you like. Feel free to ask any additiontal questions if the information above is not sufficient.

Issue #328, opened 2017-03-09
Validate Git branch naming convention


ademarco:
| Q               | A
| --------------- | ---
| Version         | `0.11.3`
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | no
| Related tickets | 

In my organisation we need to enforce specific Git branch naming conventions so I was wondering if this is something GrumPHP could take care of.

I've been sneaking around the code and it looks like we could leverage the `\Gitonomy\Git\Reference\Branch` class to gather information about the current branch and validate it using regular expressions, like the commit message task currently does.

Would this be something interesting for GrumPHP? In case it would be we could dedicate some time to it.

drupol:
👍

netlooker:
👍

veewee:
👍 That would be a nice addition. Feel free to implement.

ademarco:
Ok PR is ready for review!

veewee:
Implemented in #329.

PR #329, opened 2017-03-09
Validate Git branch naming convention #328

ademarco:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

This PR adds branch name checking as per #328 allowing to define branch names as follows:

```yaml
parameters:
  tasks:
      git_branch_name:
        matchers:
          - '/^issue-\d+-[a-z-]+/'
        case_insensitive: false
```

# New Task Checklist:

- [ ] Does the task contains phpspec tests?

### DIFF
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -19,6 +19,7 @@ parameters:
         gherkin: ~
         git_blacklist: ~
         git_commit_message: ~
+        git_branch_name: ~
         git_conflict: ~
         grunt: ~
         gulp: ~
@@ -61,6 +62,7 @@ Every task has it's own default configuration. It is possible to overwrite the p
 - [Gherkin](tasks/gherkin.md)
 - [Git blacklist](tasks/git_blacklist.md)
 - [Git commit message](tasks/git_commit_message.md)
+- [Git branch name](tasks/git_branch_name.md)
 - [Git conflict](tasks/git_conflict.md)
 - [Grunt](tasks/grunt.md)
 - [Gulp](tasks/gulp.md)
--- a/src/Configuration/GrumPHP.php
+++ b/src/Configuration/GrumPHP.php
@@ -41,6 +41,14 @@ class GrumPHP
     }
 
     /**
+     * @return \Gitonomy\Git\Repository
+     */
+    public function getGitRepository()
+    {
+        return $this->container->get('git.repository');
+    }
+
+    /**
      * @return string
      */
     public function getHooksDir()
--- a/src/Task/Git/AbstractRegex.php
+++ b/src/Task/Git/AbstractRegex.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace GrumPHP\Task\Git;
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Task\TaskInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+abstract class AbstractRegex implements TaskInterface {
+
+    /**
+     * @var GrumPHP
+     */
+    protected $grumPHP;
+
+    /**
+     * @param GrumPHP $grumPHP
+     */
+    public function __construct(GrumPHP $grumPHP)
+    {
+        $this->grumPHP = $grumPHP;
+    }
+
+    /**
+     * @return array
+     */
+    public function getConfiguration()
+    {
+        $configured = $this->grumPHP->getTaskConfiguration($this->getName());
+
+        return $this->getConfigurableOptions()->resolve($configured);
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+          'case_insensitive' => true,
+          'matchers' => [],
+          'additional_modifiers' => ''
+        ]);
+
+        $resolver->addAllowedTypes('case_insensitive', ['bool']);
+        $resolver->addAllowedTypes('matchers', ['array']);
+        $resolver->addAllowedTypes('additional_modifiers', ['string']);
+
+        return $resolver;
+    }
+}
--- a/src/Task/Git/BranchName.php
+++ b/src/Task/Git/BranchName.php
@@ -2,16 +2,17 @@
 
 namespace GrumPHP\Task\Git;
 
+use Gitonomy\Git\Reference\Branch;
 use GrumPHP\Runner\TaskResult;
 use GrumPHP\Task\Context\ContextInterface;
-use GrumPHP\Task\Context\GitCommitMsgContext;
+use GrumPHP\Task\Context\RunContext;
 use GrumPHP\Util\Regex;
 use GrumPHP\Exception\RuntimeException;
 
 /**
- * Git CommitMessage Task
+ * Git BranchName Task
  */
-class CommitMessage extends AbstractRegex
+class BranchName extends AbstractRegex
 {
 
     /**
@@ -19,7 +20,7 @@ class CommitMessage extends AbstractRegex
      */
     public function getName()
     {
-        return 'git_commit_message';
+        return 'git_branch_name';
     }
 
     /**
@@ -29,20 +30,7 @@ class CommitMessage extends AbstractRegex
      */
     public function canRunInContext(ContextInterface $context)
     {
-        return $context instanceof GitCommitMsgContext;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getConfigurableOptions()
-    {
-        $resolver = parent::getConfigurableOptions();
-        $resolver->setDefault('multiline', true);
-
-        $resolver->addAllowedTypes('multiline', ['bool']);
-
-        return $resolver;
+        return $context instanceof RunContext;
     }
 
     /**
@@ -61,10 +49,6 @@ class CommitMessage extends AbstractRegex
             $regex->addPatternModifier('i');
         }
 
-        if ((bool) $config['multiline']) {
-            $regex->addPatternModifier('m');
-        }
-
         $additionalModifiersArray = array_filter(str_split((string) $config['additional_modifiers']));
         array_map([$regex, 'addPatternModifier'], $additionalModifiersArray);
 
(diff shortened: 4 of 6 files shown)


ademarco:
Here a first draft, still need to provide tests.

I'm not sure how to answer to these:

```
 Is the run() method readable?
 Is the run() method using the configuration correctly?
```

ademarco:
@veewee does the PR look OK or should I polish it a little more?

src/Configuration/GrumPHP.php
```diff
@@ -41,6 +41,14 @@ public function getGitDir()
     }
 
     /**
+     * @return \Gitonomy\Git\Repository
+     */
+    public function getGitRepository()
```
veewee: It's better to inject the repository directly in the task constructor instead of using the configuration as additional service container. Can you remove this method?

src/Task/Git/AbstractRegex.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace GrumPHP\Task\Git;
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Task\TaskInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+abstract class AbstractRegex implements TaskInterface
```
veewee: Currently I don't think this abstract class is the way to go. There are some people working on the `git_commit_message` task in e.g. #319. This will surely break the implementation and doesn't add much value since it only contains the configurable options. I think it's better to drop this class and create a separate implementation.

src/Task/Git/AbstractRegex.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+          'case_insensitive' => true,
```
veewee: I don't know if the case_insensitive and addition_modifiers are the way to go. Maybe it's better to enforce strict modifiers in the pattern. This was also mentioned in #319.

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * @param array $config
+     * @param string $name
+     * @param string $rule
+     * @param string $ruleName
+     *
+     * @throws RuntimeException
+     */
+    protected function runMatcher(array $config, $name, $rule, $ruleName)
```
veewee: You can mark this one as private

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,97 @@
...
+
+        return TaskResult::createPassed($this, $context);
+    }
+
+    /**
+     * Get current branch name.
+     *
+     * @return string
+     */
+    public function getCurrentBranchName()
```
veewee: You can mark this one as private

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $gitRepository = $this->grumPHP->getGitRepository();
+        $branch = new Branch($gitRepository, $gitRepository->getHead()->getRevision());
```
veewee: I am not sure this will work in all cases. A commit can be in multiple branches. How does this code handle these?
Maybe it's better to execute the `git symbolic-ref --short HEAD` command through the `Repository::run()` command?

src/Task/Git/CommitMessage.php
```diff
@@ -2,32 +2,17 @@
...
-use GrumPHP\Task\TaskInterface;
 use GrumPHP\Util\Regex;
-use Symfony\Component\OptionsResolver\OptionsResolver;
+use GrumPHP\Exception\RuntimeException;
 
 /**
  * Git CommitMessage Task
  */
-class CommitMessage implements TaskInterface
+class CommitMessage extends AbstractRegex
```
veewee: Can you revert this class to the old version?

doc/tasks.md
```diff
@@ -19,6 +19,7 @@ parameters:
         gherkin: ~
         git_blacklist: ~
         git_commit_message: ~
+        git_branch_name: ~
```
ademarco: Done, also task link below in the same file.

README.md
```diff
@@ -102,6 +102,7 @@ parameters:
         gherkin: ~
         git_blacklist: ~
         git_commit_message: ~
+        git_branch_name: ~
```
ademarco: Done

src/Configuration/GrumPHP.php
ademarco: Yep, sorry! Removed, I'm using dependency injection now.

src/Task/Git/AbstractRegex.php
ademarco: Ok, removed in favour of already existing `additional_modifiers` option

src/Task/Git/BranchName.php
ademarco: Done.

src/Task/Git/BranchName.php
ademarco: Yep, removed that in favour of `::run()` method

src/Task/Git/CommitMessage.php
ademarco: Reverted

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    public function canRunInContext(ContextInterface $context)
+    {
+        return $context instanceof RunContext;
```
veewee: Shouldn't this also run in pre-commit context? Now it will not be triggered during a git commit.

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,130 @@
...
+     * @param string $rule
+     * @param string $ruleName
+     *
+     * @throws RuntimeException
+     */
+    private function runMatcher(array $config, $name, $rule, $ruleName)
+    {
+        $regex = new Regex($rule);
+
+        $additionalModifiersArray = array_filter(str_split((string) $config['additional_modifiers']));
```
veewee: The additional modifiers will always be a string since it is enforce by the OptionsResolver.

src/Task/Git/BranchName.php
ademarco: Ok done, I've added the pre commit context too.

veewee: Great, can you add a test for this as well?

ademarco: Ok, done!

src/Task/Git/BranchName.php
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * @param ContextInterface|RunContext $context
+     *
+     * @return TaskResult
+     */
+    public function run(ContextInterface $context)
+    {
+        $name = trim($this->repository->run('symbolic-ref', ['HEAD', '--short']));
```
veewee: When the current revision is a non symbolic reference, the command will throw an exception like: 
> Error code: 128
> Message: fatal: ref HEAD is not a symbolic ref

This can happen in detached head or if you checked out a specific commit.
This will throw a `Gitonomy\Git\Exception\ProcessException`.

How would you like to handle this issue on the task? Should it be configurable to allow non symbolic references or not? We might want to fetch the error message and display a user-friendly error message instead?

ademarco: Maybe we should not check for branch naming convention when you are on a detached head or you checkout a specific commit? We could just catch the exception and skip the step by returning a `TaskResult::createSkipped()`, does that make sense?

veewee: Maybe it's best to make this configurable through an extra configuration flag?
Some people might want to allow it, others may have something against it.

Rules:
- If you are allowed to commit on non symbolic references: task passes
- If you are not allowed to commit on non symbolic references: task fails with a decent error message.

ademarco: Ok shall we call the boolean option `allow_symbolic_references`?

ademarco: something like https://github.com/phpro/grumphp/pull/329/commits/4f95d276414a0a6aa50cec6fda5fe524d8be0820 ?

src/Task/Git/BranchName.php
```diff
@@ -63,12 +64,14 @@ public function getConfigurableOptions()
...
-          'matchers' => [],
-          'additional_modifiers' => ''
+            'matchers' => [],
+            'additional_modifiers' => '',
+            'allow_symbolic_references' => true,
         ]);
 
         $resolver->addAllowedTypes('matchers', ['array']);
         $resolver->addAllowedTypes('additional_modifiers', ['string']);
+        $resolver->addAllowedTypes('allow_symbolic_references', ['boolean']);
```
veewee: You are trying to search the branch name based on the symbolic reference 'HEAD'. Since in detached mode there is no symbolic reference HEAD, you retrieve the error. So I don't think the naming is correct here.
Maybe something like `allow_detached_head` is easier to understand?

src/Task/Git/BranchName.php
```diff
@@ -110,10 +113,18 @@ private function runMatcher(array $config, $name, $rule, $ruleName)
...
         $config = $this->getConfiguration();
         $exceptions = [];
 
+        try {
+            $name = trim($this->repository->run('symbolic-ref', ['HEAD', '--short']));
+        } catch (ProcessException $e) {
+            if ($config['allow_symbolic_references'] === false) {
+                $message = "Branch naming convention task is not allowed to run on symbolic references.";
+                return TaskResult::createFailed($this, $context, $message);
+            }
```
veewee: If the command fails and the detached head is allowed, you'll have to return a success result. Otherwise the matchers will fail.
If the detached head is not allowed, you can return the error message. Maybe it should be a bit more clear for less experienced users?

src/Task/Git/BranchName.php
veewee: Looks good. I've added some remarks.
Can you add documentation and tests for this new flag?

src/Task/Git/BranchName.php
ademarco: Ok I'll go for that one.

src/Task/Git/BranchName.php
ademarco: Done.

veewee:
Thanks for your work!

drupol:
:+1:

ademarco:
Thanks for merging @veewee !

jfhovinne:
:+1:

netlooker:
:+1: Massive feature :D

PR #333, opened 2017-03-18
Add timeout configuration to shell task

tgiordmaina:
Hi,

I add this feature to your task shell today because for a custom script of code verification which exceed 60 seconds I had this error :
```
[Symfony\Component\Process\Exception\ProcessTimedOutException]                             
The process "'/bin/sh' 'your_custom_shell_script.sh'" exceeded the timeout of 60 seconds.
```

Can you accept this contribution for your next release ?

I used it as a patch today and it works very well,

Thanks by advance,

 Regards,

Thomas Giordmaina

Additional Source : http://symfony.com/doc/current/components/process.html#process-timeout

### DIFF
--- a/src/Task/Shell.php
+++ b/src/Task/Shell.php
@@ -30,11 +30,13 @@ class Shell extends AbstractExternalTask
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'scripts' => [],
+            'timeout' => 60,
             'triggered_by' => ['php']
         ]);
 
         $resolver->addAllowedTypes('scripts', ['array']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
+        $resolver->addAllowedTypes('timeout', ['int']);
         $resolver->setNormalizer('scripts', function ($resolver, $scripts) {
             return array_map(function ($script) {
                 return is_string($script) ? (array) $script : $script;
@@ -84,10 +86,12 @@ class Shell extends AbstractExternalTask
      */
     private function runShell(array $scriptArguments)
     {
+        $config = $this->getConfiguration();
         $arguments = $this->processBuilder->createArgumentsForCommand('sh');
         $arguments->addArgumentArray('%s', $scriptArguments);
 
         $process = $this->processBuilder->buildProcess($arguments);
+        $process->setTimeout($config['timeout']);
         $process->run();
 
         if (!$process->isSuccessful()) {
--- a/doc/tasks/shell.md
+++ b/doc/tasks/shell.md
@@ -9,6 +9,7 @@ parameters:
     tasks:
         shell:
             scripts: []
+            timeout: 60
             triggered_by: [php]
 ```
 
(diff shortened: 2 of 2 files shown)


jyggen:
Note that you can already configure the global timeout with [`process_timeout`](https://github.com/phpro/grumphp/blob/master/doc/parameters.md).

tgiordmaina:
Yes I tried with this paramater **process_timeout** but it had no effect on the Symfony process runner which trigger an Exception of type ProcessTimedOutException (after 60 second by default)

src/Task/Shell.php
```diff
@@ -84,10 +86,12 @@ public function run(ContextInterface $context)
      */
     private function runShell(array $scriptArguments)
     {
+        $config = $this->getConfiguration();
         $arguments = $this->processBuilder->createArgumentsForCommand('sh');
         $arguments->addArgumentArray('%s', $scriptArguments);
 
         $process = $this->processBuilder->buildProcess($arguments);
```
veewee: This `buildProcess()` method here makes sure the timeout is being set to the process builder:

```php
    public function buildProcess(ProcessArgumentsCollection $arguments)
    {
        $builder = SymfonyProcessBuilder::create($arguments->getValues());
        $builder->setTimeout($this->config->getProcessTimeout());
        $process = $builder->getProcess();

        $this->logProcessInVerboseMode($process);
        $this->guardWindowsCmdMaxInputStringLimitation($process);

        return $process;
    }
```

veewee:
@tgiordmaina,

Thanks for the PR. 
I do think you misconfugred something, because the global `process_timeout` just works on the shell task.
For example:

Use following setup:
```sh
mkdir tmp && cd tmp
composer require phpro/grumphp
echo "sleep $1" > sleep.sh 
```

Use following config:
```yaml
# grumphp.yml
parameters:
    git_dir: .
    bin_dir: vendor/bin
    process_timeout: 5
    tasks:
        shell:
            scripts:
              - [sleep.sh, 10]
            triggered_by: [sh]
```

Result:
```sh
git add -A
./vendor/bin/grumphp run

Running task 1/1: Shell...

  [Symfony\Component\Process\Exception\ProcessTimedOutException]
  The process "'/bin/sh' 'sleep.sh' '10'" exceeded the timeout of 5 seconds.
```

As you can see, the process timeout has been set to 5 seconds.

tgiordmaina:
I'm really sorry for the mistake, I tried yesterday (may be with my patch) and this parameter had no effect.

Indeed without my patch the build process set the timeout correctly.

Thanks for your time and your answer @veewee 
Thanks to @jyggen too, finaly you had right..

I have an additionnal question before purpose another patch :yum: :
A parameter already exists to display the output of shell (while the process is executed ?) for example :
```bash
Running task 1/3: PhpLint
Running task 2/3: SecurityChecker
Running task 3/3: Shell
+ My awesome output 1
+ My awesome output 2...
```

PS: this PR may will be useful for users which search a solution for the Exception 'ProcessTimedOutException' because I found nothing on google by typing 'grumphp error [Symfony\Component\Process\Exception\ProcessTimedOutException]'.

Your project is amazing and full of stuff, thanks you again for you quickly answer !

Regards

veewee:
I've added information about the exception to the parameters file: https://github.com/phpro/grumphp/blob/master/doc/parameters.md

It is not possible to display success messages from a task. This is by design to keep the output clean.

Issue #278, opened 2017-01-30
Add deptrac task


veewee:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | yes
| Related tickets | 

https://github.com/sensiolabs-de/deptrac

> Deptrac is a static code analysis tool that helps to enforce rules for dependencies between software layers.
> For example, you can define a rule like "controllers may not depend on models". To ensure this, deptrac analyzes your code to find any usages of models in your controllers and will show you where this rule was violated.

veewee:
Fixed in 337.

PR #337, opened 2017-03-31
Feature/deptrac 

lykciv:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #278

# New Task Checklist:

The deptrac tool should be installed manually as recommended by their documentation.

### DIFF
--- a/doc/tasks/deptrac.md
+++ b/doc/tasks/deptrac.md
@@ -1,0 +1,51 @@
+# Deptrac
+
+Follow the [installation instructions](https://github.com/sensiolabs-de/deptrac#installation) to add deptrac to your 
+project.
+
+The Deptrac task will check for dependencies between the software layers of your project. It lives under the `deptrac` 
+namespace and has following configurable parameters:
+
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        deptrac:
+            formatter_graphviz: ~
+            formatter_graphviz_display: ~
+            formatter_graphviz_dump_image: ~
+            formatter_graphviz_dump_dot: ~
+            formatter_graphviz_dump_html: ~
+```
+
+
+**formatter_graphviz**
+
+*Default: 0*
+
+Set to `1` to enable the graphviz formatter.
+
+**formatter_graphviz_display**
+
+*Default: false*
+
+Open the generated graphviz image. Set to `true` to activate.
+
+**formatter_graphviz_dump_image**
+
+*Default: null*
+
+Set path to a dumped png file.
+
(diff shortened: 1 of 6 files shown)


src/Task/Deptrac.php
```diff
@@ -0,0 +1,91 @@
...
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'formatter_graphviz' => 0,
+            'formatter_graphviz_display' => false,
+            'formatter_graphviz_dump_image' => null,
+            'formatter_graphviz_dump_dot' => null,
+            'formatter_graphviz_dump_html' => null,
+        ]);
+
+        $resolver->addAllowedTypes('formatter_graphviz', ['int']);
```
veewee: Shouldn't this be a boolean as well?

src/Task/Deptrac.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'formatter_graphviz' => 0,
```
veewee: How can I specify a depfile?

doc/tasks/deptrac.md
```diff
@@ -0,0 +1,51 @@
+# Deptrac
+
+Follow the [installation instructions](https://github.com/sensiolabs-de/deptrac#installation) to add deptrac to your
```
veewee: Can't deptrac also be installed through composer?
Since it has a composer.json file in it's repository, I guess it should be possible.

src/Task/Deptrac.php
lykciv: I have added the depfile option

src/Task/Deptrac.php
lykciv: Updated this option so it allows boolean configuration. In deptrac itself this option is an integer though.

doc/tasks/deptrac.md
lykciv: The project itself [recommends installing the phar file](https://github.com/sensiolabs-de/deptrac#download-the-phar-recommended). 

Also, the package does not come with a bin file. I guess this should also work when you run `php deptrac.php` in the package, but I didn't look further into that option.

veewee: Ok. Let keep it this way for now.

Issue #309, opened 2017-02-22
Run same task twice with different configuration


Bilge:
| Q               | A
| --------------- | ---
| Version         | 0.11.1
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | n/a

One may wish to run some tasks twice with different configurations. For example, run phpcs with warnings treated as errors for some directories and with warnings disabled for different directories. However, this GrumPHP configuration uses the task name as the configuration key, so it is not possible to specify the same task twice due to the format of the configuration.

veewee:
I am not in favour of this option since it is making the configuration part very difficult. Currently it is very easy to get the project up and running with some basic configuration. If you want this kind of advanced solutions, you might just want to create a spefic task for your use case?

Feel free to add some thoughts on how you see the configuration part. Maybe it is not clear to me at the moment.

Bilge:
Presently I haven't put any thought into how the configuration should look, I just wanted to open the discussion on whether this is something we want to support.

patkar:
We ran in this issue too. We have two phpcsfixer config files. One `.php_cs` file for all files except some special TYPO3 files. And a `.typo3.php_cs` with the same settings as the previous one except `declare_strict_types`. So we need to run phpcsfixer two times with the different config files. Sadly the special TYPO3 files are requiring this.

I could imagine the following enhancement as example to the config:

```yaml
parameters:
    git_dir: .
    bin_dir: vendor/bin
    tasks:
        phpcsfixer2:
            allow_risky: true
            path_mode: intersection
            triggered_by: ['php']
        phpcsfixer2_typo3:
            task: phpcsfixer2
            allow_risky: true
            config: .typo3.php_cs
            path_mode: intersection
            triggered_by: ['php']
```

As you can see, if there is a task parameter in the config, the key of the config will be ignored and the value of the task parameter will be used instead. No BC break and the easy of configuration stays. But special snowflakes ( :trollface: ) like me and others in need can use the extended syntax :grimacing:

szinya:
I could use this too. We have separate phpmd configs for the project files and the tests. With the changes proposed by @patkar our grumphp.yml would look something like this:

```yaml
parameters:
    tasks:
        phpmd:
            exclude: [tests]
            ruleset: [phpmd.xml.dist]
        phpmd_tests:
            task: phpmd
            exclude: [src]
            ruleset: [phpmd_tests.xml.dist]
```

[...]

PR #350, opened 2017-04-16
Changes filenames: phpcsfixer, securitychecker

cafferata:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

Expected that the filenames would be the same as there namespaces


veewee:
Thanks!

patkar:
Aaaaaaaaaaaand it's a BC break if you use redefine php-cs-fixer tasks to have multiple ones :sob: (see https://github.com/phpro/grumphp/issues/309#issuecomment-295249909)

cafferata:
Thanks @patkar for pointing this out.

veewee:
Hi @patkar, 

Thanks for noticing. I've created a BC break label and will try to review these kind of BC breaks as well. 
Let's hope that this is sufficient to avoid BC breaks as much as possible. Do note that we are in the 0.x phase, so there might be some additional breaks in the future.

patkar:
Of course it's 0.x so I'm aware of bigger changes. And also this is a subtile one, because is this an extension point or not.. 

And it was easy enough and fastly fixed in our conventions checker. Gladfully we did that in a central repo and not every project created this config on their own. I only wanted to point that out for further considerations/issues/PRs ;)

Issue #339, opened 2017-04-01
Cannot commit because of two execs in pre-commit


hoffmannjonas:
Version: 0.11.3
Bug: yes

I cannot run a commit becasue of two execs in pre-commit hook:
```
#!/bin/sh

#
# Run the hook command.
# Note: this will be replaced by the real command during copy.
#

# Fetch the GIT diff and format it as command input:
DIFF=$(git diff -r -p -m -M --full-index --staged | cat)

# Run GrumPHP
(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | exec exec $(HOOK_COMMAND) '--skip-success-output')
```

Can easily be fixed by removing an exec

veewee:
Hi @hoffmannjonas,

Thanks for reporting this issue.

Can you tell me which preset you are using?
As far as I can see, there is only one exec in the template:
https://github.com/phpro/grumphp/blob/master/resources/hooks/local/pre-commit

phux:
Please check [PR 352](https://github.com/phpro/grumphp/pull/352).

veewee:
This is a Symfony Process related issue:

It looks like this additional exec is added by the Symfony Process component here in this PR: symfony/symfony#21474
This issue mentions the break: symfony/symfony#22476
And a PR is on it's way: symfony/symfony#22476

Since the fix is not merged yet, you can solve this issue in your own project by copying the adjusted git hook templates from #352 and use the hooks_dir parameter (https://github.com/phpro/grumphp/blob/master/doc/parameters.md) .

PR #352, opened 2017-04-21
Removed duplicate 'exec' in local githooks

phux:
| Q             | A
| ------------- | ---
| Branch        | 0.11.4
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #339

The hook templates contain a duplicated `exec` because there is already `exec` defined in the `HOOK_COMMAND`.

### DIFF
--- a/resources/hooks/local/commit-msg
+++ b/resources/hooks/local/commit-msg
@@ -13,4 +13,4 @@ COMMIT_MSG_FILE=$1
 DIFF=$(git diff -r -p -m -M --full-index --staged | cat)
 
 # Run GrumPHP
-(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | exec $(HOOK_COMMAND) "--git-user=$GIT_USER" "--git-email=$GIT_EMAIL" "$COMMIT_MSG_FILE")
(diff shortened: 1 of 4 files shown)


voidtek:
I applied this patch on two machines, the first one have this issue and the second one works perfectly without this patch.
After applying this patch, these two machines work now perfectly, the issue mentioned doesn't occur and we have not regression bugs. 

Thanks for your job.

veewee:
I would rather deal with the source of the issue instead of removing the exec from the hook template since it is there for a reason.
Can you give me some additional information about systems that are adding those second "exec" statement? (OS, PHP version, composer show, ...)

phux:
You're right, it's due to my symfony/process version which is the mentioned 3.3 :)
And you're right, it's not the responsibility of this repo to fix it.

I am fine with closing this PR. I will just use my fork until `process` is fixed.

veewee:
Alright, I'll close this one.
You could also copy the adjusted hook templates to your project and use the `hooks_dir` parameter (https://github.com/phpro/grumphp/blob/master/doc/parameters.md) . This way you don't have to maintain your own patch.

phux:
@veewee even better. Thanks for mentioning it!

PR #357, opened 2017-05-17
Add PhpMnd task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | n/a

This PR adds support for [PHP Magic Number Detector](https://github.com/povils/phpmnd) (PHPMND).

> `phpmnd` is a tool that **helps** you detect magic numbers in PHP code.
>
> **What is Magic Number?**
>Magic number is a numeric literal that is not defined as a constant that may change at a later stage, but that can be therefore hard to update. It is a bad programming practice of using numbers directly in source code without explanation. In most cases this makes programs harder to read, understand, and maintain.

# New Task Checklist:

### DIFF
--- a/src/Task/PhpMnd.php
+++ b/src/Task/PhpMnd.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace GrumPHP\Task;
+
+use GrumPHP\Runner\TaskResult;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * PhpMnd task
+ */
+class PhpMnd extends AbstractExternalTask
+{
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return 'phpmnd';
+    }
+
+    /**
+     * @return OptionsResolver
+     */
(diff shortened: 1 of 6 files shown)


src/Task/PhpMnd.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'directory' => '.',
```
veewee: Won't this also run on ignored folders like vendor by default? Maybe that is something that can be added to exclude_path by default if there is no way to pass files to the command?

veewee:
Thanks @jyggen! The PR looks good, I've added one little question.

src/Task/PhpMnd.php
jyggen: Nice catch! I looked into it and it turns out that PHPMND automatically adds `vendor` to the exclude list ([as seen here](https://github.com/povils/phpmnd/blob/a85c7c746c52f15b8326596523889b16b39a7665/src/PHPFinder.php#L32)), so it should be fine as-is.

veewee:
Great. Thanks for the addition!

PR #358, opened 2017-05-18
Maintain PHP 7.0 support

mharbers:
Update ocramius/proxy-manager dependency to minor version v2.0.* to maintain PHP 7.0 support. As of version ocramius/proxy-manager v2.1.* PHP 7.1 is required.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "doctrine/collections": "~1.2",
         "gitonomy/gitlib": "~1.0",
         "monolog/monolog": "~1.16",
-        "ocramius/proxy-manager": "~0.4|~1.0|~2.0",
+        "ocramius/proxy-manager": "~0.4|~1.0|~2.0.0",
         "seld/jsonlint": "~1.1",
         "symfony/config": "~2.7|~3.0",
         "symfony/console": "~2.7|~3.0",


jyggen:
I'm not what this is trying to fix because PHP 7.0 is still supported even if `2.1.x` requires PHP 7.1+. If you include GrumPHP in a PHP 7.0 project Composer is smart enough to install `2.0.x` while in a PHP 7.1 project it'll install `2.1.x`. This PR will prevent PHP 7.1+ projects from installing `2.1.x`.

mharbers:
Yeah, you’re right. Appears I run grumphp in docker-container with PHP 7.0, which was installed with composer on PHP 7.1. Sorry for the PR.



Op 18 mei 2017, om 18:33 heeft Jonas Stendahl <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


I'm not what this is trying to fix because PHP 7.0 is still supported even if 2.1.x requires PHP 7.1+. If you include GrumPHP in a PHP 7.0 project Composer is smart enough to install 2.0.x while in a PHP 7.1 project it'll install 2.1.x. This PR will prevent PHP 7.1+ projects from installing 2.1.x.

PR #364, opened 2017-05-31
Fix symfony progressbar issues

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

Symfony 3.3.0 makes the ProgressBar final. This causes all mocked ProgressBar objects to crash in phpspec.

Additional info:  https://github.com/symfony/symfony/pull/20487

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - hhvm
   - nightly
 
 env:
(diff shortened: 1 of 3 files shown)


veewee:
I also dropped HHVM support in this ticket since I retrieve following Travis error:

```
HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.
```

Support for HHVM is also dropped by other package like Symfony and Doctrine because of compatibility issues with PHP 7. It looks like a good idea to drop support for HHVM in this package as well. More information: https://symfony.com/blog/symfony-4-end-of-hhvm-support

PR #369, opened 2017-06-14
Add various enforcements to git_commit_message

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

These are the same enforcements introduced by the tasks in #319, but built into the existing `git_commit_message` task instead (as discussed [here](https://github.com/phpro/grumphp/pull/319#issuecomment-284313781)).

While I agree that it's probably a good idea to drop `case_insensitive`, `multiline` and `additional_modifiers` and let the matchers define their desired modifiers themselves, that is a breaking change and not really in the scope of this PR.

### DIFF
--- a/doc/tasks/git_commit_message.md
+++ b/doc/tasks/git_commit_message.md
@@ -1,6 +1,6 @@
 # Git commit message
 
-The Git commit message task ensures commit messages match the specified patterns.  
+The Git commit message task ensures commit messages match the specified patterns.
 For example: if you are working with JIRA, it is possible to add a pattern for the JIRA issue number.
 
 ```yaml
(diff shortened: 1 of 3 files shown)


veewee:
Hi @jyggen,

Thanks for you work on this PR. I am currently a bit busy, but try to review this one by the end of the month. Sorry for the late review :)

veewee:
Thanks for your work @jyggen, the PR looks really good! Also a big +1 for the edge case tests!

Issue #327, opened 2017-03-08
PHP CS Fixer v2 integration is slow


keradus:
v2 allows to run single process with multiple files.
current implementation runs multiple processes per single file: https://github.com/phpro/grumphp/blob/53816aef86fb497e79fad024d5299f002c9ed1b7/src/Task/AbstractPhpCsFixerTask.php#L120
as you dont run the Runner class but call CLI command, this will run 10 commands instead of one, making it 10 times slower.

veewee:
It was added in V1 to run the task in parallel and just copied for V2. I did not know it had support for multiple files in V2. Feel free to improve the task at any time.

SerkanYildiz:
What kind of solution do we want here? We can pass the value of the project dir to the fix parameter that will scan all files in 1 time. Or we can check the changed files/files to commit and find common directories and scan dirs instead of files. I would be happy to fix this.

keradus:
I would suggest to either:
- pass multiple files to `fix` command (regardless current or suggested solution, modified file could be excluded in PHP CS Fixer config (eg .php_cs.dist), so it would be nice to respect that running command with path and intersection mode)
- don't execute CLI at all, just run `Runner` of PHP CS Fixer programmatically.

veewee:
Fixed in #395

PR #371, opened 2017-06-16
Set default path mode to intersection

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | /

### DIFF
--- a/src/Task/PhpCsFixerV2.php
+++ b/src/Task/PhpCsFixerV2.php
@@ -32,7 +32,6 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
             'config' => null,
             'rules' => [],
             'using_cache' => true,
-            'path_mode' => null,
             'verbose' => true,
             'diff' => false,
             'triggered_by' => ['php'],
@@ -43,12 +42,9 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
         $resolver->addAllowedTypes('config', ['null', 'string']);
         $resolver->addAllowedTypes('rules', ['array']);
         $resolver->addAllowedTypes('using_cache', ['bool']);
-        $resolver->addAllowedTypes('path_mode', ['null', 'string']);
         $resolver->addAllowedTypes('verbose', ['bool']);
         $resolver->addAllowedTypes('diff', ['bool']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
-
-        $resolver->setAllowedValues('path_mode', [null, 'override', 'intersection']);
 
         return $resolver;
     }
@@ -82,7 +78,6 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
         }
 
         $arguments->addOptionalArgument('--using-cache=%s', $config['using_cache'] ? 'yes' : 'no');
-        $arguments->addOptionalArgument('--path-mode=%s', $config['path_mode']);
         $arguments->addOptionalArgument('--verbose', $config['verbose']);
         $arguments->addOptionalArgument('--diff', $config['diff']);
         $arguments->add('fix');
@@ -90,6 +85,8 @@ class PhpCsFixerV2 extends AbstractPhpCsFixerTask
         if ($context instanceof RunContext && $config['config'] !== null) {
             return $this->runOnAllFiles($context, $arguments);
         }
+
+        $arguments->add('--path-mode=intersection');
 
         return $this->runOnChangedFiles($context, $arguments, $files);
     }
--- a/doc/tasks/phpcsfixer2.md
+++ b/doc/tasks/phpcsfixer2.md
@@ -22,7 +22,6 @@ parameters:
             config: ~
             rules: []
             using_cache: true
-            path_mode: ~
             verbose: true
             diff: false
             triggered_by: ['php']
@@ -95,14 +94,6 @@ The caching mechanism is enabled by default.
 This will speed up further runs by fixing only files that were modified since the last run.
 The tool will fix all files if the tool version has changed or the list of fixers has changed.
 Cache is supported only for tool downloaded as phar file or installed via composer.
-
-
-**path_mode**
-
-*Default: null*
-
-Specify path mode (can be override or intersection).
-
 
 **verbose**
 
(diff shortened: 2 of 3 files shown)


veewee:
Hi @Landerstraeten,

Thanks for your PR! 

I've tested the PR and found out that during `grumphp run` no files were validated in php-cs-fixer2. I think that this is due that no files were added to the input, so the intersection is empty.
On the other hand, it does solve the issue of ignored files being validated during `grumphp git:pre-commit`.

I think it's best to fix this issue by removing the ability to configure the path modus and hardcode it:

- During grumphp run: no path mode specified
- During grumphp git:pre-commit: intersection path mode 

In the last case, it's also an option to implement this issue:
https://github.com/phpro/grumphp/issues/327


@keradus : Can you confirm that running intersection path mode with empty path arguments is causing an empty set to validate?

keradus:
I can confirm. intersection is basically operation of sets. when one take common part of anything and empty set, it's always empty set

keradus:
@veewee , hardcoded path mode would actually follow reasoning why we introduced intersection.
for regular run - always normal mode, for precommit/CI - always intersection

keradus:
> BC breaks?	no

disagree, public argument was removed

veewee:
It is indeed a BC break, I've changed the description and the milestone.
The code looks good at first sight. I'll do a more in-depth review this friday.

src/Task/PhpCsFixerV2.php
```diff
@@ -91,6 +86,8 @@ public function run(ContextInterface $context)
             return $this->runOnAllFiles($context, $arguments);
         }
 
+        $arguments->add('--path-mode=intersection');
```
keradus: it will crash with `Cannot create intersection with not-fully defined Finder in configuration file.` exception if there is no finder configured in `.php_cs(.dist)?` file.

veewee: How would you suggest to solve it?

keradus: you cannot know it without executing config file.

veewee: We could execute the config file, but this won't be able if php-cs-fixer is ran in PHAR modus. In that case, the `PhpCsFixer\Config` won't be available. 
I do understand why it crashes, but isn't it more logical if it runs in intersect modus without a finder, it runs on the specified files only? However, this would require a change in php-cs-fixer2.

keradus: sorry, not sure what are `modus`

> We could execute the config file, but this won't be able if php-cs-fixer is ran in PHAR modus. 

not true.

> isn't it more logical if it runs in intersect modus without a finder

no, it intersect you need to have at least two collections. here, it's cli args + finder from config.

veewee: @keradus,
If people [installed php-cs-fixer globally](https://github.com/FriendsOfPHP/PHP-CS-Fixer#globally-manual), then that executable could be used in grumphp. This means that we cannot autoload the php-cs-fixer configuration class / interface at that point and we won't be able to check if the config contains a finder.

Maybe another path-mode could be added that only uses the files added to the cli command?

keradus: then it's a matter how you execute php-cs-fixer from grumphp.

that's `--path-mode=override`, take everything passed to cli command and ignore finder from config file. yet, it will ignore whole finder, including exclusions. you cannot partially honour finder and partially not.

keradus: poor, quick and dirty solution would be to execute command with `intersection` mode, if it crashes on not fully defined finder, run it with `override` mode...
or, as suggested somewhere, execute `Runner` programatically instead of calling CLI command.

veewee: Hi @keradus,

The problem with override is that ignored files are being checked by php-cs-fixer which results in errors.
We cannot use the `Runner` for the same reason as mentioned above: some people are using the global installled php-cs-fixer, so we can't autoload the php-cs-fixer classes.

This is the problem we are trying to fix in this ticket, but there doesn't seem to be one fitting solution. As you mentioned, currently the best option is to handle errors with intersection and an empty finder.

Feel free to suggest any other solution if you think of something we didn't look at yet.

keradus: > As you mentioned, currently the best option is to handle errors with intersection and an empty finder.

and that means you need to load config from file and change it, it finder is not fully configured.
but that, again, would require grumphp to understand the code it's executing (know the classes etc), which, as other solution I suggested, you refused.
you want to be aware of configuration but refuse to know it classes. not much you can do with that restriction (except run cli up to 2 times approach)

veewee: Hi @keradus ,

It's not that I refuse to know it classes, it's about some configurations using the globally installed php-cs-fixer. For locally installed fixers it wouldn't be a problem. 
Instead of running the cli tool twice, it is also possible to make the path mode configurable again, but this time with a boolean that looks like `intersect_on_commit` with a default value of true. That would probably also fix this problem.

keradus: you could know the classes and still execute phar from cli.

that new mode is a hacky way of doing this which abuse package that is external for you, putting special mode only for one user is no-go.
and that option is not `intersect_on_commit`, it's more like `intersect_wich_finder_from_config_but_ignore_if_there_is_no_finder_in_config`... hacky

veewee: @keradus, 

The only way we could know these classes is by adding php-cs-fixer to the dependencies, even for projects who don't want to use that tool. So this is something I would like to avoid. Or are you talking about some other way?

To be sure we are talking about the same thing: I was talking about a configuration inside grumphp named `intersect_on_commit` which will use the `intersect` path mode during a commit inside GrumPHP. There is no need for any changes in php-cs-fixer for that.

keradus: > I was talking about a configuration inside grumphp named intersect_on_commit which will use the intersect path mode during a commit inside GrumPHP. There is no need for any changes in php-cs-fixer for that.

Cool it's gonna be internal setting.
yet, when `intersect_on_commit` would be set up, then you will end up with current issue as well, key is configured, intersection is tried to be used, yet no local config finder is provided.

> Or are you talking about some other way?

if you don't want to have that as dependency and treat it as a blackbox, then:
> poor, quick and dirty solution would be to execute command with intersection mode, if it crashes on not fully defined finder, run it with override mode...

veewee:
Fixed in #395

Issue #293, opened 2017-02-13
Wrong BlacklistTask output


sc0rp10:
| Q               | A
| --------------- | ---
| Branch          | master
| Bug?            | yes

Hi!
I have legacy project with some codegenerating classes with ugly code inside.
And I have blacklisted keyword 'private $' for prevent creating private field (in favor of protected).
When I make some changes in classes with codegenerating, i'm got ugly formatted error about my blacklisted word:
```
t.php
  1:14:  null;
```
which is less informative: it contains wrong line/symbol number and cutted and unreadable description.

**My configuration**
```yaml
# grumphp.yml
parameters:
    tasks:
      git_blacklist:
        keywords:
          - "private \\$"
        triggered_by: [php]
```

`cat t.php`
```php
<?php

class F
{
    protected function foo()
    {
        $code = <<<PHP
    private $' . $this->callSomeVeryVeryVeryLongMethodName() . ' = null;
PHP;
    }
}

```
# Result
```
# Run GrumpHP:
$ git ci -m "init"
GrumPHP detected a pre-commit command.
GrumPHP is sniffing your code!
Running task 1/1: Blacklist
# skip ascii
You have blacklisted keywords in your commit:
t.php
  1:14:  null; # <- wut?
To skip commit checks, add -n or --no-verify flag to commit command
```

veewee:
I would suggest using a custom visitor in combination with the phpparser task for this kind of checks.
The blacklist is made for easy matches. If you want to check more advanced patterns, the phpparser is the way to go.

Nevertheless, we should fix the output of the blacklist task.

veewee:
Fixed in #375

PR #375, opened 2017-06-30
Blacklisted keywords on lines bigger then 80 chars will now display their error message correctly.

bvanderstee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #293 

For blacklisted keywords on lines longer then 80 characters the output
was not well formatted. The wrong starting position was used
within the mb_substr method, a negative one which resulted that the substr started
at the end of the error message. This has been
resolved by making sure the starting position will not be negative.

The outcome is now the following:
```
You have blacklisted keywords in your commit:
t.php
1:20: 8:    private $' . $this->callSo
```
The last number is the actual line number where the keyword is in.

### DIFF
--- a/src/Formatter/GitBlacklistFormatter.php
+++ b/src/Formatter/GitBlacklistFormatter.php
@@ -93,6 +93,10 @@ class GitBlacklistFormatter implements ProcessFormatterInterface
             } while ($pos2 < $pos);
 
             $pos -= static::SPACE_BEFORE;
+            //$pos can't be lower then 0 else we start at the end of the string instead of the beginning
+            if ($pos < 0) {
+                $pos = 0;
+            }
             $pos2 += static::SPACE_AFTER;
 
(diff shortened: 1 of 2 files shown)


src/Formatter/GitBlacklistFormatter.php
```diff
@@ -93,6 +93,10 @@ private function trimOutputLine($line, $lineNumber)
             } while ($pos2 < $pos);
 
             $pos -= static::SPACE_BEFORE;
+            //$pos can't be lower then 0 else we start at the end of the string instead of the beginning
+            if ($pos < 0) {
```
veewee: Can you add a testcase for this fix?

bvanderstee: Yes I can.

Issue #374, opened 2017-06-30
Add file size check on commit


lykciv:
| Q               | A
| --------------- | ---
| Version         | -
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | -

Git does not cope well with big file sizes. I think it would be good if grumphp would stop you from trying to commit files over a certain size. 

I did a quick google and I found this, so I think it should be possible to implement this feature. I haven't tried it out yet myself:
https://stackoverflow.com/questions/39576257/how-to-limit-file-size-on-commit

veewee:
Since we are using `SplFileObject`s internally, you can easily create a new task and check the `SplFileObject::getSize()` method for every file against a configurable maximum allowed filesize.

Landerstraeten:
@veewee, you can assign this issue to me

PR #379, opened 2017-07-13
Add file size task

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #374 

# New Task Checklist:

### DIFF
--- a/spec/Task/FileSizeSpec.php
+++ b/spec/Task/FileSizeSpec.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace spec\GrumPHP\Task;
+
+use ArrayIterator;
+use GrumPHP\Collection\FilesCollection;
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Process\ProcessBuilder;
+use GrumPHP\Runner\TaskResult;
+use GrumPHP\Runner\TaskResultInterface;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use GrumPHP\Task\FileSize;
+use PhpSpec\ObjectBehavior;
+use Symfony\Component\Finder\SplFileInfo;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class FileSizeSpec extends ObjectBehavior
+{
+    function let(GrumPHP $grumPHP)
+    {
+        $grumPHP->getTaskConfiguration('file_size')->willReturn([]);
+        $this->beConstructedWith($grumPHP);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(FileSize::class);
+    }
+
+    function it_should_have_a_name()
+    {
+        $this->getName()->shouldBe('file_size');
+    }
+
+    function it_should_have_configurable_options()
+    {
+        $options = $this->getConfigurableOptions();
+        $options->shouldBeAnInstanceOf(OptionsResolver::class);
+        $options->getDefinedOptions()->shouldContain('max_size');
+    }
+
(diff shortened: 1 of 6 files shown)


spec/Task/FileSizeSpec.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $context->getFiles()->willReturn(new FilesCollection([
+            new SplFileInfo('src/Collection/TaskResultCollection.php', 'src/Collection', 'TaskResultCollection.php'),
+        ]));
+
+        $result = $this->run($context);
+        $result->shouldBeAnInstanceOf(TaskResultInterface::class);
+        $result->isPassed()->shouldBe(true);
+    }
+}
```
veewee: Can you add a tests for the failing task?

src/Task/FileSize.php
```diff
@@ -0,0 +1,97 @@
...
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'max_size' => '10M',
+        ]);
+
+        $resolver->addAllowedTypes('max_size', ['string']);
```
veewee: It is also possible to just add a number in bytes without a suffix. (see \Symfony\Component\Finder\Comparator\NumberComparator)
So maybe it should also be possible to allow ints?

src/Task/FileSize.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'max_size' => '10M',
```
veewee: Is there a specific reason you've choosen 10M here?

veewee:
@Landerstraeten,

Thanks for the PR! Looks good but I've added some small comments / questions.

src/Task/FileSize.php
```diff
@@ -0,0 +1,97 @@
...
+        return $this->getConfigurableOptions()->resolve($configured);
+    }
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
```
veewee: Maybe it's a good idea to add a whitelist or blacklist  like e.g. `triggered_by` which whitelists extensions. This way you can still allow for example PDFs, ... to be committed.

src/Task/FileSize.php
Landerstraeten: I have based this on SourceTree which gives a warning for files over 10M.
Github gives a warning for files over 50M and restricts files over 100M.
What do you suggest?

spec/Task/FileSizeSpec.php
Landerstraeten: What do you mean with the failing task? What test do I need to write exactly?

veewee: A test that checks the failed return when a file is bigger than the configured max_size.

src/Task/FileSize.php
veewee: Let's keep it at 10M for now.

src/Task/FileSize.php
lykciv: I'm not sure a whitelist has an added value in this case. Big files will cause problems in git regardless of file type. If you would whitelist PDF files, someone could still commit a 500MB PDF. 

If you have multiple bigger files in your repository, you can always set the max_size higher. And if you find that you have to set the max_size too high, that's probably a sign that you should review whether git is the right solution for storing your bigger files.

veewee:
Thanks for the PR @Landerstraeten !

PR #386, opened 2017-07-28
Let phpcsfixer2 select the correct files and improve performance

veewee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #371, #327

### DIFF
--- a/doc/tasks/phpcsfixer2.md
+++ b/doc/tasks/phpcsfixer2.md
@@ -103,7 +103,8 @@ Cache is supported only for tool downloaded as phar file or installed via compos
 
 Intersection mode can only be used when you have a configuration file which contains a Finder.
 This mode works best since only files that are being commit and are in your configuration will be checked.
-When there is no Finder in your configuration, all committed files will be checked.
+When there is no Finder in your configuration, you'll have set this parameter to false. 
+Otherwise php-cs-fixer will crash the execution.
 
 **verbose**
 
--- a/doc/tasks/phpcsfixer2.md
+++ b/doc/tasks/phpcsfixer2.md
@@ -97,12 +97,13 @@ The tool will fix all files if the tool version has changed or the list of fixer
 Cache is supported only for tool downloaded as phar file or installed via composer.
 
 
-**path_mode**
+**can_intersect**
 
-*Default: null*
+*Default: true*
 
-Specify path mode (can be override or intersection).
-
+Intersection mode can only be used when you have a configuration file which contains a Finder.
+This mode works best since only files that are being commit and are in your configuration will be checked.
+When there is no Finder in your configuration, all committed files will be checked.
 
 **verbose**
 
--- a/resources/config/tasks.yml
+++ b/resources/config/tasks.yml
@@ -232,7 +232,6 @@ services:
         arguments:
           - '@config'
           - '@process_builder'
-          - '@async_process_runner'
           - '@formatter.phpcsfixer'
         tags:
           - {name: grumphp.task, config: phpcsfixer}
--- a/spec/Task/PhpCsFixerV2Spec.php
+++ b/spec/Task/PhpCsFixerV2Spec.php
@@ -78,16 +78,79 @@ class PhpCsFixerV2Spec extends ObjectBehavior
         $this->canRunInContext($context)->shouldReturn(true);
     }
 
-    function it_runs_phpcsfixer2(
+    function it_runs_phpcsfixer2_on_finder_in_run_context_with_intersection(
         GrumPHP $grumPHP,
         ProcessBuilder $processBuilder,
         Process $process,
         RunContext $context,
         PhpCsFixerFormatter $formatter
     ) {
-        $grumPHP->getTaskConfiguration('phpcsfixer2')->willReturn(['config' => '.php_cs']);
+        $grumPHP->getTaskConfiguration('phpcsfixer2')->willReturn([
+            'config' => '.php_cs',
+            'can_intersect' => true,
+        ]);
         $formatter->resetCounter()->shouldBeCalled();
 
+        $context->getFiles()->willReturn(new FilesCollection([
+            $file1 = new SplFileInfo('file1.php', '.', 'file1.php'),
+            $file2 = new SplFileInfo('file2.php', '.', 'file2.php'),
+        ]));
+
+        $processBuilder->createArgumentsForCommand('php-cs-fixer')->willReturn(new ProcessArgumentsCollection());
+        $processBuilder->buildProcess(Argument::that(function (ProcessArgumentsCollection $args) use ($file1, $file2) {
+            return $args->contains('--path-mode=intersection')
+                && !$args->contains($file1->getPathname())
+                && !$args->contains($file2->getPathname());
+        }))->willReturn($process);
+
+        $process->run()->shouldBeCalled();
+        $process->isSuccessful()->willReturn(true);
+
+        $result = $this->run($context);
+        $result->shouldBeAnInstanceOf(TaskResultInterface::class);
+        $result->isPassed()->shouldBe(true);
+    }
+
+    function it_runs_phpcsfixer2_on_all_files_in_run_context_without_intersection(
+        GrumPHP $grumPHP,
+        ProcessBuilder $processBuilder,
+        Process $process,
+        RunContext $context,
+        PhpCsFixerFormatter $formatter
+    ) {
+        $grumPHP->getTaskConfiguration('phpcsfixer2')->willReturn([
+            'config' => '.php_cs',
+            'can_intersect' => false,
+        ]);
+        $formatter->resetCounter()->shouldBeCalled();
+
+        $context->getFiles()->willReturn(new FilesCollection([
+            $file1 = new SplFileInfo('file1.php', '.', 'file1.php'),
+            $file2 = new SplFileInfo('file2.php', '.', 'file2.php'),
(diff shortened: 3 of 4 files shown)


veewee:
Hi @Landerstraeten , @keradus,

This is an attempt to fix the issues in #371 and #327.
Can you give it a review to make sure this works as expected?

Thanks!

spec/Task/PhpCsFixerV2Spec.php
```diff
@@ -107,35 +106,8 @@ function it_runs_the_suite_for_all_files(
         $result->isPassed()->shouldBe(true);
     }
 
-    function it_runs_the_suite_for_changed_files(
```
keradus: why dropping the test instead of replacing it with test of new feature ?

doc/tasks/phpcsfixer2.md
```diff
@@ -97,12 +97,13 @@ The tool will fix all files if the tool version has changed or the list of fixer
...
+**can_intersect**
 
-*Default: null*
-
-Specify path mode (can be override or intersection).
+*Default: true*
 
+Intersection mode can only be used when you have a configuration file which contains a Finder.
+This mode works best since only files that are being commit and are in your configuration will be checked.
+When there is no Finder in your configuration, all committed files will be checked.
```
keradus: when there is no finder in configuration, setting path-mode=intersection will crash the execution

src/Task/PhpCsFixerV2.php
```diff
@@ -82,15 +89,23 @@ public function run(ContextInterface $context)
         }
 
         $arguments->addOptionalArgument('--using-cache=%s', $config['using_cache'] ? 'yes' : 'no');
-        $arguments->addOptionalArgument('--path-mode=%s', $config['path_mode']);
+        $arguments->addOptionalArgument('--path-mode=intersect', $config['can_intersect']);
```
keradus: did you try this out? please add tests to cover this change, as it is not working.
```diff
-intersect
+intersection
```

veewee: We currently don't have a system to do e2e tests with 3rd party tools. I looked over that one during commit and will test these changes thoroughly before merging.

veewee:
Thanks for your feedback @keradus!

I've changed the logic a bit:

- Run context with can_intersect=true will never receive the files since the finder in the php-cs-fixer2 config can be used. (Works better on windows since they have a limited CLI input charakters)
- Run context with can_intersect=false will always retrieve the files since there is no finder in configuration
- PreCommit context with can_intersect=true will always retrieve the files and will run in intersection path mode. When there is no finder in the config file, the exectuable will fail.
- PreCommit context with can_intersect=false will always retrieve the files and will run in override path mode

I assume this will cover all cases at the moment, but will test this thoroughly.

doc/tasks/phpcsfixer2.md
```diff
@@ -103,7 +103,8 @@ Cache is supported only for tool downloaded as phar file or installed via compos
 
 Intersection mode can only be used when you have a configuration file which contains a Finder.
 This mode works best since only files that are being commit and are in your configuration will be checked.
-When there is no Finder in your configuration, all committed files will be checked.
+When there is no Finder in your configuration, you'll have set this parameter to false. 
+Otherwise php-cs-fixer will crash the execution.
```
keradus: btw, could there be consistency with naming? sometimes you use `PHP-CS-Fixer`, sometimes you use `php-cs-fixer` after binary file, while official name is `PHP Coding Standards Fixer` or `PHP CS Fixer` when one used shorter version.

src/Task/PhpCsFixerV2.php
```diff
@@ -89,11 +89,14 @@ public function run(ContextInterface $context)
...
         $arguments->addOptionalArgument('--using-cache=%s', $config['using_cache'] ? 'yes' : 'no');
-        $arguments->addOptionalArgument('--path-mode=intersect', $config['can_intersect']);
+        $arguments->addOptionalArgument('--path-mode=intersection', $config['can_intersect']);
         $arguments->addOptionalArgument('--verbose', $config['verbose']);
         $arguments->addOptionalArgument('--diff', $config['diff']);
         $arguments->add('fix');
-        $arguments->addFiles($files);
+
+        if ($context instanceof GitPreCommitContext || !$config['can_intersect']) {
+            $arguments->addFiles($files);
```
keradus: not sure about condition here. why files are not used if `can_intersect=true` ?

keradus:
> - Run context with can_intersect=true will never receive the files since the finder in the php-cs-fixer2 config can be used. (Works better on windows since they have a limited CLI input charakters)
> - Run context with can_intersect=false will always retrieve the files since there is no finder in configuration

Then, what is the purpose of passing files in grumphp integration together with can_intersect=true ?

you are trying to use some extra logic on top of integration changing decision of user.
if user passed files, let use it.
if user configure mode=intersection, let use it.
that's it

doc/tasks/phpcsfixer2.md
veewee: Sure, I'll try to keep that in mind.

veewee:
> Then, what is the purpose of passing files in grumphp integration together with can_intersect=true ?
> you are trying to use some extra logic on top of integration changing decision of user.
> if user passed files, let use it.
> if user configure mode=intersection, let use it.
> that's it


I disagree: 

In run context, all files are being checked. That means that the task contains ALL files known to git. It is a bit overkill to paste them all the the CLI when there is a finder config in PHP CS Fixer. Besides that: Windows can't handle too many input characters.
So as long as there is a finder in configuration (can_intersect=true), we shouldn't pass the files in run context in my opninion.

The PreCommit context only contains the changed files. In this context, the intersection path-mode is the best option. If this is not possible (can_intersect=false), the override path-mode will be used.

Does this makes sense to you or would you still drop that logic?

keradus:
> So as long as there is a finder in configuration (can_intersect=true), we shouldn't pass the files in run context in my opninion.

so, why this scenario is allowed by code ?

veewee:
I don't really understand your question. Can you give me some additional context?

keradus:
```php
    if ($context instanceof GitPreCommitContext || !$config['can_intersect']) {
        $arguments->addFiles($files);
    }
```

let discuss not-precommit context, thus we have:
```php
    if (!$config['can_intersect']) {
        $arguments->addFiles($files);
    }
```

vs

> In run context, all files are being checked.

No, looking at code, in run-context, the paths are sometimes passed, sometimes not

keradus:
could `can_intersect` be different for run-context and precommit-context?

veewee:
@keradus, I see what you are trying to say. Probably I explained incorrect:

`can_intersect` can't be different for run or pre-commit context, it is always the same since it is in configration. Maybe a better name would be `config_contains_finder` or something like that.

So lets asume `can_intersect` will be renamed to `config_contains_finder`. In that case you have this if:

```php
    if ($context instanceof GitPreCommitContext || !$config['config_contains_finder']) {
        $arguments->addFiles($files);
    }
```

Let's focus on RunContext:
```
    if (!$config['config_contains_finder']) {
        $arguments->addFiles($files);
    }
```

As you can see: the files are being added when there is no finder configuration. If we don't do this, PHP CS Fixer will crash since it isn't validating any files.
When there is a finder in configuration, this one can be used instead of adding all files. This is because it should run on all files configured in the finder anyway.


Off course: for PreComit context, we'll always pass the files since we only want to check a limited subset.

keradus:
cleaner to me, yet I do still not clear enough.

> RunContext:
> As you can see: the files are being added when there is no finder configuration. If we don't do this, PHP CS Fixer will crash since it isn't validating any files.

great, I do agree that if there is no finder, we shall pass the files.

then, you have execution scenario:
- context: RunContext
- config_contains_finder: true
what is gonna happen with current implementation?
you gonna configure `--path-mode=intersection` yet no files for it, and intersecting finder with empty set is... empty set.

doc/tasks/phpcsfixer2.md
```diff
@@ -97,12 +97,14 @@ The tool will fix all files if the tool version has changed or the list of fixer
 Cache is supported only for tool downloaded as phar file or installed via composer.
 
 
-**path_mode**
+**can_intersect**
```
Landerstraeten: I think you forgot to change the example on line 25

resources/config/tasks.yml
```diff
@@ -232,7 +232,6 @@ services:
         arguments:
           - '@config'
           - '@process_builder'
-          - '@async_process_runner'
```
Landerstraeten: You deleted the argument from another task. Delete line 244 instead.

doc/tasks/phpcsfixer2.md
```diff
@@ -97,12 +97,14 @@ The tool will fix all files if the tool version has changed or the list of fixer
...
 
-*Default: null*
-
-Specify path mode (can be override or intersection).
+*Default: true*
 
+Intersection mode can only be used when you have a configuration file which contains a Finder.
+This mode works best since only files that are being commit and are in your configuration will be checked.
+When there is no Finder in your configuration, you'll have set this parameter to false. 
+Otherwise php-cs-fixer will crash the execution.
```
Landerstraeten: This is clear 👍

veewee:
Fixed in #395

PR #398, opened 2017-08-31
Fix exclude_name parameter for phpmnd task

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | -
| Fixed tickets | -

Seems like the documentation for phpmnd ~is~ [was](https://github.com/povils/phpmnd/pull/58) wrong and the option for excluding files is actually named `exclude-file` (_gasp!_).

I chose to keep the name `exclude_name` for GrumPHP though to avoid unnecessary BC break.


veewee:
Thanks for noticing!

PR #399, opened 2017-08-31
Write task progress to error output

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | -
| Fixed tickets | -

Currently things that uses `$this->output` (like `✔` and `✘`) writes to _stdout_ while everything that utilizes ProgressBar usually writes to _stderr_. This PR simply replicates how ProgressBar handles output internally to make output a bit more consistent.

### DIFF
--- a/src/Console/Helper/TaskRunnerHelper.php
+++ b/src/Console/Helper/TaskRunnerHelper.php
@@ -9,6 +9,7 @@ use GrumPHP\Runner\TaskRunner;
 use GrumPHP\Runner\TaskRunnerContext;
 use Symfony\Component\Console\Helper\Helper;
 use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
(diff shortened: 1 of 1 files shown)


src/Event/Subscriber/ProgressSubscriber.php
```diff
@@ -34,6 +35,10 @@ class ProgressSubscriber implements EventSubscriberInterface
      */
     public function __construct(OutputInterface $output, ProgressBar $progressBar)
     {
+        if ($output instanceof ConsoleOutputInterface) {
```
veewee: Shouldn't this be fixed here instead?
https://github.com/phpro/grumphp/blob/master/src/Console/Helper/TaskRunnerHelper.php#L96

Landerstraeten: I also agree with @veewee

veewee:
Hi @jyggen,

Do you agree with the comments above?

jyggen:
Sorry for my inactivity here, but hopefully the PR looks better now!

Whether it's better to output everything to stdout instead of stderr is something worth discussing indeed, but the point of this PR is simply to normalize the usage. I've seen build systems where stdout and stderr are sent to different places/files, and it looks really weird when the task names end up in one place and ✔ and ✘ in another :)

Landerstraeten:
Thank you @jyggen!

PR #401, opened 2017-09-01
Use argument arrays for phpmnd's exclude options

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | -
| Fixed tickets | -

Another fix for the phpmnd task. Unlike some of the other options, the three for excluding files wants one argument per file instead of a comma-separated list (so `--exclude=a.php --exclude=b.php` instead of `--exclude=a.php,b.php`).


veewee:
Thanks for reporting! I am not using that tool myself, but based on the command (InputOption::VALUE_IS_ARRAY), I assume this is correct.

Issue #392, opened 2017-08-09
Add Infection task


Rastusik:
| Q               | A
| --------------- | ---
| Version         | `grumphp -V`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

More info https://infection.github.io/

PR #405, opened 2017-09-11
Add infection task

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #392 

# New Task Checklist:

### DIFF
--- a/doc/tasks/infection.md
+++ b/doc/tasks/infection.md
@@ -9,6 +9,7 @@ It lives under the `infection` namespace and has following configurable paramete
 parameters:
     tasks:
         infection:
+            configuration: ~
             threads: ~
             test_framework: ~
             only_covered: false
@@ -18,6 +19,14 @@ parameters:
             mutators: []
             triggered_by: [php]
 ```
+
+**configuration**
+
+*Default: null*
+
+By defaut the `infection.json.dist` wil be used.
+You can specify an alternate location for this file by changing this option.
+
 
 **threads**
 
--- a/doc/tasks/infection.md
+++ b/doc/tasks/infection.md
@@ -1,0 +1,79 @@
+# Infection
+
+Infection is a PHP mutation testing framework based on Abstract Syntax Tree.
+
+It lives under the `infection` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        infection:
+            threads: ~
+            test_framework: ~
+            only_covered: ~
+            configuration: ~
+            min_msi: ~
+            min_covered_msi: ~
+            mutators: ~
+            triggered_by: [php]
+```
+
+**threads**
+
+*Default: null*
+
+If you want to run tests for mutated code in parallel, set this to something bigger than 1.
+It will dramatically speed up the mutation process.
+Please note that if your tests somehow depends on each other or use a database, this option can lead to failing tests which give many false-positives results.
+
+
+**test_framework**
+
+*Default: null*
+
+This is the name of a test framework to use. Currently Infection supports `PhpUnit` and `PhpSpec`.
+
+
+**only_covered**
+
+*Default: null*
+
+Run the mutation testing only for covered by tests files.
+
+
+**configuration**
+
+*Default: null*
+
+The path or name to the infection configuration file.
+
+
+**min_msi**
+
+*Default: null*
+
+This is a minimum threshold of Mutation Score Indicator (MSI) in percentage.
+
+
+**min_covered_msi**
+
+*Default: null*
+
+This is a minimum threshold of Covered Code Mutation Score Indicator (MSI) in percentage.
+
+
+**mutators**
+
+*Default: null*
+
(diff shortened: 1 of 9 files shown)


doc/tasks/infection.md
```diff
@@ -0,0 +1,79 @@
+# Infection
+
```
veewee: Should we suggest to load `infection/infection` trough composer locally / globally ?
In other task documentation, this is added to the top + in the composer.json file under suggestions.

Since the documentation of infection is suggesting global installation, we might only add it in the task documentation?

src/Task/Infection.php
```diff
@@ -0,0 +1,97 @@
...
+            'configuration' => null,
+            'min_msi' => null,
+            'min_covered_msi' => null,
+            'mutators' => null,
+            'triggered_by' => ['php'],
+        ]);
+
+        $resolver->addAllowedTypes('threads', ['null', 'int']);
+        $resolver->addAllowedTypes('test_framework', ['null', 'string']);
+        $resolver->addAllowedTypes('only_covered', ['null', 'bool']);
```
veewee: Since it's a bool, doesn't it make more sense to set the default to false and only support bool?

src/Task/Infection.php
```diff
@@ -0,0 +1,97 @@
...
+        $resolver->addAllowedTypes('configuration', ['null', 'string']);
+        $resolver->addAllowedTypes('min_msi', ['null', 'integer']);
+        $resolver->addAllowedTypes('min_covered_msi', ['null', 'integer']);
+        $resolver->addAllowedTypes('mutators', ['null', 'string']);
```
veewee: An array might be better here to make the configuration easier to read.
We can always implode the array to a comma separated string

src/Task/Infection.php
```diff
@@ -0,0 +1,97 @@
...
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('infection');
+
+        $arguments->addOptionalArgument('--threads=%s', $config['threads']);
+        $arguments->addOptionalArgument('--test-framework=%s', $config['test_framework']);
+        $arguments->addOptionalArgument('--only-covered', $config['only_covered']);
+        $arguments->addOptionalArgument('--configuration=%s', $config['configuration']);
```
veewee: This option is not documented: https://infection.github.io/guide/command-line-options.html
Does it work?

Landerstraeten: Documentation updated

doc/tasks/infection.md
Landerstraeten: I would suggest to load it locally. I already updated the documentation.

pierredup:
Mutation testing can be a very time consuming and resource intensive task, especially if it runs the entire test suite on every commit. Ideally I think this should filter to only mutate files that have changed (using the `filter` option) and `test-framework-options` needs to be set with the respective values for the test framework to only run test files that have changed.

src/Task/Infection.php
```diff
@@ -0,0 +1,93 @@
...
+    public function run(ContextInterface $context)
+    {
+        $config = $this->getConfiguration();
+        $files = $context->getFiles()->extensions($config['triggered_by']);
+
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('infection');
```
pierredup: Infection requires xdebug or phpdbg to be installed, so I think this task should be skipped if at least one of those extensions is not installed

src/Task/Infection.php
```diff
@@ -0,0 +1,93 @@
...
+
+        $arguments->addOptionalArgument('--threads=%s', $config['threads']);
+        $arguments->addOptionalArgument('--test-framework=%s', $config['test_framework']);
+        $arguments->addOptionalArgument('--only-covered', $config['only_covered']);
+        $arguments->addOptionalArgument('--configuration=%s', $config['configuration']);
+        $arguments->addOptionalArgument('--min-msi=%s', $config['min_msi']);
+        $arguments->addOptionalArgument('--min-covered-msi=%s', $config['min_covered_msi']);
+        $arguments->addOptionalCommaSeparatedArgument('--mutators=%s', $config['mutators']);
+
+        $process = $this->processBuilder->buildProcess($arguments);
```
pierredup: Interaction should be disabled (with the `-n` flag)

src/Task/Infection.php
Landerstraeten: I don't think grumphp is responsible to check that. If you don't have xdebug or phpdbg enabled, your tests will not run anyway.

doc/tasks/infection.md
veewee: Can you add the composer require command to the documentation as well? e.g. https://github.com/phpro/grumphp/blob/master/doc/tasks/phpunit.md

Issue #406, opened 2017-09-12
max_body_width setting of git task also is applied to commented lines


schmengler:
| Q               | A
| --------------- | ---
| Version         | 0.12.0
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

The `max_body_witdth` check for git commit messages should exclude lines starting with `#`. These are comments that do not end up in the commit and often are automatically generated with long lines, for example in merge commits or with a localized git installation.

**My configuration**
```yaml
# grumphp.yml
parameters:
    tasks:
        git_commit_message:
            max_subject_width: 80
            max_body_width: 80
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer require --dev phpro/grumphp

# Your actions

git commit --allow-empty -m "Test

# this looks like a long line, but it won't actually end up in the commit. I am writing random stuff now to reach the limit. Knowledge is power, France is bacon.
"
```

**Result:**
```
Line 3 of commit message has > 80 characters.
```

veewee:
Fixed in #407

PR #407, opened 2017-09-13
Ignore max body width for commented lines in the commit message

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | /
| Fixed tickets | #406

### DIFF
--- a/src/Task/Git/CommitMessage.php
+++ b/src/Task/Git/CommitMessage.php
@@ -164,6 +164,12 @@ class CommitMessage implements TaskInterface
 
         $errors = [];
         $lines = preg_split('/\R/u', $commitMessage);
+        foreach ($lines as $index => $line) {
+            if (strpos($line, '#') === 0) {
+                unset($lines[$index]);
+            }
+        }
+
         $subject = rtrim($lines[0]);
         $maxSubjectWidth = $config['max_subject_width'] + $this->getSpecialPrefixLength($subject);
 
@@ -172,10 +178,6 @@ class CommitMessage implements TaskInterface
         }
 
         foreach (array_slice($lines, 2) as $index => $line) {
-            if (strpos($line, '#') === 0) {
-                continue;
-            }
-
             if (mb_strlen(rtrim($line)) > $config['max_body_width']) {
                 $errors[] = sprintf(
                     'Line %u of commit message has > %u characters.',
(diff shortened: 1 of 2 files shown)


src/Task/Git/CommitMessage.php
```diff
@@ -172,6 +172,10 @@ private function enforceTextWidth(ContextInterface $context)
         }
 
         foreach (array_slice($lines, 2) as $index => $line) {
+            if (strpos($line, '#') === 0) {
```
veewee: It's also possible that the comment is above the message (e.g. if you rebase)
This might result in a subject line being a comment.

Maybe the comments should always be dropped before doing any of these commit message cehcks instead of only during the max length check?

src/Task/Git/CommitMessage.php
```diff
@@ -164,6 +164,12 @@ private function enforceTextWidth(ContextInterface $context)
 
         $errors = [];
         $lines = preg_split('/\R/u', $commitMessage);
+        foreach ($lines as $index => $line) {
```
jyggen: Maybe use `array_filter()`?
```php
$lines = array_filter($lines, function ($line) {
    return strpos($line, '#') !== 0;
});
```

src/Task/Git/CommitMessage.php
```diff
@@ -164,6 +164,12 @@ private function enforceTextWidth(ContextInterface $context)
...
+            if (strpos($line, '#') === 0) {
+                unset($lines[$index]);
+            }
+        }
+
         $subject = rtrim($lines[0]);
```
jyggen: This line will throw a notice if index 0 is filtered away above.

veewee: You cannot have an empty commit message right? Or are there some dark ways I don't know about ... ?

src/Task/Git/CommitMessage.php
veewee: I also think this removal should not be done inside the `enforceTextWidth()` method. It might also break other checks like e.g. `subjectHasTrailingPeriod()`. It's probably best to move this normalisation to a separate method and run that method here:
https://github.com/Landerstraeten/grumphp/blob/89ca2b71fe6073346ca00fec09109a1727795b7e/src/Task/Git/CommitMessage.php#L101

veewee:
Thanks for the PR @Landerstraeten

Kanti:
@veewee Could you plases release this bugfix. The bug is very disturbing
Thanks a lot!

PR #411, opened 2017-10-05
Only add DIRECTORY_SEPARATOR if needed

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | -
| Fixed tickets | -

The currently code isn't so friendly to relative paths since it always adds a `DIRECTORY_SEPARATOR`.

For example if I have `phpunit.xml` in the root of the project and add `vendor/phpunit/phpunit/phpunit.xsd` as a scheme in it, the linter would end up resolving its path as `/vendor/phpunit/phpunit/phpunit.xsd`. This is due to `$xmlFile->getPath()` returning an empty string, and thus the enforced `DIRECTORY_SEPARATOR` turns it into an absolute path.

This change will make sure that `DIRECTORY_SEPARATOR` is only added when the file actually has a path, and the `rtrim()` is there to make sure we don't end up with some `//` weirdness.

### DIFF
--- a/src/Linter/Xml/XmlLinter.php
+++ b/src/Linter/Xml/XmlLinter.php
@@ -237,7 +237,10 @@ class XmlLinter implements LinterInterface
             return $this->loadFromNet ? $scheme : null;
         }
 
-        $schemeFile = new SplFileInfo($xmlFile->getPath() . DIRECTORY_SEPARATOR . $scheme);
+        $xmlFilePath = $xmlFile->getPath();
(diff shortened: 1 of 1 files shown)


jyggen:
Not entirely sure why AppVeyor fails.

veewee:
@jyggen, I've restarted the apveyor build : all good. Can you apply the requested changes from @Landerstraeten  so that we can merge this one in? Thanks!

jyggen:
@veewee Done!

veewee:
Thanks!

PR #426, opened 2017-11-16
Removing unneeded phpdocblocks and updating useful phpdoc blocks.

SerkanYildiz:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | n/a

Removing unneeded phpdoc blocks when used type hinting in method name. Also improving the phpdoc blocks when they are useful. Still work in progress.


src/Console/Command/ConfigureCommand.php
```diff
@@ -24,31 +24,18 @@ class ConfigureCommand extends Command
 {
     const COMMAND_NAME = 'configure';
 
-    /**
-     * @var GrumPHP
-     */
+    /** @var GrumPHP */
```
Landerstraeten: I would leave it as it is

src/Console/Command/ConfigureCommand.php
```diff
@@ -24,31 +24,18 @@ class ConfigureCommand extends Command
...
     protected $repository;
 
-    /**
-     * @var InputInterface
-     */
+    /** @var InputInterface */
     protected $input;
 
-    /**
-     * @param GrumPHP    $config
```
Landerstraeten: I would also leave it as it is, since we still use PHP5.
_This actually counts for every removed docblock in your PR._

src/Console/Command/ConfigureCommand.php
```diff
@@ -80,7 +80,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         }
 
         // Check write action
-        $written = $this->writeConfiguration($configuration);
```
Landerstraeten: Why would you remove the variable? It's used in the next line.

SerkanYildiz:
Okay, I'm fine with it. Indeed we should asap drop support for php 5.6

PR #432, opened 2017-11-21
Add parse_constant and parse_custom_tags options to yamllint task

pierredup:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | N/A

In Symfont 3.2 and 3.3, the Yaml component added support for PHP constants and custom tags.
Currently when using constants or custom tags in a Yaml file, the yamllint task fails because the syntax is invalid (and the necessary flags is not passed through).

This PR adds 2 new options to the yamllint task: `parse_constant` & `parse_custom_tags`. These options default to false to keep current behaviour. If they are set to `true`, then the required flags is passed through to the Yaml::parse method so that the task doesn't fail.

### DIFF
--- a/doc/tasks/yamllint.md
+++ b/doc/tasks/yamllint.md
@@ -11,6 +11,8 @@ parameters:
             ignore_patterns: []
             object_support: false
             exception_on_invalid_type: false
+            parse_constant: false
+            parse_custom_tags: false
 ```
 
 **ignore_patterns**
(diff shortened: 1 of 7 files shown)


veewee:
Looks great! Thanks for the addition.

PR #435, opened 2017-11-29
Remove proxy manager

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? |  no
| Documented?   | no
| Fixed tickets | /

As the title says, remove the proxy manager.

@veewee (and everyone else), please leave your remaks below.

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -4,7 +4,7 @@ namespace GrumPHP\Console;
 
 use GrumPHP\Configuration\ContainerFactory;
 use GrumPHP\Exception\RuntimeException;
-use GrumPHP\IO\IOInterface;
+use GrumPHP\IO\ConsoleIO;
 use GrumPHP\Locator\ConfigurationFile;
 use GrumPHP\Util\Composer;
 use GrumPHP\Util\Filesystem;
@@ -14,6 +14,7 @@ use Symfony\Component\Console\Application as SymfonyConsole;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\ConsoleOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
@@ -42,9 +43,6 @@ class Application extends SymfonyConsole
      */
     protected $composerHelper;
 
-    /**
-     * Set up application:
-     */
     public function __construct()
     {
         $this->filesystem = new Filesystem();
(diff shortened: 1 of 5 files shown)


src/Console/Application.php
```diff
@@ -51,6 +52,8 @@ public function __construct()
         $this->filesystem = new Filesystem();
         $this->container = $this->getContainer();
         $this->setDispatcher($this->container->get('event_dispatcher'));
+        $this->container->set('console.input', new ArgvInput());
```
veewee: This won't work since the output / input wont be configured meaning the verbosity level is not available on the output in the service `grumphp.io.console` (which is still marked as lazy). The `grumphp.io` is required before the `configureIO()` method was called.
You can validate this by running the code with the verbose flag `-v`: You won't see the additional information.

veewee:
Hi @Landerstraeten,
Can you fetch the latest master branch and rebase your branch against master?

EugeneZub:
Hi @Landerstraeten,
Could you give me some forecasts about date when the new version will be released (0.14.0)?
Because we are waiting for this new version with our team and in another case we will have to refuse using of this vendor((

veewee:
@EugeneZub,
Lander told me there are still some minor issues with this PR when using low dependencies. We would like to get this one in for 0.14. Like most open-source projects, we don't set ourselves deadlines and won't be able to answer to your forecast question at this point.

EugeneZub:
Hi @veewee,
Thanx for your explanations, we'll be watching for changes because this vendor is really good!)

src/Console/Application.php
```diff
@@ -146,35 +155,33 @@ protected function getContainer()
...
+        $io = $this->container->get('grumphp.io.console');
 
-        $container = $this->getContainer();
+        $input = $io->getInput();
+        $output = $io->getOutput();
 
-        // Register the console input and output to the container
-        $container->set('console.input', $input);
-        $container->set('console.output', $output);
+        parent::configureIO($input, $output);
```
veewee: This won't work since you now configure the input and output from the container and not the ones that are available in the commands. This means the arguments and options won't work in the commands. As you can see, the configureIO() does not return anything and should just configure the arguments that are being passed to the method.

src/Console/Application.php
```diff
@@ -49,11 +51,18 @@ public function __construct()
...
         $this->container = $this->getContainer();
-        $this->setDispatcher($this->container->get('event_dispatcher'));
 
         parent::__construct(self::APP_NAME, self::APP_VERSION);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function setDispatcher(EventDispatcherInterface $dispatcher)
```
veewee: Why did you move this one outside the constructor? It seems better to set a dispatcher then to overwrite thet set* method.

src/IO/ConsoleIO.php
```diff
@@ -92,6 +92,22 @@ public function writeError($messages, $newline = true)
         $this->doWrite($messages, $newline, true);
     }
 
+    /**
+     * @return InputInterface
+     */
+    public function getInput()
```
veewee: Why did you add these getters? They are not being used at the moment, so I would not make them available at this point.

PR #428, opened 2017-11-18
Add Symfony 4 support

UrGuardian4ngel:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #423

This PR adds support for the newest version of Symfony 4.0.

### Additional information

#SymfonyConHackday2017 :)

This PR will depend on the adoption of Symfony v4.0 in some of its dependencies:

- [x] composer/composer#6788
- [ ] gitonomy/gitlib#122
- [x] phpspec/phpspec#1166

PR #437, opened 2017-12-27
Fix deprecated ProcessBuilder

daaru00:
I have trouble using grumphp with Laravel 5.5 (related issue https://github.com/laravel/framework/issues/20995#issuecomment-327647935).

The problem seems the usage of deprecated method to exec process, instead using `ProcessBuilder` I used  directly `new Process` and this fix the issue.

### DIFF
--- a/src/Composer/DevelopmentIntegrator.php
+++ b/src/Composer/DevelopmentIntegrator.php
@@ -4,7 +4,7 @@ namespace GrumPHP\Composer;
 
 use Composer\Script\Event;
 use GrumPHP\Util\Filesystem;
-use Symfony\Component\Process\ProcessBuilder;
+use Symfony\Component\Process\Process;
 
 class DevelopmentIntegrator
 {
@@ -23,7 +23,7 @@ class DevelopmentIntegrator
             self::noramlizePath($composerExecutable)
         );
 
-        $process = ProcessBuilder::create([$composerExecutable, 'git:init'])->getProcess();
+        $process = new Process($composerExecutable.' git:init');
         $process->run();
         if (!$process->isSuccessful()) {
             $event->getIO()->write(
--- a/src/Composer/GrumPHPPlugin.php
+++ b/src/Composer/GrumPHPPlugin.php
@@ -19,7 +19,7 @@ use GrumPHP\Console\Command\Git\DeInitCommand;
 use GrumPHP\Console\Command\Git\InitCommand;
 use GrumPHP\Locator\ExternalCommand;
 use Symfony\Component\Process\ExecutableFinder;
-use Symfony\Component\Process\ProcessBuilder;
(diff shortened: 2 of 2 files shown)


veewee:
This is also fixed in #428.

daaru00:
Happy to hear that :+1:

Snaver:
Confirmed fixes direct ProcessBuilder error 👍

daaru00:
@UrGuardian4ngel I understood what your are saying, this was a quick-and-dirty fix. 

Reading [Symfony Process Component documentation](https://symfony.com/doc/current/components/process.html) I noticed is possible to pass an array instead of a single string, maybe using this approach the arguments will be escaping. 

Replacing this:
```php
$process = new Process($executable.' '.$command.' --no-interaction');
```
to this:
```php
$process = new Process([$executable, $command, '--no-interaction']);
```

UrGuardian4ngel:
**Correct.** That "array notation" is indeed internally escaping all arguments...  
**_But_** the component accepts only arrays from a specific version (3.3+), so it will break in all versions before that. And we're still supporting the "legacy" Symfony components (version `~2.7 -> 2.8.x`) for now.

These deprecations in Symfony 3.3 API (w/ breaking changes in 4.0), have already been made in #428 (in a backward-compatible way) so they should work for _all_ Symfony versions this project supports. And not only for `symfony/process`, but also for all other affected Symfony components...  
I think that's the reason why this PR has a **duplicate** tag, @veewee? Maintainers will probably be able to give us a bit more information about what's going to happen next? And maybe even a "timeline"? But not sure about that...

---

In the meantime, for more information, please check:

- [src/Process/ProcessFactory](https://github.com/phpro/grumphp/pull/428/files?diff=unified#diff-9a08194c7e449f5d3951cb3814bf513a), which you can use to construct a `Process` (in a backward compatible way where necessary) from GrumPHP's `ProcessArgumentsCollection`.  
Here, all arguments are currently 'manually' escaped w/ Symfony 4.0's escape "helper" implementation. This implementation can be simplified to pass in a simple array (as you saw) from the moment support for versions lower than `symfony/process: ^3.3` get dropped.
- [src/Process/ProcessBuilder](https://github.com/phpro/grumphp/pull/428/files?diff=unified#diff-1b45e15f19108f48f0bfe7672c76f9ce) removes the now-deprecated Symfony `ProcessBuilder`, in favor of immediately creating processes. But we'll probably keep around GrumPHP's ProcessBuilder utility for now?

And maybe you can give that `symfony-4` branch a spin, to see if it solves your problems? ;)

veewee:
Hi @UrGuardian4ngel, @daaru00,

As mentioned above: this is fixed in the symfony 4 branch.
I am currently running it om some projects to see if it works. Will probably merge it in a couple of days.

I'll close this one for now.

PR #434, opened 2017-11-26
Bump to PHP 7

TomasVotruba:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | no

### Why?

- majority of PHP [world bumping to PHP 7.1](https://gophp71.org/)
- upcoming [PHP 7.2](https://github.com/php-ai/php-ml/pull/147) in few days
- we could add PHP 7.0/7.1 typehints via coding standards] more easily in the future
- [as measured on Packagist - 67 % people](https://seld.be/notes/php-versions-stats-2017-2-edition) are on PHP 7.0+ and growing
- more Grumpy :rocket: 

What do you think?

PR #440, opened 2018-01-10
bump gitlib requirement

keradus:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | kind of
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

This is a must have for : 
- Sf4 support (#428), as without it composer will refuse to install v4
- PHP7 support (#434), as without it we will have ugly crashes on PHP7 - bug was fixed on gitlib side and just released

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "composer-plugin-api": "~1.0",
         "composer/composer": "^1.0",
         "doctrine/collections": "~1.2",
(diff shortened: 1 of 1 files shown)


keradus:
thanks @veewee :+1:

veewee:
Thanks for the work on gitlib and grumphp ! :)

PR #450, opened 2018-01-24
Ignore symlinks when registering files.

dzschille:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes| BC breaks?    | yes/no
| Deprecations? | /no
| Documented?   | no
| Fixed tickets | n/a

While using the **file_size** task i got this unexpected message:
```
Large files detected:
- images exceeded the maximum size of 10M.
```
"images" is a symlink to my assets:
`lrwxrwxrwx 1 user user   17 Jan 24 08:45 images -> ../assets/images/`
The task **file_size** does not complain about the directory "assets/images" but about the symlink "images".

It makes no sense to add files which are symlinks to the collection of registered files because we do not need to check the files twice, they will be tested on their real location.

### DIFF
--- a/src/Locator/RegisteredFiles.php
+++ b/src/Locator/RegisteredFiles.php
@@ -31,9 +31,8 @@ class RegisteredFiles
 
         $files = [];
         foreach ($filePaths as $file) {
-            $fileInfo = new SplFileInfo($file, dirname($file), $file);
-
-            if (! $fileInfo->isLink()) {
+            $fileInfo = new SplFileInfo($file, \dirname($file), $file);
+            if (!$fileInfo->isLink()) {
                 $files[] = $fileInfo;
             }
         }
--- a/src/Locator/RegisteredFiles.php
+++ b/src/Locator/RegisteredFiles.php
@@ -31,7 +31,11 @@ class RegisteredFiles
 
         $files = [];
         foreach ($filePaths as $file) {
-            $files[] = new SplFileInfo($file, dirname($file), $file);
+            $fileInfo = new SplFileInfo($file, dirname($file), $file);
+
+            if (! $fileInfo->isLink()) {
+                $files[] = $fileInfo;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


veewee:
Hi @dzschille,

Thanks for contributing!
Can you also update the spec tests?
During pre-commit, the files are located through the ChangedFiles Locator. This one should also be updated.

dzschille:
Hi @veewee , i already thought about how to test this but i have no idea yet. We are not testing with real files in the spec tests and i don't know how i can mock a file so that _RegisteredFiles->locate()_ will have the mock object for _SplFileInfo()_. Or does it make sense to use vfsStream here?
I am willing to learn. :)

veewee:
Hi @dzschille,

Sorry for the late feedback.
Maybe there are some tasks in which you DO want to check something on symlinks.
Isn't it better to add a `noSymlinks()` filter on the `FilesCollection` and call that method in the `file_size` task to make sure no symlinks are checked for filesize?

dzschille:
Hi @veewee ,

i have no idea for a scenario where i need to test symlinked files. I see the two possibilities:

1. My symlinks link to files in my project. The original files will already be tested so no need to test also the symlinks.
2. My symlinks link to files outside of my project. So i don't want to check them.

Do you have an idea about a scenario where you like to check also the symlinks?

veewee:
For example: https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/check_symlinks.py

dzschille:
@veewee , yeah, i see that it makes sense not to ignore the symlinks when you want to check them, so your idea with a ignoreSymlinks() filter makes totally sense. I will try to make the filter when i have some time.

Landerstraeten:
Thank you @dzschille for reporting this issue!
In the meantime, I fixed it myself on a different way.

dzschille:
@Landerstraeten that's great!

PR #451, opened 2018-01-25
Filter home folder for getting absolute file paths

jplou-onestic:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

When setting grumphp.yml config or failed/success ascii file paths in a home folder, grumphp is getting absolute path to files wrong.

I.e. if in grumphp.yml success ascii file is defined in  _~/test-project/grumphp/grumphp-happy.txt_
method will filter path to _/home/johndoe/test-project/grumphp/grumphp-happy.txt_

### DIFF
--- a/src/Console/Helper/PathsHelper.php
+++ b/src/Console/Helper/PathsHelper.php
@@ -16,6 +16,7 @@ use Symfony\Component\Console\Helper\Helper;
 class PathsHelper extends Helper
 {
     const HELPER_NAME = 'paths';
+    const HOME_CHAR = '~';
 
     /**
      * @var GrumPHP
@@ -97,6 +98,7 @@ class PathsHelper extends Helper
     public function getAsciiContent($resource)
     {
         $file = $this->config->getAsciiContentPath($resource);
+        $file = $this->filterHomeFolder($file);
 
         // Disabled:
         if (is_null($file)) {
@@ -246,6 +248,7 @@ class PathsHelper extends Helper
     public function getAbsolutePath($path)
     {
         $path = trim($path);
+        $path = $this->filterHomeFolder($path);
         $realpath = realpath($path);
         if (false === $realpath) {
             throw new FileNotFoundException($path);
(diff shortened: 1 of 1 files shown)


jplou-onestic:
As I'm start using grumphp in a light way, I don't know if there's a native way to set this paths :)

veewee:
Hi @jplou-onestic,

Thanks for your work.

I am not sure what you are trying to achieve with this: the idea is that grumphp is added as a dev  dependency to a project. There is no way we can make sure that an ascii image is available in the home folder of every user who works on the project.

I would suggest to move the ascii's to the project you are working on or add it to a grumphp extension instead of pointing it to a home folder. If you are working by yourself, you could just hardcode the absolute path instead.

Any feedback on the topic is highly appreciated.

jplou-onestic:
Hello @veewee!
The point is, as I am working with other developers in almost the same kind of projects with the same kind of validation, we want to unify all configuration files for checking code in one repository (grumphp.yml and ascii files for start with).
In this case, grumphp will be installed globally and initiated in the "exotic way" with grumphp.yml in config repository (located in a ~ subfolder). If we need to add more validation tasks just will update grumphp.yml file in that repo.

Anyway, I'm gonna try composing grumphp with our config repo in our projects.
Please, close PR if you consider it.
Thanks for your time! :)

src/Console/Helper/PathsHelper.php
```diff
@@ -283,4 +286,23 @@ public function getName()
...
+    /**
+     * @param $path
+     *
+     * @return mixed
+     */
+    private function filterHomeFolder($path)
+    {
+        $homeCharPos = strpos($path, self::HOME_CHAR);
+
+        if ($homeCharPos === false) {
```
Landerstraeten: I recommend [yoda](https://en.wikipedia.org/wiki/Yoda_conditions)

veewee:
I don't see any additional value in this PR, so I will close this one. 
At this page, you can find the recommended way of sharing configuration + ascii files: https://github.com/phpro/grumphp/blob/master/doc/conventions.md instead.

Thanks for your work anyways!

Issue #447, opened 2018-01-21
Path issue with git hooks


Snaver:
| Q               | A
| --------------- | ---
| Version         | 0.13.1
| Bug?            | yes
| New feature?    | no
| Question?       | yes
| Documentation?  |no
| Related tickets | 

Seems there's an error with git hooks on windows? Are slashes being stripped?

Same error using both Sourcetree and gitkraken:

```git -c diff.mnemonicprefix=false -c core.quotepath=false commit -q -F C:\Users\XXX\AppData\Local\Temp\itngomnn.xpn
.git/hooks/pre-commit: line 12: exec: C:UsersXXXRepositoriesXXX-laravelvendorbingrumphp: not found
Completed with errors, see above.
```

Can run `.\vendor\bin\grumphp run` from Powershell with no issues.

* PHP 7.2.0
* Windows 10 
* git version 2.15.1.windows.2

pedro-p:
Same issue for me.
Adding double quotation marks to pre-commit seems to fix this on windows: https://github.com/phpro/grumphp/pull/458/commits/c774186c4d47e9c44df75484d30c65d3d7a64894

Ainaras:
Same issue for me too. I can't start using this in my org, until this is fixed. 
Patch from pedro-p does the thing. Can't wait until this patch is merged.

veewee:
Can you confirm that #463 fixes the issue?

Snaver:
Just pulled in the fix-force-unix-hooks branch, seemed to fix the issue for me 👍

veewee:
Thanks for checking @Snaver ! Fixed in #463

PR #458, opened 2018-02-02
Fix not found error in Windows

pedro-p:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

without this fix on Windows 10 Pro (Powershell) the pre-commit hook fails with error:
.git/hooks/pre-commit: line 12: exec: C:UsersDellwwwelevationvendorbingrumphp: not found

### DIFF
--- a/resources/hooks/local/pre-commit
+++ b/resources/hooks/local/pre-commit
@@ -9,4 +9,4 @@
 DIFF=$(git diff -r -p -m -M --full-index --staged | cat)
 
 # Run GrumPHP
-(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | exec $(HOOK_COMMAND) '--skip-success-output')
+(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | exec "$(HOOK_COMMAND)" '--skip-success-output')


pedro-p:
Probably this is not the right way to fix #447 as it may break Linux where the command is already surrounded by single quotes :(

veewee:
Thanks for investigating!
This is indeed not the way to fix the problem.
Can you tell me what your git hook looks like? Normally, the hook command should be escaped since it is being returned from the `symfony/process` component.

pedro-p:
Yes, on Windows it it looks like this:

`#!/bin/sh

#
# Run the hook command.
# Note: this will be replaced by the real command during copy.
#

DIFF=$(git diff -r -p -m -M --full-index --staged | cat)

# Run GrumPHP
(cd "./" && printf "%s\n" "${DIFF}" | exec C:\Users\Dell\www\myproject\vendor\bin\grumphp git:pre-commit '--skip-success-output')
`

veewee:
Can you tell me which version of symfony/process you are using? (`composer show`)
Internally, we use this component to generate the command. See: `\GrumPHP\Console\Command\Git\InitCommand::generateHookCommand()`
Normally that component should add quotes around the cli arguments - which does not happen.

pedro-p:
symfony/process v2.8.33

veewee:
We've added a compatibility layer for symfony/process to make it compatible with Symfony 2, 3 and 4.
Probably that is where it goes wrong. Currently I am not owning a windows computer anymore, so there is no easy way for me to find out what exactly goes wrong.

My best guess is that the issue is located at one of these positions:

- `\GrumPHP\Process\ProcessFactory::fromArguments()`
- `\GrumPHP\Process\ProcessUtils::escapeArgument()`

Can you validate the `$commandline` string that is being built in the ProcessFactory method?

golddragon007:
The problem is with C:\Users\Dell\www\myproject\vendor\bin\grumphp is the "\" symbol which meaning in linux and in windows the escape the next character. Also the same error exist in the commit-msg file.
Also in windows you need to use "" if your patch contains space character.
You can use \\ instead of \ and "" but, than it will break if the user has a path which contains space.

veewee:
Closing this one. Lets continue fixing it in #463

PR #460, opened 2018-02-05
Add ignore_parse_errors option to ComposerRequireChecker

jyggen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

This adds an option to enable `--ignore-parse-errors`, which is available in ComposerRequireChecker since 0.2.0.

One thing to note is that ~the world will crash and burn~ ComposerRequireChecker will throw a hissy fit if you have <0.2.0 installed and enable this. I think pointing this out in the docs should be good enough since the option is opt-in. We could always go down the route of version detection, but `composer-require-checker --version` identifies itself as `0.1.1-dev` on 0.2.0 :(

### DIFF
--- a/doc/tasks/composer_require_checker.md
+++ b/doc/tasks/composer_require_checker.md
@@ -17,6 +17,7 @@ parameters:
         composer_require_checker:
             composer_file: 'composer.json'
             config_file: ~
+            ignore_parse_errors: false
             triggered_by: ['composer.json', 'composer.lock', '*.php']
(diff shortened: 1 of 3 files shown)


veewee:
Looks good to me. The documentation is clear to me. I would also avoid the version detection.

PR #461, opened 2018-02-06
Use correct path to .git/COMMIT_EDITMSG when using non-standard .git dir location

pascal-hofmann:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | -

When grumphp is used with a `.git` dir in a non-standard location the `commit-msg` hook may look for the `.git/COMMIT_EDITMSG` file in the wrong location.

This PR prefixes the path with `git_dir` configured in `grumphp.yml` if it's a relative path, so the correct path is used in all cases.

Note: The path supplied to the hook by git may be absolute. In this case there is no need to fix it. This happens when grumphp is used inside a git submodule.

### DIFF
--- a/src/Console/Command/Git/CommitMsgCommand.php
+++ b/src/Console/Command/Git/CommitMsgCommand.php
@@ -78,6 +78,11 @@ class CommitMsgCommand extends Command
         $gitUser = $input->getOption('git-user');
         $gitEmail = $input->getOption('git-email');
         $commitMsgPath = $input->getArgument('commit-msg-file');
+
+        if (!$this->filesystem->isAbsolutePath($commitMsgPath)) {
+            $commitMsgPath = $this->paths()->getGitDir() . $commitMsgPath;
+        }
+
         $commitMsgFile = new SplFileInfo($commitMsgPath);
         $commitMsg = $this->filesystem->readFromFileInfo($commitMsgFile);
 


veewee:
Hi @pascal-hofmann,

Thanks for the PR.

Can you give me steps to reproduce this issue?
The git hook is passing the commit message file directly to grumphp, so this shouldn't be an issue.

See:
https://github.com/phpro/grumphp/blob/master/resources/hooks/local/commit-msg#L10

pascal-hofmann:
Yep, but the path may be relative and directory is changed, so the relative path is incorrect afterwards. Will post steps/versions required to reproduce.

pascal-hofmann:
Steps / versions to reproduce:

phpro/grumphp (v0.13.1)
git version 2.15.1

```
mkdir issue461/
cd issue461/
git init
mkdir app
cd app
echo -e "parameters:\n    git_dir: .." >grumphp.yml
composer require --dev phpro/grumphp
git add composer.json composer.lock grumphp.yml
git commit -m 'Added grumphp'
```

Result for me is this error message on red background:
>  SplFileInfo::openFile(.git/COMMIT_EDITMSG): failed to open stream: No such
>  file or directory

pascal-hofmann:
This PR has now been open for 4 weeks. How can I help to get this merged?

ddebasilio:
This is happening to me too, can this get merged ?

pascalhofmann:
Thanks for merging! 👍

PR #462, opened 2018-02-07
[FIX] Windows path issue

SerkanYildiz:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | n/a
| Fixed tickets | #447

The path to the executable wasn't quoted by escapeArgument, it will be quoted now if the  hook is generated on a Windows platform.

Hope this fixes the problem (worked on my old-student-laptop)

#AKebabADayKeepsTheBugsAway /ping @veewee

PR #463, opened 2018-02-08
Force git hooks to contain unix styled commands

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #458, #462, #447

This bug was introduced in version 0.13 : before we used to force the git hook command to be unix styled. 

(reference: https://github.com/phpro/grumphp/commit/415fd1a9c70f5905326f0651ac5009a70d42441e#diff-aed6b638f2d7de8614c3516d7939c1a0L135 )

### DIFF
--- a/spec/Process/ProcessBuilderSpec.php
+++ b/spec/Process/ProcessBuilderSpec.php
@@ -27,9 +27,19 @@ class ProcessBuilderSpec extends ObjectBehavior
 
     function it_should_be_able_to_create_process_arguments_based_on_taskname(ExternalCommand $externalCommandLocator)
     {
-        $externalCommandLocator->locate('grumphp')->willReturn('/usr/bin/grumphp');
+        $externalCommandLocator->locate('grumphp', false)->willReturn('/usr/bin/grumphp');
 
-        $arguments = $this->createArgumentsForCommand('grumphp');
+        $arguments = $this->createArgumentsForCommand('grumphp', false);
+        $arguments->shouldHaveType(ProcessArgumentsCollection::class);
+        $arguments[0]->shouldBe('/usr/bin/grumphp');
+        $arguments->count()->shouldBe(1);
+    }
+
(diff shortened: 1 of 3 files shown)


spec/Process/ProcessBuilderSpec.php
```diff
@@ -27,9 +27,19 @@ function it_is_initializable()
 
     function it_should_be_able_to_create_process_arguments_based_on_taskname(ExternalCommand $externalCommandLocator)
     {
-        $externalCommandLocator->locate('grumphp')->willReturn('/usr/bin/grumphp');
+        $externalCommandLocator->locate('grumphp', false)->willReturn('/usr/bin/grumphp');
 
-        $arguments = $this->createArgumentsForCommand('grumphp');
+        $arguments = $this->createArgumentsForCommand('grumphp', false);
+        $arguments->shouldHaveType(ProcessArgumentsCollection::class);
+        $arguments[0]->shouldBe('/usr/bin/grumphp');
```
Landerstraeten: `$arguments->shouldHaveKeyWithValue(0, '/usr/bin/grumphp');`

src/Process/ProcessBuilder.php
```diff
@@ -43,11 +43,12 @@ public function __construct(GrumPHP $config, ExternalCommand $externalCommandLoc
     /**
      * @param string $command
      *
+     * @param bool $forceUnix
```
Landerstraeten: I think this should be on line 45 (swap the newline)

spec/Process/ProcessBuilderSpec.php
veewee: To me, it's more readable as it is right now

Landerstraeten: 👍

Issue #469, opened 2018-02-16
PhpCsFixerV2 is overriding .php_cs config values with its own defaults


vyshkant:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.13.1`
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no

**My configuration**
```yaml
# grumphp.yml
parameters:
    tasks:
        phpcsfixer2:
            config: .php_cs
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer require --dev phpro/grumphp

# create .php_cs file with `setRiskyAllowed(true)` and `setUsingCache(false)`

# create a php file breaking rules defined in .php_cs

# Run GrumpHP:
git add -A && git commit -m "Test"
```

**Result:**
```
GrumPHP detected a pre-commit command.
GrumPHP is sniffing your code!
Running task 1/1: PhpCsFixerV2... ✘
You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.
If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
Loaded config default from ".php_cs".

In ConfigurationResolver.php line 345:
                                                                               
  [PhpCsFixer\ConfigurationException\InvalidConfigurationException (16)]       
  The rules contain risky fixers (no_php4_constructor, void_return, pow_to_ex  
  ponentiation, declare_strict_types, function_to_constant, is_null, non_prin  
  table_character, random_api_migration, dir_constant, ereg_to_preg, moderniz  
  e_types_casting, no_alias_functions, no_homoglyph_names, silenced_deprecati  
  on_error, strict_comparison, strict_param, php_unit_construct, psr4), but t  
  hey are not allowed to run. Perhaps you forget to use --allow-risky option?  
                                                                               

Exception trace:
 PhpCsFixer\Console\ConfigurationResolver->getFixers() at /path/to/vendor/friendsofphp/php-cs-fixer/src/Console/Command/FixCommand.php:207
 PhpCsFixer\Console\Command\FixCommand->execute() at /path/to/vendor/symfony/console/Command/Command.php:252
 Symfony\Component\Console\Command\Command->run() at /path/to/vendor/symfony/console/Application.php:865
[...]

PR #471, opened 2018-02-25
Changed phpcsfixer2 default options

vyshkant:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #469

Existed non-null default values of "allow_risky" and "using_cache" options were overriding the values from .php_cs config.

The current behavior is that if these options are not specified, the values will be taken from the .php_cs config.

### DIFF
--- a/spec/Collection/ProcessArgumentsCollectionSpec.php
+++ b/spec/Collection/ProcessArgumentsCollectionSpec.php
@@ -122,4 +122,22 @@ class ProcessArgumentsCollectionSpec extends ObjectBehavior
 
         $this->getValues()->shouldBe(['--argument=file1.txt,file2.txt']);
     }
+
+    function it_should_be_able_to_add_boolean_nullable_argument_with_null_value()
+    {
+        $this->addOptionalBooleanArgument('--argument=%s', null, 'yes', 'no');
+        $this->getValues()->shouldBe([]);
+    }
+
+    function it_should_be_able_to_add_boolean_nullable_argument_with_true_value()
+    {
+        $this->addOptionalBooleanArgument('--argument=%s', true, 'yes', 'no');
+        $this->getValues()->shouldBe(['--argument=yes']);
+    }
+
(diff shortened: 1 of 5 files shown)


src/Task/PhpCsFixerV2.php
```diff
@@ -76,7 +76,11 @@ public function run(ContextInterface $context)
         $arguments = $this->processBuilder->createArgumentsForCommand('php-cs-fixer');
         $arguments->add('--format=json');
         $arguments->add('--dry-run');
-        $arguments->addOptionalArgument('--allow-risky=%s', $config['allow_risky'] ? 'yes' : 'no');
+
+        if (null !== $config['allow_risky']) {
```
veewee: We try to avoid these if statements inside the tasks. Can you move this nullable argument to a new `$arguments->add*` method?

vyshkant: > We try to avoid these if statements inside the tasks. Can you move this nullable argument to a new $arguments->add* method?

@veewee Done!===

spec/Collection/ProcessArgumentsCollectionSpec.php
```diff
@@ -122,4 +122,16 @@ function it_should_be_able_to_add_an_argument_with_comma_separated_files()
 
         $this->getValues()->shouldBe(['--argument=file1.txt,file2.txt']);
     }
+
+    function it_should_be_able_to_add_boolean_nullable_argument()
```
Landerstraeten: I would make three separate tests. Now you need to take the previous result into account.

vyshkant: @Landerstraeten sure! Don't go anywhere, I'll fix it right now :)

Landerstraeten:
@vyshkant thank you!

Issue #477, opened 2018-03-17
Add phpcpd "paths_exclude" option


vyshkant:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.14.0`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no

My suggestion is to add `phpcpd` `"paths_exclude"` option, which will correspond to the `--regexps-exclude` console option of `phpcpd` (which actually allows non-regex arguments).

This will allow user to exclude a specific file ("names_exclude" option operates only with file name, not with the path to the file).

PR #478, opened 2018-03-17
Add phpcpd "paths_exclude" option

vyshkant:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #477

Added the phpcpd "paths_exclude" option, which corresponds to the `--regexps-exclude` console option of phpcpd.

### DIFF
--- a/doc/tasks/phpcpd.md
+++ b/doc/tasks/phpcpd.md
@@ -20,6 +20,7 @@ parameters:
             directory: '.'
             exclude: ['vendor']
             names_exclude: []
+            paths_exclude: []
             fuzzy: false
             min_lines: 5
             min_tokens: 70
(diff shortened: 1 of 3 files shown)


jyggen:
Any particular reason why it's not simply called `regexps_exclude`? Having the options named the same in GrumPHP as in the tool itself seems more user friendly/less confusing to me.

vyshkant:
The only reason is that the path don't have to be a regexp. So it's also confusing to set raw string to the "regexps_exclude" field.

In addition, I've opened an issue on `phpcpd` repository in which I suggested renaming this option (https://github.com/sebastianbergmann/phpcpd/issues/159)

veewee:
Hi @vyshkant,

Thanks for the PR! Can you rebase against master?
As @jyggen mentions: It seems best to keep the naming of the tool (`regexps_exclude`). Can you change the parameter?

Once the issue in phpcpd is fixed, we can still rename the property here.

PR #479, opened 2018-03-21
Let symfony developers use phpunit-bridge.

DevMDamien:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

`symfony/phpunit-bridge` doesn't work with grumphp. With this feature, you can specify the bin file to use for the phpunit task.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,8 @@
         "phpunit/phpunit": "^5.7.27|^6.4.4",
         "sebastian/comparator": "^1.2.4",
         "sensiolabs/security-checker": "^4.0",
-        "squizlabs/php_codesniffer": "~2.4"
+        "squizlabs/php_codesniffer": "~2.4",
+        "symfony/phpunit-bridge": "~2.7|~3.0|~4.0"
     },
     "suggest": {
         "atoum/atoum": "Lets GrumPHP run your unit tests.",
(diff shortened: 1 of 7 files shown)


Landerstraeten:
Can you also update the spec tests?

veewee:
Hi @DevMDamien,

Thanks for the PR! I haven't used that tool myself so I am not sure it works exactly the same as the regular phpunit. Maybe it makes more sense to create a separate task for the simple-phpunit tool?

You can also add the phpunit bridge package to the suggest section of composer.json.

veewee:
@DevMDamien : I think it is a safer option to move it to a separate task. I can't tell if the cli tool accepts and will always accept the same arguments as phpunit. Feel free to create a separate task.

composer.json
```diff
@@ -30,7 +30,8 @@
         "phpunit/phpunit": "^5.7.27|^6.4.4",
         "sebastian/comparator": "^1.2.4",
         "sensiolabs/security-checker": "^4.0",
-        "squizlabs/php_codesniffer": "~2.4"
+        "squizlabs/php_codesniffer": "~2.4",
+        "symfony/phpunit-bridge": "~2.7|~3.0|~4.0"
```
veewee: We do not require this dev dependency

doc/tasks/phpunitbridge.md
```diff
@@ -0,0 +1,56 @@
+# Phpunit
```
veewee: Phpunit bridge

veewee: A link to this file should also be added to `docs/tasks.yml`, the key should also be added to the global readme file.

src/Task/PhpunitBridge.php
```diff
@@ -0,0 +1,81 @@
...
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * PhpunitBridge task
+ *
+ * @link https://symfony.com/doc/current/components/phpunit_bridge.html
+ */
+class PhpunitBridge extends AbstractExternalTask
```
veewee: This task is not registered in the service config file located at `resources/config/tasks.yml`

hmonglee:
@veewee Hello, is this PR OK for merge ? We need it. Thank you

Landerstraeten:
Closed for #558

PR #484, opened 2018-03-30
Ignore symlinks filesize task

Landerstraeten:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | https://github.com/phpro/grumphp/pull/450

### DIFF
--- a/src/Collection/FilesCollection.php
+++ b/src/Collection/FilesCollection.php
@@ -245,4 +245,14 @@ class FilesCollection extends ArrayCollection
 
         return $newFiles;
     }
+
+    /**
+     * @return FilesCollection
+     */
+    public function ignoreSymlinks()
+    {
+        return $this->filter(function (SplFileInfo $file) {
+            return !$file->isLink();
+        });
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Collection/FilesCollection.php
```diff
@@ -245,4 +245,19 @@ public function ensureFiles(FilesCollection $files)
 
         return $newFiles;
     }
+
+    /**
+     * @return FilesCollection
+     */
+    public function ignoreSymlinks()
```
veewee: Please use the `FilesCollection->filter()` method internally. These methods should be immutable instead of manipulating the internal array.

src/Task/FileSize.php
```diff
@@ -76,12 +76,14 @@ public function run(ContextInterface $context)
...
-        if (0 === count($context->getFiles())) {
+        if (0 === $context->getFiles()->count()) {
             return TaskResult::createSkipped($this, $context);
         }
 
         $maxSize = $config['max_size'];
-        $files = $context->getFiles()->size(sprintf('>%s', $maxSize));
+        $files = $context->getFiles()
+            ->size(sprintf('>%s', $maxSize))
+            ->ignoreSymlinks();
```
veewee: Beter to run the ignoreSymlinks() method before checking the size. That was the actual problem right?

veewee:
Thanks!

PR #497, opened 2018-04-16
Add names_exclude to GitBranchName

mgeoffray:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | no

This PR add a new `names_exclude`option to `GitBranchName` task. It will be used to filter branch name to be validated.

``` yml
git_branch_name:
    names_exclude: 
        - master
        - feature
```

### DIFF
--- a/doc/tasks/git_branch_name.md
+++ b/doc/tasks/git_branch_name.md
@@ -11,6 +11,7 @@ parameters:
             matchers:
                 Branch name must contain JIRA issue number: /JIRA-\d+/
             additional_modifiers: ''
+            names_exclude: []
 ```
 
 **matchers**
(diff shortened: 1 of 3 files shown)


veewee:
Hi @mgeoffray,

Thanks for the PR. We've discussed this internally and think that this will cause some confusion.
Suggested way of altering the task:

- Add a blacklist option: if a branch regex is in the blacklist: fail
- Add a whitelist option: If a branch regex is in the whitelist: pass
- If the branch is in non of the lists: Fail / pass depending on a config option
- (Blacklist will be checked before whitelist)

mgeoffray:
Hi @veewee ,
Yes that's right I agree on that.   
I will make these changes as soon as I can in this PR.  
Thanks!

PR #502, opened 2018-04-18
Added -c argument to run command properly

Mavlyan:
Added -c option  in order to run command not a file.
---
Quote from sh man: 
The -c option causes the commands to be read from the string operand instead of from the standard input.

### DIFF
--- a/doc/tasks/shell.md
+++ b/doc/tasks/shell.md
@@ -18,7 +18,7 @@ parameters:
 
 This options specifies the paths to your shell scripts.
 You can specify one or more scripts. 
-You also can specify one or more shell commands.
+You also can specify one or more shell commands. Add `-c` as a first argument. The -c option causes the commands to be read from the string.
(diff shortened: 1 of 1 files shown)


doc/tasks/shell.md
```diff
@@ -30,7 +30,7 @@ parameters:
         shell:
             scripts:
                - script.sh
-               - ["./bin/command", "arg1", "arg2"]
+               - ["-c", "./bin/command", "arg1", "arg2"]
```
keradus: i would still leave non `-c` call

question: `["-c", "foo", "arg1"]` or `["-c", "foo arg1"]` ?

veewee: Valid argument. The `-c`option takes the next argument as input string.
You'll need to configure it as 2 parameters
Example: `sh -c "ls -al"`

keradus:
wow ;)

Issue #482, opened 2018-03-30
Gradually upgrade to PHP 7.0


veewee:
Follow up issue of #434.

These steps should be done in separate - easy to review - PRs.

- [ ] Inject strict parameters in functions
- [ ] Inject strict return types in functions
- [ ] Remove unnecesary docblocks
- [ ] Declare strict types
- [ ] Upgrade configuration

Landerstraeten:
7.0 or 7.1 ?

veewee:
Lets start with adopting the new 7.0 options.  ;)

veewee:
@Landerstraeten Any updates on this? Does the grumpy-seventies branch works for you?

richardhughes:
Hey @veewee, How is this progressing? If you need a hand, I'll be happy to help

PR #504, opened 2018-04-19
Update docs to make it clear that PHP 7 isn't supported yet.

nicrodgers:


jyggen:
What does not work exactly? I've been running GrumPHP on PHP 7 and above for ages.

nicrodgers:
See https://github.com/phpro/grumphp/issues/482

jyggen:
#482 is about dropping 5.6 support and using PHP 7.0 features in the codebase, but running GrumPHP on PHP 7.0 is already supported (as seen on [Travis](https://travis-ci.org/phpro/grumphp)). So saying "but does not currently support PHP 7" isn't true.

nicrodgers:
Ah, right - my bad then. We recently switched a project that was using grumphp from php 5.6 to php 7, and we were unable to git commit since switching, because grumphp was failing. I checked the issue queue and saw #482, and took that to mean PHP 7 isn't yet supported.

So, i'll close this PR and investigate the bug further, then create an issue for that as necessary.

PR #507, opened 2018-05-13
Add Progpilot task

eric-therond:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

Add Progpilot task

# New Task Checklist:

### DIFF
--- a/doc/tasks/progpilot.md
+++ b/doc/tasks/progpilot.md
@@ -1,0 +1,40 @@
+# Progpilot
+
+The Progpilot task will run your automated PHP tasks.
+
+***Composer***
+
+```
+composer config minimum-stability dev
+composer require --dev designsecurity/progpilot:dev-master
+```
+
+***Config***
+
+The task lives under the `progpilot` namespace and has following configurable parameters.
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        progpilot:
+            config_file: .progpilot/configuration.yml
+            triggered_by: [php]
+```
+
+**config_file**
+
+*Default: configuration.yml*
+
+You can specify your custom configuration file location with this option.
+By default `.progpilot/configuration.yml` is automatically loaded if the file exists in the current directory.
+If not or yml file format is invalid default configuration of progpilot will be used.
+
+
+**triggered_by**
+
+*Default: [php]*
+
+This option will specify which file extensions will trigger the Progpilot task.
+By default Progpilot will be triggered by altering a PHP file.
+ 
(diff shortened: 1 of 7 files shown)


doc/tasks/progpilot.md
```diff
@@ -0,0 +1,40 @@
+# Progpilot
+
+The Progpilot task will run your automated PHP tasks.
+
+***Composer***
+
+```
+composer config minimum-stability dev
```
veewee: Shouldn't we propose a released version instead?

src/Task/Progpilot.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @return OptionsResolver
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(
+            [
+                'config_file'  => '.progpilot/configuration.yml',
```
veewee: Isn't it better to set this value to null so that the defaults from the cli tool are being used?

src/Task/Progpilot.php
```diff
@@ -0,0 +1,78 @@
...
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $config = $this->getConfiguration();
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('progpilot');
+
+        $arguments->addOptionalArgumentWithSeparatedValue('--configuration', $config['config_file']);
+        $arguments->addFiles($files);
```
veewee: Maybe better not to specify files during run context?

src/Task/Progpilot.php
eric-therond: Defaults values from the tool are also used when the configuration file specified in the cli doesn't exist so if someone doesn't have a file *.progpilot/configuration.yml* the defaults values from the tool will be used and if someone has a file *.progpilot/configuration.yml* but has forgotten to define it in the parameters of the task the values of *.progpilot/configuration.yml* will be used.

src/Task/Progpilot.php
eric-therond: If I comment the line 67 the commited files are not analyzed by the tool

doc/tasks/progpilot.md
eric-therond: Unfortunately, progpilot relies on ircmaxell/php-cfg which doesn't currently have a stable version

veewee:
Sounds reasonable. Thanks for the feedback!

PR #510, opened 2018-05-25
Add psalm

Vesquen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

It adds a task for Psalm

# New Task Checklist:

### DIFF
--- a/doc/tasks/psalm.md
+++ b/doc/tasks/psalm.md
@@ -46,6 +46,12 @@ With this parameter you can run Psalm without using the cache file.
 
 With this path you can specify the path your psalm report file 
 
+**threads**
+
+*Default: null*
+
+This parameter defines on how many threads Psalm's analysis stage is ran.
+
 
 **triggered_by**
 
--- a/doc/tasks/psalm.md
+++ b/doc/tasks/psalm.md
@@ -22,17 +22,22 @@ parameters:
             triggered_by: ['php']
 ```
 
+
 **config**
 
 *Default: ~*
 
 With this parameter you can specify the path your project's configuration file.
 
+
 **ignore_patterns**
 
 *Default: []*
 
-This is a list of patterns that will be ignored by psalm. With this option you can skip files like tests. Leave this option blank to run psalm for every php file/directory specified in your configuration.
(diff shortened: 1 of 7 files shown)


src/Task/Psalm.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function getConfigurableOptions()
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'config' => null,
+            'ignore_patterns' => [],
+            'no_cache' => false,
+            'report' => null,
+            'threads' => 1,
```
veewee: Maybe it is better to set this to `null` and let psalm decide what it uses by default?

doc/tasks/psalm.md
```diff
@@ -0,0 +1,54 @@
...
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        psalm:
+            config: psalm.xml
+            ignore_patterns: []
+            no_cache: false
+            report: ~ 
+            threads: 1
```
veewee: The threads options is not described in the documentation

src/Task/Psalm.php
```diff
@@ -0,0 +1,92 @@
...
+        if (0 === count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand(self::TASK_NAME);
+        $arguments->addOptionalArgument('--config=%s', $config['config']);
+        $arguments->addOptionalArgument('--report=%s', $config['report']);
+        $arguments->addOptionalArgument('--no-cache', $config['no_cache']);
+        $arguments->addOptionalArgument('--threads=%d', $config['threads']);
+        $arguments->addFiles($files);
```
veewee: During `run`, the file list might get very big. Would it make sense to only add the files during pre-commit context and run on all files during run context?

Vesquen: That does make sense, will fix!

janvernieuwe:
Thanks for adding the docs, we'll merge it in with the next milestone release!

PR #517, opened 2018-06-26
Ignore commented lines

exussum12:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | Not sure ? Its expected as its not saved

By default git commit gives some commented lines of what is going to happen. Remove any commented line from the check

### DIFF
--- a/src/Task/Git/CommitMessage.php
+++ b/src/Task/Git/CommitMessage.php
@@ -173,6 +173,9 @@ class CommitMessage implements TaskInterface
         }
 
         foreach (array_slice($lines, 2) as $index => $line) {
+            if (isset($line[0]) && $line[0] == '#') {
+                continue;   
+            }
             if (mb_strlen(rtrim($line)) > $config['max_body_width']) {
                 $errors[] = sprintf(
                     'Line %u of commit message has > %u characters.',


janvernieuwe:
Hi @exussum12 !
Thank you for your PR

A few small remarks:
Can you use $line !== '' instead of isset and use === check for the hash symbol compare?
Also please add a test that verifies this functionality.

src/Task/Git/CommitMessage.php
```diff
@@ -173,7 +173,7 @@ private function enforceTextWidth(ContextInterface $context)
         }
 
         foreach (array_slice($lines, 2) as $index => $line) {
-            if (isset($line[0]) && $line[0] == '#') {
+            if ($line !== '' && $line[0] === '#') {
```
SerkanYildiz: IMO the first check isn't needed because the second one decides to continue or not to continue with foreach

exussum12: If the line is empty, [0] does not exist and will throw a warning. It would function the same, the first check was to avoid a warning

exussum12:
Updated and changed a test

veewee:
@exussum12, Thanks for the PR!
The method `getCommitMessageLinesWithoutComments()` which is being used for providing the `$lines` should strip all comments. Since that method is being used in other checks as well, it is better to fix it in that method instead.

Issue #523, opened 2018-07-09
PHPCS task does not output errors


ctrl-f5:
| Q               | A
| --------------- | ---
| Version         | 0.14.1
| Question?       | yes
| Related issues | #487, #275

when running the phpcs task, I'm simply getting an X, but no actual errors are shown, is this intended behavior?
```
$ include/vendor/bin/grumphp run
GrumPHP is sniffing your code!
Running task 1/1: Phpcs... ✘

                                    _
                                   | |
 _   _  ___  _   _   ___ _   _  ___| | __
| | | |/ _ \| | | | / __| | | |/ __| |/ /
| |_| | (_) | |_| | \__ \ |_| | (__|   <
 \__, |\___/ \__,_| |___/\__,_|\___|_|\_\
  __/ |
 |___/
```

veewee:
Hi @ctrl-f5,

It should output the error message. Can you run the same command by adding the options `-vvv` to the cli. That way you can debug what command is executed for phpcs and why it does not output error messages.

veewee:
Hi @ctrl-f5,

Internally, it uses a formatter to parse the json to a text format:
https://github.com/phpro/grumphp/blob/master/src/Formatter/PhpcsFormatter.php

The format might have changed in a newer version of phpcs. Can you share me the version of phpcs?

ctrl-f5:
```
$ composer info squizlabs/php_codesniffer
name     : squizlabs/php_codesniffer
descrip. : PHP_CodeSniffer tokenizes PHP, JavaScript and CSS files and detects violations of a defined set of coding standards.
keywords : phpcs, standards
versions : * 3.2.3

```

ctrl-f5:
![ouput](https://user-images.githubusercontent.com/485346/42499574-6c6d3152-842f-11e8-9c6f-4ee0eeb5de0a.png)

It isn't getting any output from the process, I'll debug it further

ctrl-f5:
seems to be due to the sheer number of files or size of the output, when I restrict phpcs to just the config folder, it works as expected:

```
    phpcs:
      standard: "./dev/phpcs/phpcs.xml"
      whitelist_patterns:
        - /^config\/(.*)/
```

When doing a run of the complete codebase, it's appending more than 10K files to the command (command length is 650K chars)

veewee:
Hi @ctrl-f5,

It seems like you are hitting the maximum CLI input length. 
[...]

PR #524, opened 2018-07-10
allow tasks with large number of parameters to write command to file

ctrl-f5:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #523

When running grump on an entire codebase instead of only on commit files, phpcs and phplint tasks receive a huge string of files, making the command too long for CLI.

To bypass this, write the command to a file and execute this.

Current implementation will only work on *nix, since it executes the file with `sh`

### DIFF
--- a/src/Task/Phpcs.php
+++ b/src/Task/Phpcs.php
@@ -41,7 +41,8 @@ class Phpcs extends AbstractExternalTask
             'severity' => null,
             'error_severity' => null,
             'warning_severity' => null,
-            'triggered_by' => ['php']
+            'triggered_by' => ['php'],
+            'use_cmd_tmp_file' => false
         ]);
 
         $resolver->addAllowedTypes('standard', ['null', 'string']);
@@ -54,6 +55,7 @@ class Phpcs extends AbstractExternalTask
         $resolver->addAllowedTypes('error_severity', ['null', 'int']);
         $resolver->addAllowedTypes('warning_severity', ['null', 'int']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
+        $resolver->addAllowedTypes('use_cmd_tmp_file', ['bool']);
 
         return $resolver;
     }
@@ -96,6 +98,10 @@ class Phpcs extends AbstractExternalTask
         $arguments->addFiles($files);
 
         $process = $this->processBuilder->buildProcess($arguments);
+        if ($config['use_cmd_tmp_file']) {
+            $process = $this->processBuilder->proxyThroughTmpFile($process);
+        }
+        $process->run();
         $process->run();
 
         if (!$process->isSuccessful()) {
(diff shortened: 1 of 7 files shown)


src/Task/Phpcs.php
```diff
@@ -96,6 +98,10 @@ public function run(ContextInterface $context)
         $arguments->addFiles($files);
 
         $process = $this->processBuilder->buildProcess($arguments);
+        if ($config['use_cmd_tmp_file']) {
+            $process = $this->processBuilder->proxyThroughTmpFile($process);
+        }
+        $process->run();
         $process->run();
```
pfrenssen: Probably not necessary to run it twice here.

veewee:
Hi @ctrl-f5,

Thanks for the PR.
Since this problem occurs on multiple tasks, I would prefer to merge in a more generic approach. Something like:

- Configure a flag in the grumphp.yml file to enable the suggested tmp file execution, optionally with a max-cli-chars amount. (We can try to detect the default setting from the OS)
- When  the input is too long and the flag is configured: run the process through a tmp file
- When the input is short enough: run the process directly.

Would that be something you'dd be interested in adding to this project?

gsomoza:
@veewee I don't think that a generic approach is going to work, because each external process will receive large input in different ways. For example:

1. `phpcs` does support a `--file-list` argument that you can use to provide the list via a temporary file.
2.  but while `phplint` doesn't have that type of option, it can receive files via STDIN – the `--stdin` argument has to be use for that to work.

veewee:
Hi @gsomoza,

The code in this PR writes the complete CLI command to a tmp file and executes it by running `sh /tmp/file`. It does  not need to know how the tool works internally.

gsomoza:
Right, but that's also the reason why it only works in UNIX, right? 

Wouldn't we have a cross-platform solution if we update each task to receive input using the mass-input options that each external programs already supports? You could then accept PRs for that in a task-by-task basis.

The one I'd like to have this feature for is PHPLint: if the task *always* uses STDIN to pass the file list to the program, then you wouldn't even need to add an option to `grumphp.yaml` because it would work even for short file lists.

It's totally possible I'm missing on lots of details though, I'm not familiar at all with the mechanics around GrumPHP. Sorry if that's the case :) just my two cents

gsomoza:
> I don't think there is a general solution for the problem.

> Passing files through stdin is only possible if the tools support this + it might require additional input arguments.

The bottom line is that each tool implements input in a different way. There's no easy, cross-platform way of standardising that I know of, so I think the only reasonable solution is to implement mass-input options specifically for each plugin. If by doing that you find a pattern, it would always be possible to extract that pattern out into a more generic solution that only applies to the tools that support it – but I think the specific implementations come first.

Landerstraeten:
Any update on this one?

veewee:
Hi @ctrl-f5,

I've tested this and doesn't seem to work the way we expect it to work.
More information can be found in this repo:
https://github.com/veewee/long-cli-command/

Closing this one for now. Feel free to contact me if you have additional questions.

Issue #508, opened 2018-05-19
Add PHP-Parser 4 support


vyshkant:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.14.0`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no

I have one case of incompatibility with PhpParser 4, but there could be more.

The type of `\PhpParser\Node\Stmt\DeclareDeclare::$key` has been changed since `4.0.0` version of `PHP-Parser`.

We expect this property to be string:
https://github.com/phpro/grumphp/blob/60ed6970d432c889aba6316bfd223b22b139cb5b/src/Parser/Php/Visitor/DeclareStrictTypesVisitor.php#L26

but in `^4.0` version it is `PhpParser\Node\Identifier`: https://github.com/nikic/PHP-Parser/blob/9c18e3db49fa469f5feffe40dbd7b1ec2b61e41f/lib/PhpParser/Node/Stmt/DeclareDeclare.php#L9-L10

So strict comparison will always return `false`, which leads to unexpected behavior.

A quick-fix of my **single case** is to cast `$declare->key` to string (`PhpParser\Node\Identifier::__toString()` will return expected value):
```php
if ((string)$declare->key !== 'strict_types') { 
```

vyshkant:
Moreover: `symfony/maker-bundle v1.5.0 requires nikic/php-parser ^4.0`.

`symfony/maker-bundle` is provided by default with Symfony 4.0.

Landerstraeten:
Fixed

vyshkant:
@Landerstraeten can you please release this fix?

PR #525, opened 2018-07-11
Make detection of strict types compatible with PhpParser 4.

pfrenssen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #508 

As described in #508 the visitor that checks the `declare(strict_types = 1);` directive currently only works with PhpParser 3 and below, and always fails on PhpParser 4.

This PR makes sure it is compatible with PhpParser 4 and above.


netlooker:
+1 - small and neat way of solving this annoying issue.

Issue #530, opened 2018-08-15
Allow array of failed / succeeded ascii art; choose one randomly


q0rban:
| Q               | A
| --------------- | ---
| Version         | `0.11.6`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | comma-separated list of related tickets

It'd be fun to have an array of different ascii txt files, and have GrumPHP choose one randomly. For example, I'd like to have a random Mister Rogers quote:

```
parameters:
  ascii:
    failed: 
      - what-do-you-do-with-the-mad-that-you-feel.txt
      - anything-mentionable-is-manageable.txt
      - go-ahead-and-cry-im-here-to-be-with-you.txt
    succeeded:
     - you-are-special.txt
     - you-are-my-friend.txt
     - its-a-beautiful-day-in-the-neighborhood.txt
```

veewee:
Hi @q0rban 

It is simple to add, I'd accept a PR that adds this. Be sure to add tests and updated docs.

q0rban:
@veewee PR in #538! 😀

PR #538, opened 2018-08-31
Choose a random ascii if list is provided #530

q0rban:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #530 

This pull requests allows users to provide a list of ascii files in their grumphp.yml. GrumPHP will choose one at random for each run.

### DIFF
--- a/src/Configuration/GrumPHP.php
+++ b/src/Configuration/GrumPHP.php
@@ -174,7 +174,7 @@ class GrumPHP
         // Deal with multiple ascii files by returning one at random.
         if (is_array($paths[$resource])) {
             shuffle($paths[$resource]);
-            $paths[$resource] = reset($paths[$resource]);
+            return reset($paths[$resource]);
         }
 
         return $paths[$resource];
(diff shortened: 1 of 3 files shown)


q0rban:
This is my first time with PHPSpec so let me know if I did the test correctly.

src/Configuration/GrumPHP.php
```diff
@@ -171,6 +171,12 @@ public function getAsciiContentPath($resource)
             return null;
         }
 
+        // Deal with multiple ascii files by returning one at random.
+        if (is_array($paths[$resource])) {
+            shuffle($paths[$resource]);
+            $paths[$resource] = reset($paths[$resource]);
```
janvernieuwe: You can just do a return of the reset here right away.

q0rban: I suppose there's already a precedent for an early return in this function, so 👍 .

PR #539, opened 2018-09-13
feat: Add additional info parameter

poppabear8883:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | none

Adds the ability to set additional information to be displayed after tasks have finished. Is useful for displaying resources ie: documentation urls

**additional_info**

*Default: null*

This parameter will display additional information at the end of a `success` *or* `error` task.

```yaml
# grumphp.yml
parameters:
  additional_info: "\nTo get full documentation for the project\nVisit https://docs.example.com\n"
```

*Example Result:*
```
Running task 1/1: Phpcs... ✔

To get full documentation for the project
Visit https://docs.example.com

```

### DIFF
--- a/src/Console/Helper/TaskRunnerHelper.php
+++ b/src/Console/Helper/TaskRunnerHelper.php
@@ -168,7 +168,9 @@ class TaskRunnerHelper extends Helper
      */
     private function returnAdditionalInfo(OutputInterface $output)
     {
-        is_null($this->config->getAdditionalInfo()) ?: $output->writeln($this->config->getAdditionalInfo());
+        if (null !== $this->config->getAdditionalInfo()) {
+            $output->writeln($this->config->getAdditionalInfo());
+        }
     }
 
     /**
--- a/src/Console/Helper/TaskRunnerHelper.php
+++ b/src/Console/Helper/TaskRunnerHelper.php
@@ -126,6 +126,8 @@ class TaskRunnerHelper extends Helper
             );
         }
 
+        $this->returnAdditionalInfo($output);
+
         return self::CODE_ERROR;
     }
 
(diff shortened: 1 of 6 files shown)


janvernieuwe:
Hi @poppabear8883,

Can't this be achieved with event subscriber(s) https://github.com/phpro/grumphp/blob/master/doc/events.md?

poppabear8883:
@janvernieuwe yes it appears so ... with more effort on the users side. 

I'm sure a lot of the integrated "features/parameters" could be achieved through the event subscribers like the `ascii` option for example. Its fairly convenient for us however to just add the parameter with the required value(s).

I think its worth a merge for the simplicity and under the notion that more users will take advantage of this feature now that they don't have to create a class to display a string ...

With that said, its only a proposed feature, and if you/they decide not to merge it, I appreciate the consideration nonetheless.

src/Console/Helper/TaskRunnerHelper.php
```diff
@@ -160,6 +163,14 @@ private function returnWarningMessages($output, array $warningMessages)
         }
     }
 
+    /**
+     * @param OutputInterface $output
+     */
+    private function returnAdditionalInfo(OutputInterface $output)
+    {
+        is_null($this->config->getAdditionalInfo()) ?: $output->writeln($this->config->getAdditionalInfo());
```
veewee: This syntax is a bit confusing. Could you just use an if instead?

poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/539/commits/d31d927dafc46cf194ac5f12cf81af4a39605b9a

veewee:
Thanks!

PR #541, opened 2018-09-13
Commit Message Features and Add new Util class

poppabear8883:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | None

This PR addresses a few things:

## New String Helper Util Class
**Str::contains($haystack, $needles)**
```php
// Can pass an `array` of needles or a single `string`
Str::contains($subject, ['.', '!', '?', ','])
Str::contains($subject, '!')
```

## The ability to `enforce no subject punctuation`

**enforce_no_subject_punctuations**

*Default: false*

Ensures that the commit message subject line doesn't have any punctuations characters.

## The ability to `Enforce Type Scope Conventions`
I'm sure you know of this convention, but just in-case you don't ...

**enforce_type_scope_conventions**

*Default: false*

Enable a commonly used convention for the subject line.

Format is as follows:
```
type[(scope)]: subject
```
*The scope is optional*

Good examples:
```
fix: Security issue with password hashing
fix(Password): Security issue with password hashing  
```

**types**

*Default: []*

*To be used when `enforce_type_scope_conventions: true`*

Specify a list of acceptable types. Default allows ***all*** types.

Add one or multiple types like:
```yaml
types:
    - build
    - ci
    - chore
    - docs
    - feat
    - fix
    - perf
    - refactor
    - revert
    - style
    - test
```

**scopes**

*Default: []*

*To be used when `enforce_type_scope_conventions: true`*

Specify a list of acceptable scopes. Default allows ***all*** scopes.

Add one or multiple scopes like:
```yaml
scopes:
    - api
    - index
    - user
    - language
    - browser
    - environment
```

### DIFF
--- a/doc/tasks/git_commit_message.md
+++ b/doc/tasks/git_commit_message.md
@@ -10,8 +10,12 @@ parameters:
         git_commit_message:
             allow_empty_message: false
             enforce_capitalized_subject: true
+            enforce_no_subject_punctuations: false
             enforce_no_subject_trailing_period: true
             enforce_single_lined_subject: true
+            enforce_type_scope_conventions: false
+            types: []
+            scopes: []
             max_body_width: 72
             max_subject_width: 60
             matchers:
@@ -32,6 +36,12 @@ Controls whether or not empty commit messages are allowed.
 *Default: true*
 
 Ensures that the commit message subject line starts with a capital letter.
+
+**enforce_no_subject_punctuations**
+
+*Default: false*
+
+Ensures that the commit message subject line doesn't have any punctuations characters.
 
 **enforce_no_subject_trailing_period**
 
@@ -95,3 +105,65 @@ additional_modifiers: 'u'
 
 additional_modifiers: 'xu'
 ```
+
+**enforce_type_scope_conventions**
+
+*Default: false*
+
+Enable a commonly used convention for the subject line.
+
+Format is as follows:
+```
+type[(scope)]: subject
+```
+*The scope is optional*
+
+Good examples:
+```
+fix: Security issue with password hashing
+fix(Password): Security issue with password hashing  
+```
+
+**types**
+
+*Default: []*
+
+*To be used when `enforce_type_scope_conventions: true`*
+
+Specify a list of acceptable types. Default allows ***all*** types.
+
+Add one or multiple types like:
+```yaml
+types:
+    - build
+    - ci
+    - chore
+    - docs
+    - feat
+    - fix
+    - perf
+    - refactor
+    - revert
+    - style
+    - test
+```
+
+**scopes**
+
+*Default: []*
+
+*To be used when `enforce_type_scope_conventions: true`*
+
+Specify a list of acceptable scopes. Default allows ***all*** scopes.
+
+Add one or multiple scopes like:
+```yaml
+scopes:
+    - api
+    - index
+    - user
+    - language
+    - browser
+    - environment
+```
+
--- a/spec/Task/Git/CommitMessageSpec.php
+++ b/spec/Task/Git/CommitMessageSpec.php
@@ -42,6 +42,10 @@ class CommitMessageSpec extends ObjectBehavior
         $options->getDefinedOptions()->shouldContain('multiline');
         $options->getDefinedOptions()->shouldContain('matchers');
         $options->getDefinedOptions()->shouldContain('additional_modifiers');
+        $options->getDefinedOptions()->shouldContain('enforce_no_subject_punctuations');
+        $options->getDefinedOptions()->shouldContain('enforce_type_scope_conventions');
+        $options->getDefinedOptions()->shouldContain('types');
+        $options->getDefinedOptions()->shouldContain('scopes');
     }
 
     function it_is_initializable()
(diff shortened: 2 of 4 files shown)


doc/tasks/git_commit_message.md
```diff
@@ -10,8 +10,12 @@ parameters:
         git_commit_message:
             allow_empty_message: false
             enforce_capitalized_subject: true
+            enforce_no_subject_punctuations: false
             enforce_no_subject_trailing_period: true
             enforce_single_lined_subject: true
+            enforce_type_scope_conventions: false
+            types: []
```
janvernieuwe: I think it would be better to nest the options that are only used by `enforce_type_scope_conventions` to be nested under it. So that it is clear that they can only be used in conjunction with the `enforce_type_scope_conventions` option.

poppabear8883: I thought about this as well, i will fix this up and will push it up. Stand by ...

poppabear8883:
@janvernieuwe 

I have the new proposal prepared. I did force-push to keep the commits clean. Let me know your thoughts. 

Thanks in advance

src/Task/Git/CommitMessage.php
```diff
@@ -21,6 +22,11 @@ class CommitMessage implements TaskInterface
      */
     private $grumPHP;
 
+    /**
+     * @var array $exceptions
+     */
+    private $exceptions = [];
```
veewee: Could you avoid the usage of this private property for temporary state inside the run context?

doc/tasks/git_commit_message.md
veewee: Invalid indentation

src/Task/Git/CommitMessage.php
```diff
@@ -235,6 +256,27 @@ private function getSpecialPrefixLength($string)
...
     }
 
+    /**
+     * @param ContextInterface $context
+     *
+     * @return bool
+     */
+    private function subjectHasPunctuations(ContextInterface $context)
+    {
+        $commitMessage = $context->getCommitMessage();
```
veewee: Probably best to use the `getCommitMessageLinesWithoutComments()` method to fetch the subject. It fixes some issues when the commit message contains comments. Maybe even better to add a private method `getSubjectLine()`, since that functionality is being duplicated in the task.

src/Util/Str.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace GrumPHP\Util;
+
+class Str
+{
+    public static function contains($haystack, $needles)
```
veewee: Could you rename this method to: containsOneOf()?
Can you add strict parameter types + return type.
I prefer to enforce the needles to be an `array` of strings - which makes more sense given the new name.

spec/Task/Git/CommitMessageSpec.php
```diff
@@ -606,4 +608,201 @@ function it_should_pass_when_subject_does_not_contain_a_trailing_period(GrumPHP
         $result->shouldBeAnInstanceOf(TaskResultInterface::class);
         $result->isPassed()->shouldBe(true);
     }
+
+    function it_should_pass_when_enforce_type_scope_conventions_is_false(
```
veewee: Thanks for adding all those tests. Really awesome! :)

doc/tasks/git_commit_message.md
poppabear8883: I'm a bit confused here, These should had be removed after the force push. I need to investigate this.

poppabear8883: Nevermind, it was just overlooked when force pushing the new proposal. A fix will be pushed soon.

src/Task/Git/CommitMessage.php
veewee: No, I'dd rather keep it as a value object to feed the tasks. Allowing the context to be changed in between the tasks might cause a lot of issues.
I would avoid the local state since the run method can theoretically be called multiple times which causes an issue and we don't really need this state internally.
A try/catch over the if statement will solve the issue you are addressing as well.

src/Task/Git/CommitMessage.php
poppabear8883: I was simply trying to be consistent with the way the rest of the code was written ...

```php
private function enforceTextWidth(ContextInterface $context)
    {
        $commitMessage = $context->getCommitMessage();
        $config = $this->getConfiguration();

        if (trim($commitMessage) === '') {
            return TaskResult::createPassed($this, $context);
        }

        $errors = [];
        $lines = $this->getCommitMessageLinesWithoutComments($commitMessage);
...
```
Above is only `1 of many` that uses the exact method I used here ?

***Or***

Are you proposing that this be refactored due to the observation of what I am stating above ? I can certainly refactor the whole task to make use of a new private method.

Please advise
Thanks

veewee: I am sorry, the comment should be one line below on how the subject is subtracted from the message.
The first line can e.g. contain a comment. Since that logic for fetching the subject is repeated, that might go in a new private method. Feel free to refactor where it makes sense.

src/Task/Git/CommitMessage.php
poppabear8883: So if i am understanding correctly, that the task could run multiple times using a single instance which ***could*** persist the state of the `$exceptions` which would yield undesired behavior?

If this is accurate, couldn't we just reset exceptions back to an empty array in the `run()` method ?

doc/tasks/git_commit_message.md
poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/541/commits/aa6d1ee5ab70d8408b731d2b645f76291f6687ce

src/Task/Git/CommitMessage.php
poppabear8883: These should resolve this conversation
https://github.com/phpro/grumphp/pull/541/commits/3217efd75d9a477b8f820553add23d6ed8f3a45d
https://github.com/phpro/grumphp/pull/541/commits/b3182857cabc6ef3d075d5c41dbdaf75a05c0483

src/Util/Str.php
poppabear8883: One concern with the method name .. Not only can needles be a array of strings, but it can also be just a single string ...

So for me it would read weird in this use case:
```php
Str::containsOneOf('This string contains this word', 'word');
```

poppabear8883: I re-read what you typed and i think you are suggesting to `only` allow needles to be an `array` of strings which then would accommodate the new method name ...

I will make these changes and push shortly.

poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/541/commits/2de742761c1857f9e3a8a79363f215b75bfb428b

src/Task/Git/CommitMessage.php
veewee: That is possible, but I prefer not to have it at all :)

poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/541/commits/0056460816fd622c9986c5eb1dd8de82e29a467c

src/Task/Git/CommitMessage.php
```diff
@@ -240,17 +262,35 @@ private function getSpecialPrefixLength($string)
...
-        $commitMessage = $context->getCommitMessage();
+        $subjectLine = $this->getSubjectLine($context);
 
-        if (trim($commitMessage) === '') {
+        if (trim($subjectLine) === '') {
             return false;
         }
 
-        $lines = $this->getCommitMessageLinesWithoutComments($commitMessage);
+        if (Str::containsOneOf($subjectLine, ['.', '!', '?', ','])) {
```
veewee: `return Str::containsOneOf($subjectLine, ['.', '!', '?', ',']);`

src/Task/Git/CommitMessage.php
```diff
@@ -332,4 +372,77 @@ private function getCommitMessageLinesWithoutComments($commitMessage)
...
+
+    private function enforceTypeScopeConventions()
+    {
+        $config = $this->getConfiguration();
+
+        if (!is_array($config['type_scope_conventions'])) {
+            return false;
+        }
+
+        if (!in_array('types', array_keys($config['type_scope_conventions']))
```
veewee: A bit shorter:
```php
$conventionsKeys = array_keys($config['type_scope_conventions']);

return in_array('types', $conventionsKeys) || in_array('scopes', $conventionsKeys);
```

src/Task/Git/CommitMessage.php
poppabear8883: Good catch!

src/Task/Git/CommitMessage.php
poppabear8883: Agreed, this looks much better thank you ...

src/Task/Git/CommitMessage.php
poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/541/commits/8ee9e9895f1886a6397d44cdc7902fb8d1d8f00c

src/Task/Git/CommitMessage.php
poppabear8883: This should resolve this conversation https://github.com/phpro/grumphp/pull/541/commits/e2e2393936e52cc061480c8e436d54f362f8913c

PR #542, opened 2018-09-20
Fix return type in interface

Landerstraeten:


src/Runner/TaskResultInterface.php
```diff
@@ -17,10 +17,7 @@ public function isPassed(): bool;
 
     public function isBlocking(): bool;
 
-    /**
-     * @return null|string
-     */
-    public function getMessage();
+    public function getMessage(): string;
```
janvernieuwe: The docblock says null or string, shouldn't it be `: ?string` ?

Landerstraeten: It's now `string` in the implementing `TaskResult` class

janvernieuwe: Ok, if the users aren't using this interface, this should be fine.

PR #543, opened 2018-09-21
Allow PHPLint to handle large number of files

gsomoza:
| Q             | A
| ------------- | ---
| Branch        | gsomoza/grumphp:phplint-input-stream
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | not needed
| Fixed tickets | none, but see #524

See #524: this task allows phplint to work with large number of files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "symfony/filesystem": "~2.7|~3.0|~4.0",
         "symfony/finder": "~2.7|~3.0|~4.0",
         "symfony/options-resolver": "~2.7|~3.0|~4.0",
-        "symfony/process": "~2.7|~3.0|~4.0",
+        "symfony/process": "~3.2|~4.0",
         "symfony/yaml": "~2.7|~3.0|~4.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


Landerstraeten:
Hi @gsomoza. Thanks for your PR. Can you take a look a the build? There is a problem with the lowest dependencies.

gsomoza:
Fixed it by bumping the minimum required version for `symfony/process` to `3.2`, which is where they introduced the `InputStream` class. I tried to find an approach where the dependency is injected and typed as an interface, but there's just no appropriate interface for `InputStream` so that idea is probably a dead end – unless we introduce our own interface and wrap `InputStream`, but I don't think there's enough value in doing that because the current approach is still not a BC break anyway:

According to semver.org you should still be able to release this as a non-BC break change because the public API didn't really change ([see this comment](https://github.com/semver/semver/issues/148)). 

In practice, (very old) clients that have a hard dependency on an older version of `symfony/process` will not be able to upgrade this minor bump because Composer will not let them.

So with the above in mind I'd recommend bumping minor version in this case (but of course that's up to you).

gsomoza:
🎉

PR #545, opened 2018-09-25
Fix issue when .git file has absolute path

leigh-ols:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

In the rare instance that a person has an absolute path in their .git file, grumphp fails to handle the issue.
This one-liner ensures that the path is always parsed as relative first.


veewee:
Thanks for the PR @leigh-ols !

PR #548, opened 2018-10-05
Php7 issues

richardhughes:
| Q             | A
| ------------- | ---
| Branch        | grumpy-seventies
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

When running the `composer install` from a new checkout of this repo `git clone grumphp` there are two errors. It looks like its expecting null values and I'm sure this wont affect the usage in projects but will affect those contributing to this repo. 

Solution is nullable types in 7.1 but this should fix it for PHP7

### DIFF
--- a/src/Configuration/GrumPHP.php
+++ b/src/Configuration/GrumPHP.php
@@ -33,6 +33,9 @@ class GrumPHP
         return $this->container->getParameter('git_dir');
     }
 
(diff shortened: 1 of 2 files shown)


veewee:
Thanks!

Issue #272, opened 2017-01-14
Reduce dependencies down to absolute minimum required


Bilge:
| Q               | A
| --------------- | ---
| Branch          | what
| Bug?            | no
| New feature?    | no
| Question?       | kinda
| Documentation?  | no
| Related tickets | n/a

The list of mandatory dependencies is excessive. I'd like to think many of them could be moved to either suggested or dev dependencies either immediately or with a little work. Let's make GrumPHP lighter.

patkar:
Especially the `composer/composer` dependency is a little bit harmful. We have `export PATH=./vendor/bin:$PATH` on our systems to ease the use of composer installed devtools like phpspec and co. The developer can simply execute `phpspec` instead of `vendor/bin/phpspec`. Now the installed composer dependency writes a `vendor/bin/composer` and this "overwrites" the usage of the global composer. So a `composer install/update` now executes your required composer package in vendor.

stof:
Installing composer itself looks indeed quite weird. What is the reason to have it as a dependency ?

For composer plugins, the code runs inside a composer process anyway, and so already has access to the composer classes (and installing composer a second time could even do weird things). If it is only for testing the composer plugin, it should be a dev requirement.

PR #551, opened 2018-10-15
Composer file changes

richardhughes:
| Q             | A
| ------------- | ---
| Branch        | grumpy-seventies
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

As part of the upgrade to PHP7 we should be able to clear out some old composer dependencies that require PHP5.3+. 

This should work towards making #272 easier and hopefully be able to get rid of the `|` in the versions.

I guess this is a nice to have so let me know if this is beyond the scope of the `grumpy-seventies` branch :+1: 

```
PHPUnit 7.1.4 by Sebastian Bergmann and contributors.

.............................S.SS.S.............................. 65 / 67 ( 97%)
..                                                                67 / 67 (100%)

Time: 960 ms, Memory: 10.00MB
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,9 +29,10 @@
         "phpspec/prophecy": "^1.6.2",
         "phpunit/phpunit": "^6.5|^7.0",
         "sebastian/comparator": "^2.1",
+        "sebastian/diff": "^2.0",
+        "sebastian/exporter": "^3.1",
         "sensiolabs/security-checker": "^4.0",
-        "squizlabs/php_codesniffer": "~2.4",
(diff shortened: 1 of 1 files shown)


Landerstraeten:
Hi @richardhughes. Thanks for the PR. We will test this intensively with older projects to make sure this doesn't break anything

veewee:
Hi @Landerstraeten ,

I agree with option 2. You can go ahead and merge it in once the conflict is resolved.

Landerstraeten:
Took over this orphan PR in #575

PR #556, opened 2018-10-25
Move composer dependency to dev

Landerstraeten:


veewee:
HI @Landerstraeten,

This won't do:

- `composer-plugin-api`: This one is required for the composer plugin to work. More information can be found here: https://getcomposer.org/doc/articles/plugins.md#creating-a-plugin
- `composer/composer` is used during runtime. See:
  - Configure command
  - `\GrumPHP\Util\Composer`
  - `\GrumPHP\Console\Helper\ComposerHelper`


It is possible to remove the composer/composer dependency if we rewrite the 3 items mentioned above. It can be done by just manually parsing the root composer.json file of the project instead of relying on the composer loader / factory.

src/Console/Application.php
```diff
@@ -183,17 +182,10 @@ protected function initializeComposerHelper(): Helper\ComposerHelper
...
-        try {
-            $composerFile = getcwd().DIRECTORY_SEPARATOR.'composer.json';
-            $configuration = Composer::loadConfiguration();
-            Composer::ensureProjectBinDirInSystemPath($configuration->get('bin-dir'));
-            $rootPackage = Composer::loadRootPackageFromJson($composerFile, $configuration);
-        } catch (RuntimeException $e) {
-            $configuration = null;
-            $rootPackage = null;
-        }
+        $composerFileLocation = getcwd().DIRECTORY_SEPARATOR.'composer.json';
```
janvernieuwe: Is this path always guaranteed correct?
if it can't find the composer.json file, grumphp will crash on `file_get_contents`

src/Util/ComposerFile.php
```diff
@@ -0,0 +1,79 @@
...
+
+        throw new RuntimeException('Could not get bin dir location');
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getConfigDefaultPath()
+    {
+        return null;
```
veewee: Shouldn't you check and return following param?
`$extra['grumphp']['config-default-path'];`

src/Console/Application.php
veewee: No it is not always available. probably better to detect composer.json from current directory and from all parents until one is found?

composer.json
```diff
@@ -22,6 +21,7 @@
         "symfony/yaml": "~2.7|~3.0|~4.0"
     },
     "require-dev": {
+        "composer/composer": "^1.0",
```
veewee: No need for this dependency anymore right?

src/Util/ComposerFile.php
Landerstraeten: Ow yes. Clearly something I forgot todo.

composer.json
Landerstraeten: Not sure. I check this in Symfony Flex. There, they do it too.

src/Console/Application.php
Landerstraeten: There is a check in ComposerFile.php on line 21

src/Util/ComposerFile.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GrumPHP\Util;
+
+use GrumPHP\Exception\RuntimeException;
+
+class ComposerFile
```
Landerstraeten: - [ ] Add missing test

src/Util/ComposerFile.php
```diff
@@ -0,0 +1,73 @@
...
+            $_SERVER[$pathStr] = $binDir.PATH_SEPARATOR.getenv($pathStr);
+            putenv($pathStr.'='.$_SERVER[$pathStr]);
+        }
+
+        return true;
+    }
+
+    public function getBinDir(): string
+    {
+        return $this->configuration['bin-dir'] ?? 'vendor/bin';
```
Landerstraeten: - [ ] Validate folder exists

src/Console/Application.php
veewee: In the old codebase, there was a catch which returns an empty composer info.
Maybe it's best to catch the error from ComposerFile line 21 and also return an empty Composer instance?
That way, the code keeps on working if there is no composer.json in the cwd (which probably will happen more then once... ;) )

src/Console/Command/ConfigureCommand.php
```diff
@@ -183,22 +182,9 @@ protected function writeConfiguration(array $configuration): bool
...
-            return $defaultBinDir;
-        }
-
-        $config = $this->composer()->getConfiguration();
-        if (!$config->has('bin-dir')) {
-            return $defaultBinDir;
-        }
-
-        return $config->get('bin-dir', Config::RELATIVE_PATHS);
+        return $this->composer()->getComposerFile()->getBinDir();
```
veewee: Maybe also safer to force a relative path as in the old codebase? Not sure too what folder the old code is making the path relative

veewee:
@Landerstraeten  : Can you tell me what the state of this PR is? I'dd love to get this one merged in.
I'll be bumping the composer version in order to get #638 to work.
It seems better to use `composer/composer:^1.8` as a dev dependency and check if our composer plugin also works with version `composer/composer:^1.1` (which is the latest)

PR #559, opened 2018-10-26
Host our own version of the symfony XSDs

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

Currently tests are failing based on a XSD that isn't available anymore.
I've moved them to a gist which I won't remove.


cafferata:
Is it not an better idea to add this file to the GrumPHP git repository, and to the .gitattributes? Now you have created an (unnecessary) external dependency in my opinion.

Issue #501, opened 2018-04-17
FR: remove need for dedicated class when running same task twice with different configuration


keradus:
docs refs: https://github.com/phpro/grumphp/pull/483/files#diff-ccf4f1a1622265fd2805632c80129a8bR193
original request: https://github.com/phpro/grumphp/issues/309#issuecomment-377535886

I'm opening new ticket, as one in which it was requested is closed - original ticket was about documenting status quo, this one is to bring better User eXerience.

This gonna be super useful for running different shell tasks, which obviously has same implementation, just different configuration, and making new dummy-class for every single shell-task is way not friendly

keradus:
> Since we currently match the name based on the task name

@veewee , can you point place of code when this is happening, and maybe propose direction how it could be changed ?

veewee:
Hi @keradus,

Sorry for the late response.
I can point you to following locations:

https://github.com/phpro/grumphp/blob/master/doc/tasks.md#creating-a-custom-task
This information is on how the tasks are configured / registered in the service container. 
You can see that the task service contains a config key. This one is linked through a compiler pass.

https://github.com/phpro/grumphp/blob/master/src/Configuration/Compiler/TaskCompilerPass.php
This is the compiler pass that is responsible for registering the task in the Taskrunner, normalizing the configuration and storing it into new variables inside the service container. Those variables can be accessed by the `GrumPHP` configuration object at: https://github.com/phpro/grumphp/blob/master/src/Configuration/GrumPHP.php

https://github.com/phpro/grumphp/blob/master/src/Task/TaskInterface.php#L14
Finally, there is the `getName()` method inside the task which is being used in the `TaskRunner` to filter, report, fetch additional configutation, ...


Currently, I do not have a direction proposition on the issue. Since the name is used in 3 locations, I think a first phase would be to make it easier to configure and register tasks internally. 
Another thing I am thinking of, is to make it possible to create an aliases system in the configuration. However: since the `Task::getName()` is being used inside the reporting, this might require a BC break inside the `TaskInterface` - which I would like to avoid at the moment.

I hope this gives you a better understanding on the issue. Feel free to ask additional information or to propose a plan of action.

keradus:
indeed, looks like a bigger thing indeed 

thanks for detailed description, I'm not feeling able to do it myself right now, but let us have this FR open please ;) it's not first time it's requested, and now we have more info than single-line comment in src why one has to create custom task.

RaphTbm:
Hi there,

[...]

PR #565, opened 2018-11-12
Allow multi-tasking w/o dedicated class.

RaphTbm:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets |  #309, #501

No BC-break proposal to allow re-use of existing tasks without the need to create a package with new redundant classes, for more versatility.

The `grumphp.yml` file should like:

````yml
parameters:
  tasks:
    shell:
      scripts:
        - scripts/ci/global.sh
    shell.myspecifictask1:
      scripts:
        - scripts/ci/some-script.sh
      triggered_by: [scss]
    shell.myspecifictask2:
      scripts:
        - scripts/ci/some-other-script.sh
      triggered_by: [js]

services:
  task.shell.newservice1:
    parent: task.shell
    tags:
      - {name: grumphp.task, config: shell.myspecifictask1}
  task.shell.newservice2:
    parent: task.shell
    tags:
      - {name: grumphp.task, config: shell.myspecifictask2}
````

I'm just using the possibility of Symfony services to extend an existing service with the "parent" key (i'm absolutely not a Symfony expert, but it seems to be the best method). For me, it was making sense to keep this direction and re-use the existing possibilities offered by Symfony. It allows to almost not change the current code managing the service + tasks, very few lines.
[https://symfony.com/doc/current/service_container/parent_services.html](url)

When time of a BC break will come, `getName` should be added in the interface and the method_exists test for `addTask` in `TaskRunner` should be removed.

If accepted, I will add a new section in the existing "_Run the same task twice with different configuration_" of `tasks.md`.

Help / contribution welcome! Cheers.

### DIFF
--- a/spec/Task/AntSpec.php
+++ b/spec/Task/AntSpec.php
@@ -24,6 +24,7 @@ class AntSpec extends ObjectBehavior
     {
         $grumPHP->getTaskConfiguration('ant')->willReturn([]);
         $this->beConstructedWith($grumPHP, $processBuilder, $formatter);
+        $this->setName('ant');
     }
 
     function it_is_initializable()
--- a/spec/Task/AtoumSpec.php
+++ b/spec/Task/AtoumSpec.php
@@ -24,6 +24,7 @@ class AtoumSpec extends ObjectBehavior
     {
         $grumPHP->getTaskConfiguration('atoum')->willReturn([]);
         $this->beConstructedWith($grumPHP, $processBuilder, $formatter);
+        $this->setName('atoum');
     }
 
     function it_is_initializable()
--- a/spec/Task/BehatSpec.php
+++ b/spec/Task/BehatSpec.php
@@ -24,6 +24,7 @@ class BehatSpec extends ObjectBehavior
     {
         $grumPHP->getTaskConfiguration('behat')->willReturn([]);
         $this->beConstructedWith($grumPHP, $processBuilder, $formatter);
+        $this->setName('behat');
     }
 
     function it_is_initializable()
--- a/spec/Task/BrunchSpec.php
+++ b/spec/Task/BrunchSpec.php
@@ -24,6 +24,7 @@ class BrunchSpec extends ObjectBehavior
     {
         $grumPHP->getTaskConfiguration('brunch')->willReturn([]);
         $this->beConstructedWith($grumPHP, $processBuilder, $formatter);
+        $this->setName('brunch');
     }
 
     function it_is_initializable()
--- a/spec/Task/CloverCoverageSpec.php
+++ b/spec/Task/CloverCoverageSpec.php
@@ -19,6 +19,7 @@ class CloverCoverageSpec extends ObjectBehavior
     {
         $grumPHP->getTaskConfiguration('clover_coverage')->willReturn([]);
         $this->beConstructedWith($grumPHP, new Filesystem());
+        $this->setName('clover_coverage');
     }
 
     function it_is_initializable()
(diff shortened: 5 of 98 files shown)


RaphTbm:
If somebody have a clue of where to look to succeed the tests, i'm taking it!
It seems that `setName` have to be initiated somewhere.

RaphTbm:
Seems I was on the right way! :D

RoSk0:
Looking at the services section in Good idea, I was thinking about it myself but never had time to implement. Looking at the services section in the grumphp.yml I think we can take it a step forward and improved developer experience even more by removing the need of declaring services here and instead declare them in the backend. For example if we will introduce documented naming conversion, for example, "{EXISTING_TASK}.{CHILD_TASK}" where EXISTING_TASK would be "shell" and CHILD_TASK "myspecifictask1" from example above we will be able to say - "right, we have a new 'shell' task with another set of parameters" and define required service.

RaphTbm:
Hey RoSk0, I like the idea!

So, keep the possibility to define new task services as currently, but if you only need to use the existing task services, don't bother to define them but generate the extra services automatically (based on the defined task configuration). But the current system seems to rely on the symphony services as hell and i'm absolutely not a symphony services expert.
My simple idea, to stay as much as possible in the current system, would be to replace the current provided task service file by a new one generated automatically (merging tasks configuration + tasks services logical).

**So for something like this:**

File `grumphp.yml`

````yml
parameters:
  tasks:
    shell:
      scripts:
        - my_script
    shell.CHILD_TASK:
      scripts:
        - my_script_2
    shell.CHILD_TASK_2:
      scripts:
        - my_script_3
````

Existing `resources/config/tasks.yml`

````yml
services:
    task.shell:
        class: GrumPHP\Task\Shell
        arguments:
            - '@config'
            - '@process_builder'
            - '@formatter.raw_process'
        tags:
            - {name: grumphp.task, config: shell}
````

**It would give:**

A new file `resources/generated/tasks.yml` (provided to the symphony system in replacement of the existing `resources/config/tasks.yml`):

````yml
services:
    task.shell:
        class: GrumPHP\Task\Shell
        arguments:
            - '@config'
            - '@process_builder'
            - '@formatter.raw_process'
        tags:
            - {name: grumphp.task, config: shell}
    task.shell.CHILD_TASK
    	parent: task.shell
        tags:
            - {name: grumphp.task, config: CHILD_TASK}
    task.shell.CHILD_TASK_2
        parent: task.shell
            tags:
             - {name: grumphp.task, config: CHILD_TASK_2}
````

Where `CHILD_TASK` is the config key and so on the service key. 

It would'nt break the existing possibilities supposedly.
But going that way, could probably lead to a BC-break situation (I'm not sure).

Cheers.

veewee:
Hello @RaphTbm,

Thanks for your work!

There are some things I would like to avoid in current PR:
- usage of traits
- dynamic names being set in the TaskRunner (it's not the job of the taskRunner to set a name). It should be set before the task is injected in the runnen.

(That is why we won't be merging this PR.)

Good ideas:

- usage of parent services make it easier to inject dependencies.

There are some good points in the comments above as well.
Although I would avoid generating a custom configuration file, but there are some things we could do by using a compiler pass instead.

Probably the biggest problem at this moment is that the task loads it's configuration through the Grumphp configuration object. It would possibly be easier if we injected the actual configuration in the task instead.

I'll discuss this internally with @Landerstraeten and we'll see what we can do about this issue.

Landerstraeten:
@RaphTbm. Big thanks for all the work on this PR. @veewee and I will pick this up later

PR #566, opened 2018-11-12
Add a new extra parameter for task labels

RaphTbm:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | none

Following #565, add the possibility to define a custom label for more readability when multiple tasks are defined.

Usage in `grumphp.yml` would be:

````yml
parameters:
  tasks:
    shell:
      scripts:
        - scripts/ci/basic.sh
      _extra:
        label: 'Shell basic'
````

### DIFF
--- a/src/Event/Subscriber/ProgressSubscriber.php
+++ b/src/Event/Subscriber/ProgressSubscriber.php
@@ -72,8 +72,12 @@ class ProgressSubscriber implements EventSubscriberInterface
      */
     public function advanceProgress(TaskEvent $event)
     {
-        $taskReflection = new ReflectionClass($event->getTask());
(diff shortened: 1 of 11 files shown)


veewee:
Hi @RaphTbm,

Thanks for the PR.

A better place for this label would be in the metadata section, which is created for such things.
More information could be found here: https://github.com/phpro/grumphp/blob/master/doc/tasks.md#metadata

That way, we don't have to alter any configuration / tasks. You could just load the metadata in the ProgressSubscriber.

Can you also take a look at the failing tests?

PR #567, opened 2018-11-13
Add the files as argument for Shell task

RaphTbm:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | none

Add the possibility to append the files as an argument for the running scripts of Shell tasks.

**3 new options:** 

- include_files: append the files or not for the defined scripts / FALSE (default)
- include_files_parameter : name of the files argument if needed
- include_files_with_comma: files argument separated with a comma or not, if not, each file become a new argument for the script / TRUE (default)

**Examples:**

````yml
parameters:
  tasks:
    shell:
      scripts:
        - my_script
      include_files: true
````
Will be: `/bin/sh my_script file1,file2,...`

````yml
parameters:
  tasks:
    shell:
      scripts:
        - my_script
      include_files: true
      include_files_with_comma: false
````
Will be: `/bin/sh my_script file1 file2 ...`

````yml
parameters:
  tasks:
    shell:
      scripts:
        - my_script
      include_files: true
      include_files_parameter: 'arg-name'
````
Will be: `/bin/sh my_script arg-name=file1,file2,...`

### DIFF
--- a/spec/Task/ShellSpec.php
+++ b/spec/Task/ShellSpec.php
@@ -44,6 +44,9 @@ class ShellSpec extends ObjectBehavior
         $options->shouldBeAnInstanceOf(OptionsResolver::class);
         $options->getDefinedOptions()->shouldContain('scripts');
         $options->getDefinedOptions()->shouldContain('triggered_by');
+        $options->getDefinedOptions()->shouldContain('include_files');
+        $options->getDefinedOptions()->shouldContain('include_files_with_comma');
+        $options->getDefinedOptions()->shouldContain('include_files_parameter');
     }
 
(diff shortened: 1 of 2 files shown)


veewee:
Hello @RaphTbm,

Thanks for the PR!
I am not sure we should allow all these different ways of passing files. Maybe it is better to provide a list of files through STDIN instead of cli arguments? Otherwise we will probably have to deal with the max line input issue as well.
Can you also take a look at the failing tests?

PR #569, opened 2018-11-13
Allow PHP_CodeSniffer to use multiple standards.

RoSk0:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes?
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | None

As subject reads - Allow PHP_CodeSniffer to use multiple standards.


src/Task/Phpcs.php
```diff
@@ -46,7 +46,7 @@ public function getConfigurableOptions()
             'report_width' => null,
         ]);
 
-        $resolver->addAllowedTypes('standard', ['null', 'string']);
+        $resolver->addAllowedTypes('standard', ['array']);
```
veewee: Can you allow both array and string and cast a string to an array? That way, old configuration doesnt break.

RoSk0: Yeah, that was my first attempt at this, but it wasn't working as validation would fail .

RoSk0: I will try to make it work tomorrow..

RoSk0: Hi @veewee,

Turns out it actually possible - so the configuration is now backward compatible! Win. 
Thanks for bringing it up!

I've done a forced push so you may want to review it again.

Issue #580, opened 2019-01-03
Provide option to run a single/list of tasks from the config


paslandau:
| Q               | A
| --------------- | ---
| Version         | `0.14.3`
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | probably
| Related tickets | 

It would be great to have an option in the `run` command to specify a list of tasks that should be run by grumphp instead of running all tasks

Proposal
==
Consider the following grumphp.yml

```yml
parameters:
  tasks:
    composer: []
    composer_require_checker:
      ignore_parse_errors: true
    file_size:
      max_size: 1M
```

Run grumphp via

 `vendor/bin/grumphp run --tasks=composer,file_size`

The `--tasks` option would be split on "," and the resulting values would act as a "whitelist" for the given config. I.e. only `composer` and `file_size` would run:

```
Running task  1/2: Composer... ✔
Running task  2/2: FileSize... ✔
```

`composer_require_checker` would be omitted since it is not listed in the option.

If `--tasks` is empty / not given, the default is the current behavior (maintaining bc).

Use cases
===
- Being able to easily re-run a specific task
  - On "long" task chains this helps in verifying the reported issue is fixed without the need of running every other (probably unrelated) check as well
- Quickly explore "new" tasks. 
  - Currently, I need to duplicate my grumphp.yml, remove all tasks and keep only the required ones. This is really inconvenient, especially when setting up grumphp initially
- Sometimes, I only need to run a specific taks / combination of tasks (e.g. parallel-lint and phpstan in order to quickly verify that there are no syntactical errors that will interrupt a subsequent unit/integration/end2end test run [which usually takes much longer])
- Keep only one config file for different scenarios. 
  - We use grumphp not only for commit hooks but also for general code checks on the "full" codebase, e.g. before we create a pull request for code review. Those include an extented set of tasks (e.g. composer-require-checker) that needs to only run before an actual push to remote but not on every commit. Currently, we need to maintain separate files to archieve this. With the `--tasks` option we could keep one file and pick the subset of tools that we need. 

Implementation
===
I was having a quick look into the run command and this should be achievable with minimal effort, i.e.
- add `tasks` option to `RunCommand` & parse comma separated string
- add `tasks` property to `TaskRunnerContext`
- in `TaskRunner::run()` filter `TaskCollection` based on the task name, i.e.
```php
        $tasks = $this->tasks
            ->filterByContext($runnerContext->getTaskContext())
[...]

Issue #581, opened 2019-01-03
FR: Add parallel execution option and task execution stages


paslandau:
| Q               | A
| --------------- | ---
| Version         | `0.14.3`
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | probably
| Related tickets | 

On bigger code bases and / or long task chains, the grumphp execution can take a considerable amount of time. Most of the performed tasks are probably unrelated and don't affect each other (e.g. large_file and phpstan can run independently). Thus, running those tasks in parallel would greatly improve performance.

On the other hand, there might be tasks that **should** run in a certain order, e.g. `parallel-lint` should usually be run before `phpcs` or `phpstan` as failing early makes sense in this case. Thus, there should be a way to "group" tasks and define an execution order. 

Proposal
==
Consider the following grumphp.yml

```yml
parameters:
  processes: 4
  tasks:
    parallel-lint: 
      stage: 0
    composer: 
      stage: 1
    composer_require_checker:
      stage: 1
```
`processes` denotes the number of parallel running tasks (can be omitted or 1 for to keep current behavior / avoid BC)
`stage` acts a an alternative to the already supported `priority` option. The value has to be an integer so that 
- stages can be run in a predefined order (same logic as `priority`)
- tasks can be grouped into one stage
- all tasks in the same stage can run in parallel
-`priority` is ignored if stages are used

Use cases
===
- Already mentioned in the description: This will greatly (usually ~linear to the number of cores) improve performance.

Implementation
===
This is probably a bigger change. Possible solution:
- use a task runner that already implements parallel execution capabilities (e.g. Robos [ParallelExec](https://robo.li/tasks/Base/#parallelexec))
  - if that is not desirable / to big of a change, re-implementing https://github.com/consolidation/Robo/blob/954f17bfbdd78b829c0eae88055a0446ba589aed/src/Task/Base/ParallelExec.php does not seem to be an overwhelming task (this would actually have the benefit of sneaking "throttling" in there as well ;))
- if https://github.com/phpro/grumphp/issues/580 is implemented, the problem can be reduced to "find the tasks for the current stage" and "start those tasks as individual grumphp tasks" - although the specifics of how to handle / merge the results have to be hashed out. Not deep enough in the code to judge that :|
- since parallel execution usually adds another layer of complexity, I would introduce that feature explicitly as "beta" to get some real world feedback before "officially" announce it as supported.

Thoughts?

Cheers
Pascal

PR #582, opened 2019-01-03
580: Adds support for --tasks option in run command.

paslandau:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | https://github.com/phpro/grumphp/issues/580

Note: Sorry for the ugly Reflection hack to test the protected method in `RunCommandTest` - can't use anonmous classes here due to PHP >= 5.6 constraint.

### DIFF
--- a/src/Collection/TasksCollection.php
+++ b/src/Collection/TasksCollection.php
@@ -40,6 +40,21 @@ class TasksCollection extends ArrayCollection
     }
 
     /**
+     * @param string[] $tasks
+     *
+     * @return TasksCollection
+     */
+    public function filterByTaskName($tasks)
+    {
+        return $this->filter(function (TaskInterface $task) use ($tasks) {
+            if (empty($tasks)) {
+                return true;
+            }
+            return in_array($task->getName(), $tasks);
+        });
+    }
+
+    /**
      * This method sorts the tasks by highest priority first.
      *
      * @param GrumPHP $grumPHP
--- a/src/Console/Command/RunCommand.php
+++ b/src/Console/Command/RunCommand.php
@@ -53,6 +53,13 @@ class RunCommand extends Command
             'Specify which testsuite you want to run.',
             null
         );
+        $this->addOption(
+            'tasks',
+            null,
+            InputOption::VALUE_REQUIRED,
+            'Specify which tasks you want to run (comma separated). Example --tasks=task1,task2',
+            null
+        );
     }
 
     /**
@@ -66,9 +73,12 @@ class RunCommand extends Command
         $files = $this->getRegisteredFiles();
         $testSuites = $this->grumPHP->getTestSuites();
 
+        $tasks = $this->parseCommaSeparatedOption($input->getOption("tasks"));
+
         $context = new TaskRunnerContext(
             new RunContext($files),
-            (bool) $input->getOption('testsuite') ? $testSuites->getRequired($input->getOption('testsuite')) : null
+            (bool) $input->getOption('testsuite') ? $testSuites->getRequired($input->getOption('testsuite')) : null,
+            $tasks
         );
 
         return $this->taskRunner()->run($output, $context);
@@ -96,5 +106,26 @@ class RunCommand extends Command
     protected function paths()
     {
         return $this->getHelper(PathsHelper::HELPER_NAME);
+    }
+
+    /**
+     * Split $value on ",", trim the individual parts and
+     * de-deduplicate the remaining values
+     *
+     * @param string $value
+     * @return string[]
+     */
+    protected function parseCommaSeparatedOption($value)
+    {
+        $stringValues = explode(",", $value);
+        $parsedValues = [];
+        foreach ($stringValues as $k => $v) {
+            $v = trim($v);
+            if (empty($v)) {
+                continue;
+            }
+            $parsedValues[$v] = $v;
+        }
+        return $parsedValues;
     }
 }
--- a/src/Runner/TaskRunnerContext.php
+++ b/src/Runner/TaskRunnerContext.php
@@ -38,7 +38,7 @@ class TaskRunnerContext
     {
         $this->taskContext = $taskContext;
         $this->testSuite = $testSuite;
-        if($tasks === null){
+        if ($tasks === null) {
             $tasks = [];
         }
         $this->tasks = $tasks;
--- a/src/Runner/TaskRunnerContext.php
+++ b/src/Runner/TaskRunnerContext.php
@@ -23,15 +23,25 @@ class TaskRunnerContext
     private $testSuite = null;
 
     /**
+     * @var string[]
+     */
+    private $tasks = null;
+
+    /**
      * TaskRunnerContext constructor.
      *
-     * @param ContextInterface   $taskContext
+     * @param ContextInterface $taskContext
      * @param TestSuiteInterface $testSuite
+     * @param string[]|null $tasks
      */
-    public function __construct(ContextInterface $taskContext, TestSuiteInterface $testSuite = null)
+    public function __construct(ContextInterface $taskContext, TestSuiteInterface $testSuite = null, $tasks = null)
     {
         $this->taskContext = $taskContext;
         $this->testSuite = $testSuite;
+        if($tasks === null){
+            $tasks = [];
+        }
+        $this->tasks = $tasks;
     }
 
     /**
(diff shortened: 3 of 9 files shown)


src/Collection/TasksCollection.php
```diff
@@ -39,6 +39,21 @@ public function filterByTestSuite(TestSuiteInterface $testSuite = null)
...
 
+    /**
+     * @param string[] $tasks
+     *
+     * @return TasksCollection
+     */
+    public function filterByTaskName($tasks)
+    {
+        return $this->filter(function (TaskInterface $task) use ($tasks) {
+            if (empty($tasks)) {
```
veewee: No need to iterate through the list when the tasks are empty. You can put the empty check above the filter function and return a new collection through `return new TasksCollection($this->toArray());`
See `\GrumPHP\Collection\TasksCollection::filterByTestSuite()`

src/Collection/TasksCollection.php
```diff
@@ -39,6 +39,21 @@ public function filterByTestSuite(TestSuiteInterface $testSuite = null)
         });
     }
 
+    /**
+     * @param string[] $tasks
+     *
+     * @return TasksCollection
+     */
+    public function filterByTaskName($tasks)
```
veewee: Can you rename this one to `filterByTaskNames`?

src/Collection/TasksCollection.php
```diff
@@ -39,6 +39,21 @@ public function filterByTestSuite(TestSuiteInterface $testSuite = null)
...
+    {
+        return $this->filter(function (TaskInterface $task) use ($tasks) {
+            if (empty($tasks)) {
+                return true;
+            }
+            return in_array($task->getName(), $tasks);
```
veewee: You can set the third argument to true here.

src/Console/Command/RunCommand.php
```diff
@@ -66,9 +73,12 @@ public function execute(InputInterface $input, OutputInterface $output)
         $files = $this->getRegisteredFiles();
         $testSuites = $this->grumPHP->getTestSuites();
 
+        $tasks = $this->parseCommaSeparatedOption($input->getOption("tasks"));
```
veewee:
```suggestion
        $tasks = array_map('trim', explode(',', $input->getOption('tasks', '')));
```

The de-duplicating logic is a bit confusing and is not required since the filter method on the taskcollection is smart enough. I would prefer to drop it for the single line suggested above. (optionally you could add array_filter to drop the empty items, but that is also covered by the filter method)

src/Runner/TaskRunnerContext.php
```diff
@@ -22,16 +22,26 @@ class TaskRunnerContext
...
     /**
      * TaskRunnerContext constructor.
      *
-     * @param ContextInterface   $taskContext
+     * @param ContextInterface $taskContext
      * @param TestSuiteInterface $testSuite
+     * @param string[]|null $tasks
      */
-    public function __construct(ContextInterface $taskContext, TestSuiteInterface $testSuite = null)
+    public function __construct(ContextInterface $taskContext, TestSuiteInterface $testSuite = null, $tasks = null)
```
veewee: Can you make $tasks a required array?
I would prefer to always pass in a list of tasks.
(This is a BC break, but I don't think it will cause too much damage since it is an internal DTO.)

test/Console/Command/RunCommandTest.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace GrumPHPTest\Console\Command;
+
+use GrumPHP\Configuration\GrumPHP;
+use GrumPHP\Console\Command\RunCommand;
+use GrumPHP\Locator\RegisteredFiles;
+use PHPUnit\Framework\TestCase;
+
+class RunCommandTest extends TestCase
```
veewee: This test can be dropped if you change the input option logic as suggested above.

paslandau: I'dont feel quite comfortable having "nothing" that actually checks the behavior - which kinda was the initial thought on adding a test in the first place.

E.g. consider the inlined
```php
 $tasks = array_map('trim', explode(',', $input->getOption('tasks', '')));
```
If somebody were to switch "," by ";" there would be "nothing" to catch that bug. Mutation testing tools like https://github.com/infection/infection usually "make those changes" to verify if everything is tested "correctly".

---

Are you open to introducing some sort of end2end/integration test that "acutally" calls the run command and verifies the output (see https://symfony.com/doc/4.1/console.html#testing-commands )?

It's something that we do quite frequently because it's a very good high level indicator for "some things don't work as they are supposed to" (though it comes with an increased maintenance overhead due to duplication).

Something along the lines of

```php
    public function test_command()
    {
        $application = new Application();

        /**
         * @var RunCommand $command
         */
        $command = $application->find('run');

        $commandTester = new CommandTester($command);
        $commandTester->execute(array(
            'command'  => $command->getName(),
            '--tasks' => 'foo',
        ));

        $output = $commandTester->getDisplay();
        $this->assertContains('...', $output);
    }
```

src/Console/Command/RunCommand.php
paslandau: This will retain empty input values, i.e. if `--tasks` is empty or omitted, `$tasks` will be an array containing one empty element. So we would either put it in an if OR add and `array_filter` --- both makes the result less readable, imho.

I'm fine with removing deduplication, though.

Thoughts?

fyi - failing tests after change:
```
/codebase/grumphp/vendor/phpunit/phpunit/phpunit --configuration /codebase/grumphp/phpunit.xml.dist GrumPHPTest\\Console\\Command\\RunCommandTest /codebase/grumphp/test/Console/Command/RunCommandTest.php

[...]
4) GrumPHPTest\Console\Command\RunCommandTest::parses_comma_separated_options with data set "null" (null, array())
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
+    0 => ''
 )

/codebase/grumphp/test/Console/Command/RunCommandTest.php:36

5) GrumPHPTest\Console\Command\RunCommandTest::parses_comma_separated_options with data set "empty" ('', array())
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
+    0 => ''
 )

/codebase/grumphp/test/Console/Command/RunCommandTest.php:36

6) GrumPHPTest\Console\Command\RunCommandTest::parses_comma_separated_options with data set "empty after trim" (' ', array())
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
+    0 => ''
 )
```

test/Console/Command/RunCommandTest.php
veewee: Hi @paslandau,

In a normal situation, those tests would work fine.
However, the behaviour of the grumphp commands are different based on which configuration file you supply. This means we would have a lot of e2e tests to cover all possible situations.
I was looking in to this with @Landerstraeten a while ago, but didn't find a good solution yet.

Feel free to suggest something in this PR.

src/Runner/TaskRunnerContext.php
paslandau: Like this?

```php
    /**
     * TaskRunnerContext constructor.
     *
     * @param ContextInterface $taskContext
     * @param string[] $tasks
     * @param TestSuiteInterface $testSuite
     */
    public function __construct(ContextInterface $taskContext, array $tasks, TestSuiteInterface $testSuite = null)
```

veewee: yes

paslandau:
Hey @veewee 

would you like to review https://github.com/paslandau/grumphp/commit/6b1662836fe373c900650d9da3b459948bfade65 before  I make a PR against seventies?

test/Console/Command/RunCommandTest.php
paslandau: I'll give it some thought but I'm not to deep into symfony (more of a laravel guy) plus since those sort of test are usually tidly coupled to the application I guess I'd need more time digging some grumphp internals.

So I would keep i separate from this PR + keep the current test for now.

veewee:
Hi @paslandau,

The branch looks OK but it's easier to review in a PR. You can change this PR or create a new one.

paslandau:
Hey @veewee 

done: https://github.com/phpro/grumphp/pull/587

I also created a POC for parallel execution (see https://github.com/phpro/grumphp/issues/581) that would be a successor of this PR but contains substantially more changes.

veewee:
Thanks! Continuing in #587.

Issue #584, opened 2019-01-04
FR: Add paratestphp/paratest task


paslandau:
| Q               | A
| --------------- | ---
| Version         | new release
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | yes
| Related tickets | 

- https://github.com/paratestphp/paratest
- run phpunit in parallel
- greatly improved performance
- would only implement this for the `grumpy_seventies` branch

Would create a PR if interested.

Landerstraeten:
Done in #617 by @janvernieuwe

Issue #394, opened 2017-08-16
Provide mechanism to run GrumPHP without Git for a list of specified files


Bilge:
| Q               | A
| --------------- | ---
| Version         | n/a
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | n/a

GrumPHP currently relies on Git to feed it the list of files to process. However, it is a common requirement to want to reuse GrumPHP's functionality in contexts where the list of files may come from another source. It should be possible for the user (or a script) to specify a list of files to check on GrumPHP's command line.

An example use case might be compiling a list of all files changed between two commit refs and asking GrumPHP to validate them.

As a consideration towards future work, consider how support might be added to specify directories that are recursively searched in addition to specific files.

veewee:
Sounds like a good but rather complex idea. Thanks for adding it to the feature list :)

thomscode:
Theoretically it's not too complicated. If GrumPHP had a way to accept a list of files from the command line `grumphp run <space separated file/folder list>` then the burden of providing that listing would be on the user. If the user wants the list to be fed from a list of files in a diff, then that user could write a diff parser to pull the file listing.

Slightly more complicated is to allow plugins that could parse input and provide a list of files. In this way users can write plugins themselves that fit their needs. If you pursue this way, then the file list is simply another parser that explodes the space separated list into an array.

Anyway, just a couple of ideas.

UrGuardian4ngel:
Or maybe you could read the file list from `STDIN` when a certain flag is defined, like e.g. `-i|--from-stdin`?

```bash
# from e.g. a static file list
$ cat file-list.txt | grumphp run --from-stdin

# extracted from a git diff
$ git diff HEAD~3... --name-only | grumphp run --from-stdin

# or generated by any other tool (in this case, silver searcher)
$ ag -l --php -- src/AppBundle/SomeComponent | grumphp run --from-stdin
```

veewee:
@UrGuardian4ngel : in the pre-commit task, we allready support this. We do need a git diff at this moment though. (See `ChangedFiles` locator for more details)

UrGuardian4ngel:
Oh yes. But I was assuming he was talking about the `grumphp run` command, as per the issue title:

> Provide mechanism to **run GrumPHP** without Git **for a list of specified files**

which is using `RegisteredFiles` (`$ git ls-files`), and depends on a `Git\Repository`.

[...]

PR #586, opened 2019-01-07
Introducing parallel execution of tasks

paslandau:
| Q             | A
| ------------- | ---
| Branch        | seventies
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no (hopefully?)
| Deprecations? | no
| Documented?   | no (poc => please feedback)
| Fixed tickets | #584,#581,#580,#394

## tl;dr
- made some bigger changes to allow parallel execution of tasks
- currently only works for external tasks - internal tasks like the `linting` would have to be exposed as a command line executable
- introduced e2e/integration tests (because I don't know PhpSpec :D)
- for this "test", I added some depencies to `composer.json` and added a `grumphp_parallel.yml`,
  so you can run `composer install && vendor/bin/grumphp run --config="grumphp_parallel.yml"`
  to see the thing in action. I also add `verbosity` output to see the actual command, see
  `vendor/bin/grumphp run --config="grumphp_parallel.yml" -vvv` (messy, but helpful for understanding a task)
- this is a POC and I realise that I might have a somewhat different coding style, sorry for that.
  I don't cling to any of the implemenation, I just wanted to get something "runnable" to get an impression
  of how a parallel execution might look like and I think that this might actually be a pretty nice
  extension to `grumphp` :)

See https://github.com/paslandau/grumphp/blob/parallel/CHANGELOG.md for further info.


veewee:
Parallel tasks are implemented in #741. Already borrowed the good ideas from this PR already :) Thanks!

PR #587, opened 2019-01-07
 Provide option to run a single/list of tasks from the config

### LINKED ISSUE #580: Provide option to run a single/list of tasks from the config

paslandau:
| Q             | A
| ------------- | ---
| Branch        | master 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | https://github.com/phpro/grumphp/issues/580

Fixes #580


src/Console/Command/RunCommand.php
```diff
@@ -83,4 +93,25 @@ protected function paths(): PathsHelper
...
     }
+
+    /**
+     * Split $value on ",", trim the individual parts and
+     * de-deduplicate the remaining values
+     *
+     * @param string $value
+     * @return string[]
+     */
+    protected function parseCommaSeparatedOption(string $value)
```
veewee: It is better to move this method to the `\GrumPHP\Util\Str` class.
Maybe something like:
```
public static function explode(string $delimiter, string $string): array
```

This way the logic can be tested in a normal way and it is reusable if we want to add more comma separated options?

Landerstraeten:
Hi @paslandau. Thanks for your PR! Can you rebase against master? I will review it soon.

src/Collection/TasksCollection.php
```diff
@@ -34,6 +34,20 @@ public function filterByTestSuite(TestSuiteInterface $testSuite = null): self
         });
     }
 
+    /**
+     * @param string[] $tasks
+     */
+    public function filterByTaskNames(array $tasks): self
+    {
+        if (empty($tasks)) {
+            return new self($this->toArray());
```
veewee: `return $this;` ? No need to create  a new instance in that case right?

Issue #576, opened 2018-12-11
[PHP73] Deprecation warning


veewee:
| Q               | A
| --------------- | ---
| Version         | 0.14.2
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

```
Deprecated: strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior in xxx/vendor/phpro/grumphp/src/Locator/ConfigurationFile.php on line 75
```

janvernieuwe:
Is this already fixed?
Otherwise can I have some quick steps to reproduce this?

```
~/Projects/grumphp$ /usr/bin/php7.3 vendor/bin/grumphp run
GrumPHP is sniffing your code!
Running task 1/7: PhpParser... ✔
Running task 2/7: Phpcs... ✔
Running task 3/7: Phpspec... ✔
Running task 4/7: Phpunit... ✔
Running task 5/7: Composer... ✔
Running task 6/7: PhpLint... ✔
Running task 7/7: YamlLint... ✔
             ▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄
           ▄▄▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌           
         ▄▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌
        ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌
        ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌        
  ▄▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌
 ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▌        
[...]

PR #588, opened 2019-01-08
Remove deprecated warning for strpos with non-string needles

tdoedtmann:
With PHP 7.3, the second parameter of `strpos` is always interpreted
as a string.
So the specification of -5 as second parameter, as in
`ConfigurationFile::locateConfigFileWithDistSupport`, deprecated.
At this point `substr` would be more obvious anyway and will be
replaced with this function.

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Documented?   | /
| Fixed tickets | /

### DIFF
--- a/src/Locator/ConfigurationFile.php
+++ b/src/Locator/ConfigurationFile.php
@@ -72,7 +72,7 @@ class ConfigurationFile
      */
     private function locateConfigFileWithDistSupport($defaultPath)
     {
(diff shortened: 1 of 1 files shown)


veewee:
Fixed in #576

PR #591, opened 2019-01-22
Add memory-limit parameter to PhpStan

peter-osre:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | none

This adds the option for PhpStan to set a memory-limit for the analyser.

### DIFF
--- a/doc/tasks/phpstan.md
+++ b/doc/tasks/phpstan.md
@@ -20,6 +20,7 @@ parameters:
             force_patterns: []
             ignore_patterns: []
             triggered_by: ['php']
+            memory_limit: -1
 ```
 
 **autoload_file**
@@ -57,3 +58,9 @@ This is a list of patterns that will be ignored by phpstan. With this option you
 *Default: [php]*
 
 This is a list of extensions to be sniffed.
+
+**memory_limit**
+
+*Default: -1*
+
+With this parameter you can specify the memory limit.
--- a/doc/tasks/phpstan.md
+++ b/doc/tasks/phpstan.md
@@ -61,6 +61,6 @@ This is a list of extensions to be sniffed.
 
 **memory_limit**
 
-*Default: -1*
+*Default: null*
 
 With this parameter you can specify the memory limit.
--- a/src/Task/PhpStan.php
+++ b/src/Task/PhpStan.php
@@ -34,11 +34,13 @@ class PhpStan extends AbstractExternalTask
             'ignore_patterns' => [],
             'force_patterns' => [],
             'triggered_by' => ['php'],
+            'memory_limit' => '-1',
         ]);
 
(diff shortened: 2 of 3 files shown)


doc/tasks/phpstan.md
```diff
@@ -20,6 +20,7 @@ parameters:
             force_patterns: []
             ignore_patterns: []
             triggered_by: ['php']
+            memory_limit: -1
```
localheinz: Probably out of scope of this PR (since the list of options is not sorted already), but what do you think about keeping the list of options sorted by name (also applies to other occurrences in this PR)?

starred-arthur:
👍

src/Task/PhpStan.php
```diff
@@ -81,6 +83,7 @@ public function run(ContextInterface $context)
         $arguments->addOptionalArgument('--autoload-file=%s', $config['autoload_file']);
         $arguments->addOptionalArgument('--configuration=%s', $config['configuration']);
         $arguments->add(sprintf('--level=%u', $config['level']));
+        $arguments->add(sprintf('--memory-limit=%s', $config['memory_limit']));
```
Landerstraeten: I would make this an optional argument

doc/tasks/phpstan.md
veewee: Would be nice, but indeed out of scope here.

src/Task/PhpStan.php
```diff
@@ -34,11 +34,13 @@ public function getConfigurableOptions()
             'ignore_patterns' => [],
             'force_patterns' => [],
             'triggered_by' => ['php'],
+            'memory_limit' => '-1',
```
veewee: Can you set the default to null to make sure it uses the phpstan defaults?

src/Task/PhpStan.php
```diff
@@ -34,11 +34,13 @@ public function getConfigurableOptions()
...
         ]);
 
         $resolver->addAllowedTypes('autoload_file', ['null', 'string']);
         $resolver->addAllowedTypes('configuration', ['null', 'string']);
         $resolver->addAllowedTypes('level', ['int']);
+        $resolver->addAllowedTypes('memory_limit', ['string']);
```
veewee: can be string or null

doc/tasks/phpstan.md
```diff
@@ -57,3 +58,9 @@ This is a list of patterns that will be ignored by phpstan. With this option you
 *Default: [php]*
 
 This is a list of extensions to be sniffed.
+
+**memory_limit**
+
+*Default: -1*
```
veewee: default : null to fall back on defaults inside phpstan

PR #605, opened 2019-02-28
Fix issue with set global path (~/.composer/vendor/bin) for bin_dir (For PHPStorm)

iGerchak:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

This PR fix issue for PHPStorm and global install tasks. For example, phpcs installed to ~/.composer/vendor

When we try to create the commit via PHPStorm UI form we'll get the error: 'The executable for "phpcs" could not be found.'. This PR fix it.

### DIFF
--- a/src/Locator/ExternalCommand.php
+++ b/src/Locator/ExternalCommand.php
@@ -14,7 +14,7 @@ class ExternalCommand
 
     public function __construct(string $binDir, ExecutableFinder $executableFinder)
     {
-        $this->binDir = rtrim($binDir, '/\\');
+        $this->binDir = str_replace('~', $_SERVER['HOME'], rtrim($binDir, '/\\'));
         $this->executableFinder = $executableFinder;
     }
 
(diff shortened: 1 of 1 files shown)


veewee:
Hello @iGerchak,

Thanks for the PR.
We won't accept it since we want to avoid injecting `$_SERVER` variables in every possible place.
I suggest changing the location to the newly introduced `git_hook_variables` parameters.

That way you can specify whichever path you want to include inside the grumphp.yaml file.
It would look something like:
```yaml
parameters:
    git_hook_variables:
        PATH: '~/.composer/vendor/bin:$PATH'
       ## Default: PATH: '$PATH'
```

In the git hook template, you could set the path like this:

```
PATH=${PATH}
```

I assume this could work as well and will also solve issues like this one:
https://github.com/phpro/grumphp/blob/master/doc/faq.md#how-can-i-fix-the-sourcetree-path-problem


Are you up for adding this change?

PR #611, opened 2019-03-22
End 2 end tests

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   |no
| Fixed tickets | 

This PR contains a first take on adding e2e tests.
Currently it tests some basic folder structures (of which most break 😱 )
It is possible to use the same base testcase for testing the composer plugin or tasks with various yaml settings.

Currently the e2e test will run one by one direclty on CI. This is a bit slow.
Once paratest is added as grumphp task, we can run them in parallel and possible add the testsuite directly to Grumphp file.


composer.json
```diff
@@ -11,15 +11,15 @@
...
-        "symfony/config": "~3.0|~4.0",
-        "symfony/console": "~3.0|~4.0",
-        "symfony/dependency-injection": "~3.0|~4.0",
-        "symfony/event-dispatcher": "~3.0|~4.0",
-        "symfony/filesystem": "~3.0|~4.0",
-        "symfony/finder": "~3.0|~4.0",
-        "symfony/options-resolver": "~3.0|~4.0",
-        "symfony/process": "~3.2|~4.0",
-        "symfony/yaml": "~3.0|~4.0"
+        "symfony/config": "~3.4|~4.0",
```
Landerstraeten: 👍

phpunit.xml.dist
```diff
@@ -1,10 +1,13 @@
 <phpunit bootstrap="test/Bootstrap.php" colors="true">
+    <php>
+        <ini name="error_reporting" value="-1" />
+    </php>
     <testsuites>
-        <testsuite name="Linters">
-            <directory>test/Linter</directory>
+        <testsuite name="Unit">
```
Landerstraeten: 👍

veewee:
> I see some skipped tests. I suppose we will fix them later?

Yes, those are broken scenarios which should be fixed at some point in time. The idea is to add these tests to make sure we don't break stuff in the future while fixing those issues;

Issue #593, opened 2019-02-15
Phpstan task: `level` parameter always overwrites configured level in `phpstan.neon`


prilka:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.14.3`
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

The `level` parameter of the Phpstan task always(!) overwrites the configured level in my `phpstan.neon`. That is because the task always/unconditional sets the `--level` command-line option. If I omit the `level` parameter in my `grumphp.yml`, the default value (level 0) is set and my configured level (e.g. level 3) is ignored.

To fix this, I have to set the value in my `grumphp.yml` to the same value as in my `phpstan.neon`, which is annoying. 

While the default value may fit well for most situations, it's inconvenient for this case. So maybe we need some kind of marker to tell the task not to include the `--level` option for the command-line. Something like `level: null` or `level: "unset"`

**My configuration**
```yaml
parameters:
    git_dir: .
    bin_dir: vendor/bin
    tasks:
      ...
      phpstan:
        triggered_by: ['php']
```

**phpstan.neon**
```yaml
parameters:
    level: 3
    ...
```

Pazns:
I agree with this.
Unexpected behaviour, and should be documented as such.

I lost time wondering why my GrumpPHP PHPStan was not blocking things the command-line was...

PR #612, opened 2019-03-23
Improve `level` parameter of PhpStan task

prilka:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #593

This PR allows the two new values `NULL` and `"max"` for the `level` parameter of the PhpStan task. `NULL` will omit the `--level` command line option, which is useful for projects with a phpstan.neon config file.


src/Task/PhpStan.php
```diff
@@ -80,7 +85,7 @@ public function run(ContextInterface $context): TaskResultInterface
         $arguments->addOptionalArgument('--autoload-file=%s', $config['autoload_file']);
         $arguments->addOptionalArgument('--configuration=%s', $config['configuration']);
         $arguments->addOptionalArgument('--memory-limit=%s', $config['memory_limit']);
-        $arguments->add(sprintf('--level=%u', $config['level']));
+        $arguments->addOptionalIntegerArgument('--level=%s', $config['level']);
```
veewee: This will break on PHP 7.1 upgrade which will contain typehint `?int`. Can you create a `addOptionalMixedArgument()` instead?

veewee:
Thanks for the PR, I've added a small remark.

src/Task/PhpStan.php
prilka: Sure. But without type hints both methods will look exactly the same. How should we deal with that?

prilka: is something like this okay for you?
```php
# src/Collection/ProcessArgumentsCollection.php

public function addOptionalIntegerArgument(string $argument, $value)
{
    return $this->addOptionalMixedArgument($argument, $value);
}

public function addOptionalMixedArgument(string $argument, $value)
{
    if (null === $value) {
        return;
    }

    $this->add(sprintf($argument, $value));
}
```

This still works after adding the `?int` type hint

veewee: good point! The solution you've provided is good for me. That allows us to have a type check for integers in the future.

src/Collection/ProcessArgumentsCollection.php
```diff
@@ -132,6 +132,15 @@ public function addOptionalBooleanArgument(string $argument, $value, string $tru
...
     public function addOptionalIntegerArgument(string $argument, $value)
+    {
+        $this->addOptionalMixedArgument($argument, $value);
+    }
+
+    /**
+     * @param string $argument
+     * @param mixed $value
+     */
+    public function addOptionalMixedArgument(string $argument, $value)
```
Landerstraeten: @veewee I think there is no need to add this new function? Can't we just use `addOptionalArgument()`?

veewee: No it can't.
The `addOptionalArgument()` method does not add the argument on e.g. value `0`.
In this case `0` is a valid value, so we have to use a `null` check to make sure it falls back to the phsptan configuration file or phpstan default value instead.

PR #616, opened 2019-03-29
Upgrade to php 7.1

Landerstraeten:


src/Collection/ProcessArgumentsCollection.php
```diff
@@ -14,10 +14,7 @@ public static function forExecutable(string $executable): self
         return new self([$executable]);
     }
 
-    /**
-     * @param string|null $value
-     */
-    public function addOptionalArgument(string $argument, $value)
+    public function addOptionalArgument(string $argument, $value = null): void
```
SerkanYildiz:
```suggestion
    public function addOptionalArgument(string $argument, string $value = null): void
```

Landerstraeten: @SerkanYildiz $value can also be null or an integer

SerkanYildiz: Ok, according to the removed php docblock it should be an string|null, that's why it was my suggested change :)

Landerstraeten: What about `?string` and force to cast?

PR #617, opened 2019-03-29
Add Paratest task

janvernieuwe:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

A PR with only paratest, not all tasks in parallel, so we can get this merged faster

# New Task Checklist:

- [ x ] Is the README.md file updated?
- [ x ] Are the dependencies added to the composer.json suggestions?
- [ x ] Is the doc/tasks.md file updated?
- [ x ] Are the task parameters documented?
- [ x ] Is the task registered in the tasks.yml file?
- [ x ] Does the task contains phpspec tests?
- [ x ] Is the configuration having logical allowed types?
- [ x ] Does the task run in the correct context?
- [ x ] Is the `run()` method readable?
- [ x ] Is the `run()` method using the configuration correctly?
- [ x ] Are all CI services returning green?

### DIFF
--- a/doc/tasks/paratest.md
+++ b/doc/tasks/paratest.md
@@ -1,4 +1,4 @@
-# Phpunit
+# Paratest
 
 The Paratest task will run your PHPUnit tests in parallel.
 
@@ -36,10 +36,10 @@ parameters:
 *Default: null*
 
 The number of test processes to run. (Default: auto)
-                             Possible values:
-                             - Integer (>= 1): Number of processes to run.
-                             - auto (default): Number of processes is automatically set to the number of logical CPU cores.
-                             - half: Number of processes is automatically set to half the number of logical CPU cores.
+ Possible values:
+ - Integer (>= 1): Number of processes to run.
+ - auto (default): Number of processes is automatically set to the number of logical CPU cores.
+ - half: Number of processes is automatically set to half the number of logical CPU cores.
 
 **functional**
 
@@ -85,6 +85,7 @@ Runner, WrapperRunner or SqliteRunner. (Default: Runner)
 *Default: null*
 
 Debugger settings for code coverage.
+It can have a bin and a args key. These can be used to set the debug binary and additional arguments.
 
 **coverage-xml**
 
--- a/doc/tasks/paratest.md
+++ b/doc/tasks/paratest.md
@@ -1,0 +1,113 @@
+# Phpunit
+
+The Paratest task will run your PHPUnit tests in parallel.
+
+***Composer***
+
+```
+composer require --dev brianium/paratest
+```
+
+***Config***
+
+The task lives under the `paratest` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+parameters:
+    tasks:
+        paratest:
+            processes: ~
+            functional: ~
+            group: []
+            always_execute: false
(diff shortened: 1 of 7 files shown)


doc/tasks/paratest.md
```diff
@@ -0,0 +1,113 @@
+# Phpunit
```
veewee: paratest

doc/tasks/paratest.md
```diff
@@ -0,0 +1,113 @@
...
+            log-junit: null
+            testsuite: null
+```
+
+**processes**
+
+*Default: null*
+
+The number of test processes to run. (Default: auto)
+                             Possible values:
```
veewee: The indentation is a bit off. Therefor it doesn't display a list

doc/tasks/paratest.md
```diff
@@ -0,0 +1,113 @@
...
+
+*Default: null*
+
+Runner, WrapperRunner or SqliteRunner. (Default: Runner)
+
+**debugger**
+
+*Default: null*
+
+Debugger settings for code coverage.
```
veewee: Can you specify the settings here. Unclear what should be in there from the documentation.

src/Task/Paratest.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $coverageEnabled = function ($config) {
+            return !empty($config['coverage-xml'])
+                || !empty($config['coverage-html'])
+                || !empty($config['log-junit']);
+        };
+        if ($coverageEnabled($config) && !empty($config['debugger'])) {
+            $bin = $config['debugger']['bin'];
+            $args = $config['debugger']['args'];
+            $arguments = new ProcessArgumentsCollection(array_merge([$bin], $args, $arguments->toArray()));
```
veewee: don't you need to `...$args, ...$arguments->toArray()` ?

src/Task/Paratest.php
```diff
@@ -0,0 +1,104 @@
...
+        $arguments->addOptionalArgument('-c %s', $config['configuration']);
+        $arguments->addOptionalArgument('--phpunit %s', $config['phpunit']);
+        $arguments->addOptionalArgument('--runner %s', $config['runner']);
+        $arguments->addOptionalArgument('--coverage-xml %s', $config['coverage-xml']);
+        $arguments->addOptionalArgument('--coverage-html %s', $config['coverage-html']);
+        $arguments->addOptionalArgument('--log-junit %s', $config['log-junit']);
+        $arguments->addOptionalArgument('--testsuite %s', $config['testsuite']);
+        $arguments->addOptionalCommaSeparatedArgument('--group=%s', $config['group']);
+
+        $coverageEnabled = function ($config) {
```
veewee: might also be just a variable or a private function

src/Task/Paratest.php
janvernieuwe: I don't think so? The array merge should handle it to become a single array, and that is what the constructor expects.

doc/tasks/paratest.md
veewee: @janvernieuwe  can you add an example?

PR #627, opened 2019-04-19
Improve level parameter of PhpStan task

Landerstraeten:
See #612


src/Collection/ProcessArgumentsCollection.php
```diff
@@ -132,11 +132,24 @@ public function addOptionalBooleanArgument(string $argument, $value, string $tru
...
+     * @param string $argument
+     * @param mixed $value
+     */
+    public function addOptionalMixedArgument(string $argument, $value)
     {
         if (null === $value) {
             return;
         }
 
+        if (\is_bool($value)) {
```
veewee: Maybe a bridge to far at this point since it isn't used?
We also have `addOptionalBooleanArgument()` for this use case

PR #629, opened 2019-05-06
Adds PHP7 type hinting where missing

MarkVaughn:
String

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | no

Added type hinting where missing. 
TaskResultCollection: only returns int
ConfigureCommand: return is string, which is the parameter, therefore parameter should be string.
PreCommitCommand: never returns void
RunCommand: never returns void
Str: doc says string, should be string

### DIFF
--- a/src/Collection/TaskResultCollection.php
+++ b/src/Collection/TaskResultCollection.php
@@ -28,14 +28,11 @@ class TaskResultCollection extends ArrayCollection
         return false;
     }
 
-    /**
-     * @return int
-     */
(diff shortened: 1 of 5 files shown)


Landerstraeten:
Thanks @MarkVaughn! I did some minor changes to your pr. Feel free to help us out with other open issues.

PR #641, opened 2019-05-28
Optimize phpmnd

Big-Shark:
| Q             | A
| ------------- | ---
| Branch        | features 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes

Check only changes files

### DIFF
--- a/doc/tasks/phpmnd.md
+++ b/doc/tasks/phpmnd.md
@@ -17,7 +17,6 @@ The task lives under the `phpmnd` namespace and has following configurable param
 parameters:
     tasks:
         phpmnd:
-            directory: .
             whitelist_patterns: []
             exclude: []
             exclude_name: []
@@ -28,13 +27,7 @@ parameters:
             ignore_strings: []
             strings: false
             triggered_by: ['php']
-```
-
-**directory**
-
-*Default: .*
-
-With this parameter you can define which directory you want to run `phpmnd` in (must be relative to cwd).
+`
 
 **whitelist_patterns**
 
--- a/spec/Task/PhpMndSpec.php
+++ b/spec/Task/PhpMndSpec.php
@@ -41,7 +41,6 @@ class PhpMndSpec extends ObjectBehavior
     {
         $options = $this->getConfigurableOptions();
         $options->shouldBeAnInstanceOf(OptionsResolver::class);
-        $options->getDefinedOptions()->shouldContain('directory');
         $options->getDefinedOptions()->shouldContain('whitelist_patterns');
         $options->getDefinedOptions()->shouldContain('exclude');
         $options->getDefinedOptions()->shouldContain('exclude_name');
--- a/src/Task/PhpMnd.php
+++ b/src/Task/PhpMnd.php
@@ -25,7 +25,6 @@ class PhpMnd extends AbstractExternalTask
     {
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
-            'directory' => '.',
             'whitelist_patterns' => [],
             'exclude' => [],
             'exclude_name' => [],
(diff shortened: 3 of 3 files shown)


Big-Shark:
@veewee Hi, please, check this commit, phpmnd supported working with [separate files](https://github.com/povils/phpmnd/blob/master/src/PHPFinder.php#L24), and this is more quickly than check all directory every time, but this is work BC now, what do you think about it, and how we can change this?

veewee:
Hi @Big-Shark,

Thanks for the PR. I am not using this tool myself, but is it also possible to run it without files?
In that case, it seems better to let the `run` command call the cli tool without any files.
The `pre-commit` can add the files that are changed.

Otherwise the task won't be useable on e.g. windows, which has a small max cli input length.

Big-Shark:
@veewee Sorry, this is not released yet https://github.com/povils/phpmnd/issues/95 (

exussum12:
This already exists in the released version. You can use --whitelist and provide a file containing the changes. 

Personally I'm not sure of the advantage over listing the files on the command line. 

I use whitelist with output directly from git diff --names-only

Big-Shark:
@exussum12 We waite release with this [PR](https://github.com/povils/phpmnd/pull/94), and this is not released now.
Grumphp working with php git wrapper, and try don't create files for checking, we would prefer to inject file names in console command,  as do phpcsfix, phpmd, phpstan.

exussum12:
Yeah I have seen that pr. You don't need to create a file on disk you can use <(git command) to use a file descriptor. 

Phpcs also has a file ? Not sure about C's fixer but it's not worst case creating a small file which is cleaned up

veewee:
Let's wait for the release then.

@exussum12 : Can you provide me a code sample of what you mean exactly? I am not using the tool myself yet, so it's a bit confusing to me. Doesn't the files arguments make sense to you and how would you do it from grumphp's perspective?

exussum12:
Grumphp could right now make a file containing the changes files. Then add an option to the current way phpmnd is called with --whitelist file.txt and that would work

I use git diff --name-only --diff-filter=ACMR to generate the file in my ci process. I'm not sure how grumphp does it but I imagine something similar ? Either way what ever grump does to list the files if it can write to a file one changed file per line and then call --whitelist to phpmnd that will work as expected today without waiting for master to be tagged.

veewee:
Mostly we are just passing along all files directly through CLI arguments.
The problem with this is that Windows has a rather small maximum cli input length. Therefor it will skip the task when you add a lot of files.

For the `phplint` task, we are using an input stream to pass the files. We could also write the files to a file so that `phpmnd` could open that one as a stream internally. In any case : streams are the way to go for passing files so that we won't hit any maximum cli input length.

Since this depends on how the 3rd party tools work, there isn't a lot of generic tooling around that from our package **yet**.

exussum12:
streams as in STDIN ? 

You can use bash process substitution (I imagine there is a similar trick in windows) ? Where a the info is passed as a file descriptor (rather than a file)

veewee:
Maybe my wording was a bit off.
What I meant was indeed by passing a list of files through STDIN or through a file like you suggested in a post above (`--whitelist file-with-filenames.txt`).
Depending on what the 3rth pary tool supports, both are possible.

If we were to pass a file to the cli tool, it's probably easier to put it in a temporary file. This way both windows and unix based systems can handle it. Don't think process substitution will easily work on Windows.

Example of STDIN stream in grumphp:
https://github.com/phpro/grumphp/blob/master/src/Task/PhpLint.php#L65-L73

exussum12:
ah brilliant. Yes that work work with phpmnd today (without waiting for a tag) this has been supported for multiple versions so should be safe to enable

veewee:
Closing this one because of inactivity. Feel free to provide a new solution if this still applies.

PR #643, opened 2019-05-31
Only show project files when project is in subfolder

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | 
| Fixed tickets | #607, #592

This PR will fix invalid files when using grumphp.yml + composer.jshon in a single subdirectory inside git.

For example: ./project


mattheo-geoffray:
@veewee Hi, do you when the next release is planned?

veewee:
Hi @mattheo-geoffray,

Sadly I discovered some broken scenario's when testing out [conventions](https://github.com/phpro/grumphp/blob/master/doc/conventions.md).
There are still some other scenario's well want to cover before releasing, since they will break BC.
In the meantime, you can check out master and use that one to verify if everything works for you.
You can also track progress in #644, but I am afraid I won't be able to work on that one very frequently.

mattheo-geoffray:
@veewee Hi!
Oh OK no problem I will work with that one until it can be a stable release!   
Thank you for the work! 🙂

mattheo-geoffray:
Hi @veewee !
Do you have an idea for the release date of this PR?  
Thanks!

veewee:
Not yet, first have to find time to finish the pr above.

PR #644, opened 2019-05-31
Improved exotic paths

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #643

This PR aims to fix some well known issues with [exotic paths](https://github.com/phpro/grumphp/blob/master/doc/installation/exotic.md).
Previous PR (#643) introduced an issue when using [conventions](https://github.com/phpro/grumphp/blob/master/doc/conventions.md).
This PR will also try to fix issues when composer, git and grumphp config files are in separate directories.

This PR:

- Gets rid of `bin_dir` : it is loaded from composer config instead. 
- Gets rid of `git_dir`: it tries to locate the git dir on every run.
- Introduces new environment variables to change paths
- Rewrites the paths, application, commands, container and autoloading section
- Contains a lot of testcases on various project structures
- Contains a system that tries to guess where all your config files (git, composer, bin, grumphp, ...) are located. 
- Fixes some long lasting filesystem issues (symlinks, submodules, relative / absolute path issues, ...)
- Fixes issues with locating diffed / registered git files in a project dir.
- Makes it possible  to generate phars
- Works globally installed
- It will fix most of the open issues related to windows and the path system
- Makes it possible to reuse same git hook in both local + vagrant scenarios

Manually tested: vagrant, vagrant preset, composer-bin-plugin, docker

You can test it as well:

```sh
composer remove phpro/grumphp
composer config repositories.grumphp-shim vcs https://github.com/veewee/grumphp
composer require --dev phpro/grumphp:dev-improved-paths
```

Note:
bin_dir and git_dir are removed from config. But it doesn't matter if it's still in there


src/Configuration/GuessedPaths.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * @var string
+     */
+    private $defaultConfigFile;
+
+    private function __construct()
+    {
+    }
+
+    public static function guess(Filesystem $filesystem, GitDirLocator $gitDirLocator): self
```
veewee: @Landerstraeten : moved all path guessing logic to this factory. (probably gonna move the code though)
Can you go through it to see if it makes sense to you?

This one will be constructed early in the cli boot code and will be used to tell GrumPHP where to look for stuff. (git, bins, configs, ...)

src/Locator/GuessedPathsLocator.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    private $gitDirLocator;
+
+    public function __construct(Filesystem $filesystem, GitDirLocator $gitDirLocator)
+    {
+        $this->filesystem = $filesystem;
+        $this->gitDirLocator = $gitDirLocator;
+    }
+
+    public function locate(?string $cliConfigFile): GuessedPaths
```
veewee: @Landerstraeten : moved all path guessing logic to this locator.
Can you go through it to see if it makes sense to you?

This one is constructed early in the cli boot code and will be used to tell GrumPHP where to look for stuff. (git, bins, configs, ...)

resources/hooks/local/pre-commit
```diff
@@ -8,5 +8,8 @@
 # Fetch the GIT diff and format it as command input:
 DIFF=$(git -c diff.mnemonicprefix=false --no-pager diff -r -p -m -M --full-index --no-color --staged | cat)
 
+# Grumphp env vars
+export GRUMPHP_GIT_WORKING_DIR=$(git rev-parse --show-toplevel)
+
 # Run GrumPHP
 (cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | $(EXEC_GRUMPHP_COMMAND) $(HOOK_COMMAND) '--skip-success-output')
```
guncha25: If ${GRUMPHP_GIT_WORKING_DIR} would be added  in front of ${HOOK_EXEC_PATH} then grumphp would work on any level not just git root.

doc/installation/exotic.md
```diff
@@ -1,23 +1,105 @@
 # Installation with an exotic project structure
 
-When your application has a project structure that is not covered by the default configuration settings,
-you will have to create a `grumphp.yml` *before* installing the package
-and add next config into your application's `composer.json`:
+By default, GrumPHP` is trying to auto-discover a lot of possible project structures.
```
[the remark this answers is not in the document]
veewee: good point, will improve docs here.

resources/config/subscribers.yml
```diff
@@ -3,9 +3,15 @@ services:
     class: GrumPHP\Event\Subscriber\StashUnstagedChangesSubscriber
     arguments:
       - '@config'
-      - '@git.repository'
+      - '@GrumPHP\Git\GitRepository'
       - '@grumphp.io'
```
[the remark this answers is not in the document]
veewee: Currently no plans for that. it's something that is happening gradually while we are improving stuff. 
We could create a PR to fix them all, but it might break extensions.

src/Locator/GitWorkingDirLocator.php
```diff
@@ -0,0 +1,41 @@
...
+        $process = ProcessFactory::fromArguments($arguments);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            throw new RuntimeException(
+                'The git directory could not be found. Did you initialize git? ('.$process->getErrorOutput().')'
+            );
+        }
+
+        return trim($process->getOutput());
```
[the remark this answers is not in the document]
veewee: It should be. We were using it previously as well and I don't know of a case where it doesnt do what it promises.

(Actually: when you run this in a php script in a subdir of the git root during pre-commit, it crashes. This is a git bug which is very vague, so I fixed it by adding the `GRUMPHP_GIT_WORKING_DIR` in the local git commit.)

resources/hooks/local/pre-commit
veewee: Thanks for your remark.
Can you add some additional information on what you want this to do exactly?
When running pre-commit, git always takes the git-root as cwd. Or isn't this always the case?

guncha25: I have a case where git root is not the same as composer.json and grumphp.yml. In that case working directory is set to grumphp.yml directory. Testing with setup - [git root] -> [drupal (grumphp.yml and composer.json)]. After install i have lst line : 
`(cd "drupal/" && printf "%s\n" "${DIFF}" | 'exec' './vendor/bin/grumphp' 'git:pre-commit' '--skip-success-output')` That can not recognize files staged.

guncha25: Before fix project/drupal/web only project is working. After all 3 levels are working.

veewee: Tries this out:

```sh
mkdir tmp
cd tmp
git init
mkdir drupal
cd drupal
composer init
composer config repositories.grumphp-shim vcs https://github.com/veewee/grumphp
echo 'vendor' >> .gitignore
composer require --dev phpro/grumphp:dev-improved-paths
vim grumphp.yml
./vendor/bin/grumphp r
git commit -m"test"
```

The pre-commit looks like you mentioned above and everything seems to work as expected.
Can you tell me the steps on how to get the error message?

(Or is it when you run the git hook directly from the drupal directory? `../.git/hooks/pre-commit`? In that case it seems a better idea to run `./vendor/bin/grumphp git:pre-commit`, ATM we assume the git hook always runs from git root during pre-commit)

guncha25: Ok. After doing a test in clean setup - it seems to work. It might be some weird issue from debugging things. Looks like there are no need for any change. Keep up the good work.

src/Git/GitRepository.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    private $repositoryLocator;
+
+    /**
+     * @var array
+     */
+    private $repositoryOptions;
+
+    /**
+     * @var ?Repository
```
Landerstraeten: @veewee, what's the best way to typehit: `null|Repository` or `?Repository`?

veewee: `null|Repository` works best in most IDEs I guess ...

PR #646, opened 2019-06-15
Fixed a bug when phplint is active and only non-php files are commited

glamorous:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | no

There was a bug when you try to commit (pre-hook) some non-php files.
You receive the PHP Parallel Lint usage error every time. With this it
will be marked as "passed" when no files are available to pass trough
the linter.

### DIFF
--- a/src/Task/PhpLint.php
+++ b/src/Task/PhpLint.php
@@ -53,6 +53,10 @@ class PhpLint extends AbstractExternalTask
             ->notPaths($config['ignore_patterns'])
             ->extensions($config['triggered_by']);
 
+        if ($files->isEmpty()) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
         $arguments = $this->processBuilder->createArgumentsForCommand('parallel-lint');
         $arguments->add('--no-colors');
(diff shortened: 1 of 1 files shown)


src/Task/PhpLint.php
```diff
@@ -53,6 +53,10 @@ public function run(ContextInterface $context): TaskResultInterface
             ->notPaths($config['ignore_patterns'])
             ->extensions($config['triggered_by']);
 
+        if ($files->isEmpty()) {
+            return TaskResult::createPassed($this, $context);
```
veewee: Can you change this to:
```php
```

This way, it is consequent with the other tasks.

veewee:
Hi @glamorous 

Thanks for the PR! Looks good, but I added one small remark.

src/Task/PhpLint.php
glamorous: No problem! Fix is on the way!

glamorous:
I changed the commit message too, so it aligns to the new method "skip"

veewee:
Thanks for the quick iteration.

PR #648, opened 2019-06-24
Add more output when Blacklist fails with Git errors

Shafied:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

Right now when the Blacklist task fails on Git errors, you don't see the output, but your taks passes. In my opinion the task should fail with some error output. This way you dont have false positives


Landerstraeten:
Took over due to inactivity. See #693

PR #651, opened 2019-07-03
Add $files to process $arguments (Easy Coding Standard Task)

jmatthiesen81:
$files will never be assigned to the process arguments. This has the affect that the check never fail.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

This pull request adds the `$files` variable to the process `$arguments`.

### DIFF
--- a/src/Task/Ecs.php
+++ b/src/Task/Ecs.php
@@ -70,6 +70,7 @@ class Ecs extends AbstractExternalTask
         $arguments->addOptionalArgument('--no-progress-bar', $config['no-progress-bar']);
         $arguments->addOptionalArgument('--ansi', true);
         $arguments->addOptionalArgument('--no-interaction', true);
+        $arguments->addFiles($files);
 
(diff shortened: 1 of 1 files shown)


veewee:
Hi @jmatthiesen81,

Thanks for the PR. The ECS task currently works with the whitlisted paths.
We could make it variable, for example:
- on run: use whitelist
- on pre-commit: attach files.

However, adding the files might make us hit the maximum cli input length. Therefor it would be better to pass the files in a stream. At first sight, this is not possible yet in ecs?

jmatthiesen81:
Thanks for clarifying @veewee. I'm just setting up my first project using grumphp and was wondering about the successful ECS checks. Wouldn't it be recommendable to set whitelist_patterns: ['.'] as the default value in this case? Anyway using the whitelist_patterns works in my case and this pull request can be closed. :)

PR #652, opened 2019-07-05
[ECS task] Run pre commit argument switch

jmatthiesen81:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | none

This pull request extends the task with the configuration `files_on_pre_commit` to check the staged files on the pre commit instead of the files in the directories configured in `whitelist_patterns`. To keep the backward compatibility the configuration is disabled in the default.

### DIFF
--- a/src/Task/Ecs.php
+++ b/src/Task/Ecs.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace GrumPHP\Task;
 
+use GrumPHP\Collection\FilesCollection;
+use GrumPHP\Collection\ProcessArgumentsCollection;
 use GrumPHP\Runner\TaskResult;
 use GrumPHP\Runner\TaskResultInterface;
 use GrumPHP\Task\Context\ContextInterface;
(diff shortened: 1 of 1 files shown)


src/Task/Ecs.php
```diff
@@ -52,16 +54,29 @@ public function run(ContextInterface $context): TaskResultInterface
...
         $arguments = $this->processBuilder->createArgumentsForCommand('ecs');
         $arguments->add('check');
 
-        foreach ($config['whitelist_patterns'] as $whitelistPattern) {
-            $arguments->add($whitelistPattern);
+        if ($config['files_on_pre_commit'] && $context instanceof GitPreCommitContext) {
+            $arguments->addFiles($files);
+        }
+
+        if (!$config['files_on_pre_commit'] && $context instanceof RunContext) {
```
veewee: This statement doesn't seem right: in run context it should always use the whitelist instead of the files?

veewee:
Thanks for the idea!
Added this to #809 and added docs + tests!

Issue #61, opened 2015-11-17
Add Phar file


ad3n:
Add phar file for simplicity

veewee:
We discussed the phar file in issue #42.
It seems easier to use composer to install GrumPHP in your project or globally.
At the time, we didn't find any usefull reason to create a phar file. 

We would love to hear your arguments on why we should add a phar file.
Thanks!

ihsanudin:
https://github.com/vinkla/climb/pull/28

May be, that is the reason why you need to create a phar file instead of global composer package

veewee:
@ihsanudin Just discussed the topic with @aderuwe: That is indeed a valid point. 
We are currently working on some other improvements which we think will make it easier for developers to extend the application. This means we won't put our time in this phar feature at this moment.

If you want to speed up this feature, feel free to create a PR. Maybe @igormukhingmailcom has the time to help out?

adam-paterson:
I've been looking into this one this morning, we need to clean up a couple of bits before we can use Box to generate a phar for us.

**Found Issues:**
* Autoload redeclare error due to autoload loop in `bin/grumphp`
* `File "phar:///Users/adampaterson/Code/grumphp/bin/grumphp.phar/src/GrumPHP/Console/Helper/../../../../" doesn't exists.`

veewee:
We should also keep security in mind:
https://mwop.net/blog/2015-12-14-secure-phar-automation.html

kmelia:
Hi,

A Phar file is a must have for your project! Think about composer, we use it because it's simple to get it everywhere.

With the current project, I failed trying to create a Phar file, because of the many hardcoded absolute/relative paths.

veewee:
Hi @kmelia,

We did also bump into the same issues as you trying to create a phar. I don't thinkt his will be easy and we'll need to rewrite some code for this feature. We already support many use-cases: local, global, CI, ... and need to make sure that all these cases still work.
Another problem is that GrumPHP is actually a composer plugin, which makes it a bit strange to create a standalone phar.
One last problem is that the dependencies could be local, global or in a bin folder. This makes it hard to make sure that everybody is running e.g. the tests against the same version of phpunit. 
I'm starting to think that this feature will cause a lot of work / possible bugs for little gain.

gnovaro:
Basic Tutorial
https://www.sitepoint.com/packaging-your-apps-with-phar/

m1guelpf:
Updates?

veewee:
[...]

PR #661, opened 2019-08-14
GENERATE Phar

benemohamed:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | `no`
| Fixed tickets | 

GENERATE PHP Phar archives for grumphp
I hope that this helps?

# New Task Checklist:


veewee:
Thanks for your work on this @benemohamed.
I agree that we need a phar, but this is not the way to go.
Please check my comment at : https://github.com/phpro/grumphp/issues/61#issuecomment-523813458

PR #666, opened 2019-08-30
Feature/monolog 2

pimjansen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | n/a

The PR adds support for Monolog ^2.0

# New Task Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "composer-plugin-api": "~1.0",
         "doctrine/collections": "~1.2",
         "gitonomy/gitlib": "^1.0.3",
-        "monolog/monolog": "~1.16",
+        "monolog/monolog": "~1.16 || ^2.0",
         "seld/jsonlint": "~1.1",
         "symfony/config": "~3.4 || ~4.0",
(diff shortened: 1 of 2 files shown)


pimjansen:
No idea why the builds are failing actually since it seems more like an CI implementation thing?

TomPradat:
I think you should not change the indentation of the `composer.json` file.

Can someone look into the CI which seems broken indeed ?

veewee:
CI is fixed in #644, but I am waiting for a review before I can merge it in. Should have fixed it in a separate PR. The changeset is indeed rather big for one composer version constraint.

pimjansen:
@veewee i updated the file. Updated it default to 2 tab indent instead of 4 what it was.

veewee:
Thanks for the effort! Will merge it soon.

pimjansen:
Thanks 👌🏻

PR #681, opened 2019-09-20
adjust the gitignore so the bin directory gets included in the repo

pjonckiere:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes (?)
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | none

Because of the .gitignore in this project, we cannot commit the executable for grumphp to our repository. We would like to do that, but afaik there is no way to overrule that without changing the .gitignore file itself. This PR does exactly that.

Note that not committing our vendor directory is currently not an option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
-bin
-!bin/grumphp
 vendor
 composer.lock
 grumphp.yml
 .phpunit.result.cache
+


veewee:
Hello @pjonckiere,

Thanks for reporting.

Since the gitignore whitelists the grumphp exectuable, you should be able to commit it right?

I don't understand what problem you are trying to solve with this PR. Can you give me some more details about what you are trying to achieve?

pjonckiere:
Hi @veewee ,

My goal is to get the bin/grumphp file commited to our repo. 

The whitelisting behaviour is what I was expecting as well, however [the git docs](https://git-scm.com/docs/gitignore) say: 

> It is not possible to re-include a file if a parent directory of that file is excluded.

That's why I think this change is needed before we can add the file to our repo.

veewee:
Thanks @pjonckiere,

Since it's just an old relic, it makes sense to remove it.

Issue #583, opened 2019-01-03
Add passthru option to config to pass arbitrary options to tasks


paslandau:
| Q               | A
| --------------- | ---
| Version         | `0.14.3`
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | probably
| Related tickets | 

Having a "mapping" layer from `grumphp.yml` to the specific options for each tasks makes sense, because it makes the package more resilient to changes of the underlying libraries and enables a more consistent config for the grumphp user.

However, since most of the supported tools are under active devlopment, grumphp users can't make use of newly provided options that are not yet implemented. I have two concrete examples of unsupported options (though not really new):
- `phpcs` and the `--parallel` option (see https://github.com/squizlabs/PHP_CodeSniffer/blob/390bffe2265aac8184fdf86400c0d0841e65bc4b/src/Config.php#L53 )
- `phpmnd` and the `ignore-funcs` option (see https://github.com/povils/phpmnd/blob/acf9a485673d6240110ea252ca3ff201b59e14f0/src/Console/Command.php#L55 )

I resorted to 
- duplicate the existing task
- add it as a service in `grumphp.yml`
- replace the original task with the newly created one

This is actually quite a lot of work, just to get a parameter passed to the underlying task.

Proposal
==
Consider the following grumphp.yml

```yml
parameters:
  tasks:
    phpcs: 
      passthru: "--no-cache --parallel=4"
    phpmnd: 
      passthru: "--ignore-funcs=4"
```

While parsing and providing the arguments to the underlying tool, each task would simply append the value of `passthru` to the command. Of course this should not keep contributors from adding a "stable" way to provide the missing option via `grumphp.yml`, but it provides a much more convenient way for the end user and thus increases adoptability.

Implementation
==
- Since this is a rather "global" option, it should probably be implemented in the base classes under `Task/Abstract*Task`. This would still require to "touch" a lot of Tasks, though...
- The "adding" could be implememted in `ProcessArgumentsCollection` as something like `::addRawOptionString`:
```php
    /**
     * @param string $rawOption
     */
    public function addRawOptionString($rawOption)
    {
        if (!$value) {
            return;
        }

        $this->add($rawOption);
    }
```

Thoughts?

Cheers
Pascal

veewee:
We investigeted this in #716 
There is no easy or generic way to implement this:
[...]

Issue #579, opened 2019-01-02
FR: Ability to give shell task a label


Jamesking56:
| Q               | A
| --------------- | ---
| Version         | `0.14.3`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | #566

It would be great if we could add a label to the `shell` task so that we can rename it when doing `grumphp run`.

For example, say I had a specific tool that was not a task in grumphp. Instead of it showing `Running task x/x: Shell...` in the output. It would be great if it could be re-labelled, for example `Running task x/x: MyAwesomeTool....`.

veewee:
Hello @Jamesking56,

Can you give me an extended example on what you are trying to achieve?
I don't fully understand your question.

Jamesking56:
I'm running a custom tool thats not open source by using the `Shell` task in GrumPHP. In the output I get:

`Running task x/x: Shell...`

It would be great if I could set the label, so that the output was:

`Running task x/x: MyTool...`

veewee:
Fixed in #711

Issue #509, opened 2018-05-20
Take into account tasks declaration order


vyshkant:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.14.0`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no

When building a list of tasks, I propose to consider as the initial order the one in which they are declared in `grumphp.yml` (and than apply the priority sorting to this order).

Ideally I would like to see this feature by default, but if for some reasons it is unacceptable, I'd like to have a parameter like `respect_tasks_declaration_order` in `grumphp.yml`, which will enable the described behavior.

If you agree with my idea I'll gladly write a PR.

Landerstraeten:
I don't understand your problem.

For what reason can't you use the metadata tag?

Example:
``` yaml
parameters:
  bin_dir: "./vendor/bin"
  git_dir: "."
  process_timeout: 120
  tasks:
      phpcs:
        standard: PSR2
        ignore_patterns:
          - "spec/*Spec.php"
          - "test/*.php"
        metadata:
          priority: 300
      phpspec:
        format: progress
        verbose: true
        metadata:
          priority: 200
      phpunit:
        metadata:
          priority: 100
      composer:
        no_check_lock: true
      yamllint:
        ignore_patterns:
          - "#test/(.*).yml#"
      phplint: ~
      phpparser:
        visitors:
          no_exit_statements: ~
          never_use_else: ~
          forbidden_function_calls:
            blacklist: [var_dump]
        metadata:
          priority: 100000
```

Result:
Running task 1/7: PhpParser
Running task 2/7: Phpcs
Running task 3/7: Phpspec
Running task 4/7: Phpunit
Running task 5/7: Composer
Running task 6/7: PhpLint
Running task 7/7: YamlLint

vyshkant:
> For what reason can't you use the metadata tag?

Yes I can use the `metadata` tag. But for now the default order (when not using the `metadata` tag) is completely arbitrary from the user's point of view (the user does know anything about this order). And if tasks declaration order is taken into account, then the resulting order will be clear to the user.

veewee:
Hi @vyshkant, 
It is simple to add, I'd accept a PR that adds this. Be sure to add tests and updated docs.

veewee:
Fixed in #711

PR #682, opened 2019-09-27
[WIP] Rework task system

veewee:
- [x] Lazy loaded : no more exceptions during boot phase
- [x] Name injected in task
- [x] Config injected in task
- [x] static options resolvers
- [ ] asked features :
  - [x] #579 Report task label instead of name
  - [x] #509 Preserve order of task declaration in grumphp file (when priority is equal) 
  - [ ] #583 passthru option for CLI commands - not sure we should do this (not all tasks are passtru-able - how to determine??) 
  - [x] #501 Run task with multiple configs: 
- [ ] Gracefull error message for old task configurations provided by extensions. (Can be determined by tag properties)
- [x] Better tasks testing system


veewee:
Moved to new branch on phpro/grumphp. Will be merged in #711.
Let's continue over there

PR #683, opened 2019-09-28
Fix: do not display graphiz when argument is set false

jerowork:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | none

## Description

By default, `addOptionalArgument` in `ProcessArgumentsCollection` discards an argument when the value evaluates to false.
This results in the absence of the option in the final process command.

In many cases, the absence of an option means false, but in the case of Deptrac's `--formatter-graphviz-display` it means true.

The use of `addOptionalArgument` will never allow false, so unsetting this option in Deptrac is impossible. 
In `Deptrac.php` all options are always present, so adding it just by default with `add` will fix this.

**Edit:**
Because the default value for this option `--formatter-graphviz-display` is true at Deptrac itself, the default here should be true too, to align. This means a small BC because the default is different.

## Sources
Deptrac: option `formatter_graphviz_display` is set true by default:
- https://github.com/sensiolabs-de/deptrac#graphviz-formatter

GrumPHP: option `formatter_graphviz_display` is set false by default:
- https://github.com/phpro/grumphp/blob/master/doc/tasks/deptrac.md#deptrac

### DIFF
--- a/doc/tasks/deptrac.md
+++ b/doc/tasks/deptrac.md
@@ -34,7 +34,7 @@ Set to `true` to enable the graphviz formatter.
 
 **formatter_graphviz_display**
 
-*Default: false*
+*Default: true*
 
 Open the generated graphviz image. Set to `true` to activate.
 
--- a/src/Task/Deptrac.php
+++ b/src/Task/Deptrac.php
@@ -27,7 +27,7 @@ class Deptrac extends AbstractExternalTask
         $resolver->setDefaults([
             'depfile' => null,
             'formatter_graphviz' => false,
-            'formatter_graphviz_display' => false,
+            'formatter_graphviz_display' => true,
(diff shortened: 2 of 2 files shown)


veewee:
Thanks for the detailed PR @jerowork!
The code looks good.

However : I'dd prefer to have the defaults similar to the once of the source tool as well.
Would this be something you could change + document as well?

jerowork:
@veewee I'd thought of it, but that would change the current default behavior of this parameter when not set in the config.

I am in favour of having the same default values as the source tool, so I will update this PR to align with Deptrac's defaults 👍

veewee:
Thanks for your work! Looks good!

PR #657, opened 2019-07-27
TwigCs - allocine/twigcs is abandoned. Changed to friendsoftwig/twigcs

truls1502:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

Updated TwigCs from allocine/twigcs to friendsoftwig/twigcs when [allocine/twigcs is abandoned and no longer maintained](https://packagist.org/packages/allocine/twigcs).

# New Task Checklist:

- [ ] Is the README.md file updated?
- [ ] Are the dependencies added to the composer.json suggestions?
- [ ] Is the doc/tasks.md file updated?
- [ ] Are the task parameters documented?
- [ ] Is the task registered in the tasks.yml file?
- [ ] Does the task contains phpspec tests?
- [ ] Is the configuration having logical allowed types?
- [ ] Does the task run in the correct context?
- [ ] Is the `run()` method readable?
- [ ] Is the `run()` method using the configuration correctly?
- [ ] Are all CI services returning green?

PR #686, opened 2019-10-10
twigcs exclude subfolders

oallain:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | https://github.com/phpro/grumphp/pull/657#issuecomment-539366301

Add exclude option on TwigCs after add it in https://github.com/friendsoftwig/twigcs/pull/81

I think we have to wait for the next Tag on friendsoftwig/twigcs before we merge this PR, because the feature is only on Master branch currently :wink:  (See https://github.com/friendsoftwig/twigcs/issues/82)

### DIFF
--- a/src/Task/TwigCs.php
+++ b/src/Task/TwigCs.php
@@ -21,7 +21,7 @@ class TwigCs extends AbstractExternalTask
             'severity' => 'warning',
             'ruleset' => 'FriendsOfTwig\Twigcs\Ruleset\Official',
             'triggered_by' => ['twig'],
-            'exclude' => [''],
+            'exclude' => [],
         ]);
 
         $resolver->addAllowedTypes('path', ['string']);
@@ -54,11 +54,9 @@ class TwigCs extends AbstractExternalTask
         $arguments->addOptionalArgument('--ruleset=%s', $config['ruleset']);
         $arguments->addOptionalArgument('--ansi', true);
 
-        if (\is_array($config['exclude'])) {
-            foreach ($config['exclude'] as $exclude) {
-                $arguments->addOptionalArgument('--exclude=%s', $exclude);
-            }
-        }
+        // removes all NULL, FALSE and Empty Strings
+        $exclude = array_filter($config['exclude'], 'strlen');
+        $arguments->addArgumentArray('--exclude=%s', $exclude);
 
         $process = $this->processBuilder->buildProcess($arguments);
         $process->run();
--- a/src/Task/TwigCs.php
+++ b/src/Task/TwigCs.php
@@ -21,9 +21,11 @@ class TwigCs extends AbstractExternalTask
             'severity' => 'warning',
             'ruleset' => 'FriendsOfTwig\Twigcs\Ruleset\Official',
             'triggered_by' => ['twig'],
+            'exclude' => [''],
         ]);
 
         $resolver->addAllowedTypes('path', ['string']);
+        $resolver->addAllowedTypes('exclude', ['array']);
         $resolver->addAllowedTypes('severity', ['string']);
         $resolver->addAllowedTypes('ruleset', ['string']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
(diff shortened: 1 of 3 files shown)


src/Task/TwigCs.php
```diff
@@ -57,6 +59,12 @@ public function run(ContextInterface $context): TaskResultInterface
         $arguments->addOptionalArgument('--ruleset=%s', $config['ruleset']);
         $arguments->addOptionalArgument('--ansi', true);
 
+        if (\is_array($config['exclude'])) {
```
veewee: This will always be an array since it is enforced by the options resolver

src/Task/TwigCs.php
```diff
@@ -57,6 +59,12 @@ public function run(ContextInterface $context): TaskResultInterface
...
+            foreach ($config['exclude'] as $exclude) {
```
veewee: You can skip the foreach and use the `\GrumPHP\Collection\ProcessArgumentsCollection::addArgumentArray()` method instead

src/Task/TwigCs.php
```diff
@@ -26,9 +26,11 @@ public function getConfigurableOptions(): OptionsResolver
             'severity' => 'warning',
             'ruleset' => 'FriendsOfTwig\Twigcs\Ruleset\Official',
             'triggered_by' => ['twig'],
+            'exclude' => [''],
```
veewee: don't add an empty value : just make the array empty instead

veewee:
@oallain  Thanks for the PR!
We'll wait untill it is released. In the meantime I've added some small improvements on the code.

src/Task/TwigCs.php
oallain: I can have empty values in my array, so can i use addSeparatedArgumentArray() instead of addArgumentArray() ? 

Like :
```php
$arguments->addSeparatedArgumentArray('--exclude', $config['exclude']);
```

src/Task/TwigCs.php
oallain: Yes, of course :smile:

src/Task/TwigCs.php
oallain: Ok

src/Task/TwigCs.php
veewee: Why could you have empty values? Don't you always want to specify locations to exclude?
You can indeed also use `addSeparatedArgumentArray` if the tool allows you.

If you want to remove empty values, you could wrap it with `array_filter($config['exclude'])` as well?

src/Task/TwigCs.php
```diff
@@ -57,6 +59,12 @@ public function run(ContextInterface $context): TaskResultInterface
         $arguments->addOptionalArgument('--ruleset=%s', $config['ruleset']);
         $arguments->addOptionalArgument('--ansi', true);
 
+        $arguments->addSeparatedArgumentArray('--exclude', $config['exclude']);
+
+        foreach ($config['exclude'] as $exclude) {
```
veewee: Currently you are adding the exclude items twice. is that for demonstrating your comment or some forgotten statements?

oallain: Sorry, it was atest.

src/Task/TwigCs.php
oallain: Yes, job done @veewee :smile:

truls1502:
Will a beta version https://github.com/friendsoftwig/twigcs/issues/82#issuecomment-544238857 be okay or do need to use a stable version? Since I have used patch until now and it works fine.

oallain:
@truls1502 thanks for feedback :)

---

I rebase this PR and update it with new Task tests with PHPUnit.
It works fine, but i have an error on AppVeyor, which i don't understand (and not on my code changes) :

> 1) GrumPHPTest\Unit\Linter\Xml\XmlLinterTest::it_should_validate_xml_with_dtd with data set #5 ('dtd-url-valid.xml', 0, true)
> DOMDocument::load(): php_network_getaddresses: getaddrinfo failed: No such host is known. 

Anyone know why?

test/Unit/Task/TwigCsTest.php
```diff
@@ -116,5 +117,21 @@ public function provideExternalTaskRuns(): iterable
                 '--ansi',
             ]
         ];
+
+        yield 'exclude' => [
+            [
+                'exclude' => ['src/', 'test/'],
```
veewee:
```suggestion
                'exclude' => ['src/', 'test/', null, '', false],
```

veewee:
Thanks for rebasing! Looks like we can merge it in and still keep BC to the older version of the tool, so good to go IMO.

PR #692, opened 2019-10-24
Swap implode function parameters to fix php 7.4 deprecation

patpilus:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Documented?   | -
| Fixed tickets | -

PHP 7.4 emits a deprecation warning when calling implode in reverse order from the documented order of arguments eg. implode($pieces, $glue): 
https://wiki.php.net/rfc/deprecations_php_7_4#implode_parameter_order_mix

For historical reasons, the implode() function supports passing the $glue and $pieces parameters in either order, so this change will not create any BC break:
https://www.php.net/manual/en/function.implode.php


veewee:
Thanks!

PR #694, opened 2019-10-26
Rework realpath logic to remove the possibility of a type error

BackEndTea:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | N/a
| Fixed tickets | N/a

We check that the file exists, but if for whatever reason we cant
make a realpath out of it, or if we have another process that deletes
this in the mean time, then we could have a type error.

By retrieving the real path, and then checking the error condition
we  remove the ability for a type error to occur here.

We dont change the fact that the file in question can still be
deleted in the mean time, as thats not within the scope of this method.

### DIFF
--- a/src/Util/Filesystem.php
+++ b/src/Util/Filesystem.php
@@ -33,11 +33,12 @@ class Filesystem extends SymfonyFilesystem
 
     public function realpath(string $path): string
     {
-        if (!$this->exists($path)) {
+        $realPath = realpath($path);
+        if ($realPath === false) {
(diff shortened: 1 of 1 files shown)


BackEndTea:
I don't think the CI errors are due to my PR, but i could be mistaken there

veewee:
Looks good. Thanks!
Just out of curiosity: did you really encouter an error or is it more like a theoretical thing?

BackEndTea:
This is more of a theoretical thing.

Issue #696, opened 2019-11-07
Can multiple directories be passed to phpcpd task?


steigenTI:
I'm using version `0.14.3` of grumphp.

I'm wondering if it's possible to pass multiple directories to the phpcpd task.

My task is configured as follows:
```
tasks:
  phpcpd:
    directory: src/foo src/bar
    triggered_by:
      - php
```

But when I run grumphp, I get the following result: 
```
The "src/foo src/bar" directory does not exist.
```

When I execute the following command via the command line, it correctly checks both directories:
```
vendor/sebastian/phpcpd/phpcpd src/foo src/bar
```

veewee:
Fixed in #697

PR #697, opened 2019-11-07
Pass array of directories to phpcpd

steigenTI:
| Q             | A
| ------------- | ---
| Branch        | `support-phpcs-multiple-dirs`
| Bug fix?      | yes/**no**
| New feature?  | **yes**/no
| BC breaks?    | yes/**no**
| Deprecations? | yes/**no**
| Documented?   | **yes**/no
| Fixed tickets | #696

Phpcpd allows users to pass multiple directories to the command. The tool will then check for duplicated code in those directories.

Grumphp doesn't take advantage of this - instead, users can only pass a single string for the directory.

This PR changes the string into an array, and updates the processing and docs accordingly.

### DIFF
--- a/doc/tasks/phpcpd.md
+++ b/doc/tasks/phpcpd.md
@@ -17,7 +17,7 @@ The task lives under the `phpcpd` namespace and has following configurable param
 parameters:
     tasks:
         phpcpd:
-            directory: '.'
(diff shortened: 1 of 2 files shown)


veewee:
Looks good. Thanks for the PR!

PR #709, opened 2019-12-11
Add github actions

janvernieuwe:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | -

An initial setup to use github actions, they are alot faster

# New Task Checklist:

...
-> we'll see


.github/workflows/grumphp.yml
```diff
@@ -0,0 +1,33 @@
...
+              run: php -v
+            - name: Validate composer.json and composer.lock
+              run: composer validate
+            - name: Install dependencies
+              run: composer update --prefer-dist --no-progress --no-suggest ${{ matrix.composer-options }}
+            - name: Run non windows test suite
+              run: vendor/bin/grumphp run
+              if: matrix.operating-system != 'windows-latest'
+            - name: Run windows test suite
+              run: vendor/bin/grumphp run --tasks phpunit,paratest,phpspec
```
veewee: Currently the complete testsuite is being validated on windows as well through appvoyer.
I would prefer to have all tasks tested. Why are you limiting it here?

veewee:
Thanks for the PR! Looking forward to get this one in and the 2 others CIs out :D

.github/workflows/grumphp.yml
janvernieuwe: Think i just replicated what there was.

veewee: No, we are running everything on windows. See : https://github.com/phpro/grumphp/blob/master/appveyor.yml#L74

janvernieuwe:
It took some trial and error, but I got m working now.
All tests on all envs.

https://github.com/janvernieuwe/grumphp/runs/427680393

veewee:
Gonna close this one since the configuration is not usable.
I tried some other things as well, but didn't get it to work the way I wanted it to...

For reference:
https://github.com/veewee/grumphp/blob/ae5303ee138bb44470123e033269f2d4aec04248/.github/workflows/grumphp.yml#L1-L47

This config contains following matrix, which is closer to the travis matrix we have right now:

```
            matrix:
                operating-system: [ubuntu-latest, macos-latest, windows-latest]
                php-versions: ['7.3', '7.4']
                composer-options: ['', '--prefer-lowest']
                composer-versions: ['composer:v1', 'composer:v2']
```


Issues with that configuration:

- Windows does not work
- We are not validating nighlty builds (e.g. Unreleased PHP 8 version)


(Also tried debugging why windows failed, without any success... Set-up this POC : https://github.com/veewee/poc-amp-process)


If someone with more github actions reads this : feel free to suggest some changes.
For now, we'll be using travis and appveyor - which work quite well for this repo.

PR #711, opened 2019-12-20
Improved tasks system

veewee:
- [x] Lazy loaded : no more exceptions during boot phase
- [x] Name injected in task
- [x] Config injected in task
- [x] static options resolvers
- [x] asked features :
  - [x] #579 Report task label instead of name
  - [x] #509 Preserve order of task declaration in grumphp file (when priority is equal) 
  - [x] ~~#583 passthru option for CLI commands~~ :  Won't do in this PR.
  - [x] #501 Run task with multiple configs: 
- [x] Gracefull error message for old task configurations provided by extensions. (Can be determined by tag properties)
- [x] Better tasks testing system
  - [x] Add new tests for all tasks
  - [x] Change tests for new task system
- [x] Support lowest dependencies
- [x] Update documentation
- [x] Update github PR template


Landerstraeten:
Wow! Nice work @veewee

PR #712, opened 2019-12-20
Rewrite filesize task unit test

Landerstraeten:


test/Unit/Task/FileSizeTest.php
```diff
@@ -0,0 +1,129 @@
...
+        yield 'other' => [
+            false,
+            $this->mockContext()
+        ];
+    }
+
+    public function provideFailsOnStuff(): iterable
+    {
+        /** @var ContextInterface|ObjectProphecy $context */
+        $context = $this->prophesize(RunContext::class);
```
veewee: I'dd create a specialized mockContext method for this class in this case.
You probably want to mock the file size instead of the collection.
That way you are more flexible in how it works and you will actually validate the task.

test/Unit/Task/FileSizeTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $files = new FilesCollection(array_map(
+            static function (string $file): SplFileInfo {
+                return new SplFileInfo($file, $file, $file);
+            },
+            ['file1.php', 'file2.php']
+        ));
+
+        /** @var FilesCollection $filesCollection */
+        $filesCollection = $this->prophesize(FilesCollection::class);
```
veewee: Don't mock collections

test/Unit/Task/FileSizeTest.php
```diff
@@ -0,0 +1,129 @@
...
+use GrumPHP\Collection\FilesCollection;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use GrumPHP\Task\FileSize;
+use GrumPHP\Task\TaskInterface;
+use Prophecy\Prophecy\ObjectProphecy;
+use Symfony\Component\Finder\SplFileInfo;
+
+class FileSizeTest extends AbstractTaskTestCase
```
veewee: FYI : moved the `AbstractTaskTestCase` into src so that it can be used by extensions.

Issue #719, opened 2019-12-30
Task Psalm - output format error


oallain:
| Q               | A
| --------------- | ---
| Version         | 0.17.1 (tested, but probably before)
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

**My configuration**
```yaml
# grumphp.yml
parameters:
    tasks:
        [...]
        psalm: ~
```

**Steps to reproduce:**
```sh

# Run GrumpHP:
./vendor/bin/grumphp run --tasks=psalm -vvv
```

**Result:**
```
Running task 1/1: Psalm... 
Command: '/home/oallain/public_html/SchedulerCommandPlugin/vendor/bin/psalm' '--show-info=false'
✘

In OutputFormatterStyleStack.php line 76:
                                                                  
  [Symfony\Component\Console\Exception\InvalidArgumentException]  
  Incorrectly nested style tag found.


Exception trace:
  at /home/oallain/public_html/SchedulerCommandPlugin/vendor/symfony/console/Formatter/OutputFormatterStyleStack.php:76
 Symfony\Component\Console\Formatter\OutputFormatterStyleStack->pop() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/symfony/console/Formatter/OutputFormatter.php:169
 Symfony\Component\Console\Formatter\OutputFormatter->formatAndWrap() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/symfony/console/Formatter/OutputFormatter.php:129
 Symfony\Component\Console\Formatter\OutputFormatter->format() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/symfony/console/Output/Output.php:157
 Symfony\Component\Console\Output\Output->write() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/symfony/console/Output/Output.php:132
 Symfony\Component\Console\Output\Output->writeln() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/phpro/grumphp/src/Console/Helper/TaskRunnerHelper.php:106
 GrumPHP\Console\Helper\TaskRunnerHelper->returnErrorMessages() at /home/oallain/public_html/SchedulerCommandPlugin/vendor/phpro/grumphp/src/Console/Helper/TaskRunnerHelper.php:75
[...]

PR #720, opened 2019-12-30
Psalm add output-format argument

oallain:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #719 

Add `output-format` argument to Psalm Task.

### DIFF
--- a/src/Task/Psalm.php
+++ b/src/Task/Psalm.php
@@ -26,6 +26,7 @@ class Psalm extends AbstractExternalTask
             'ignore_patterns' => [],
             'no_cache' => false,
             'report' => null,
+            'output_format' => 'compact',
             'threads' => null,
             'triggered_by' => ['php'],
             'show_info' => false,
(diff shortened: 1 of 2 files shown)


src/Task/Psalm.php
```diff
@@ -26,6 +26,7 @@ public function getConfigurableOptions(): OptionsResolver
             'ignore_patterns' => [],
             'no_cache' => false,
             'report' => null,
+            'output_format' => 'compact',
```
veewee: Can you set the default value to null? That way it uses the default 'console' output from psalm, but makes it possible to overwrite in your local grumphp.yaml config.

https://github.com/vimeo/psalm/blob/master/src/psalm.php#L211-L213

veewee:
Thanks for the PR! Added a small remark.

oallain:
Thanks @veewee

Issue #656, opened 2019-07-22
git commit -v fails because of line length


vHanda:
Git has a verbose commit option that shows you the diff in the commit editor. The diff is not a comment, but doesn't get included in the final commit. This diff's line length should not be matched by the git commit message line length rule.

$ echo 'A line which is really really really really really really really really really really really really really really long' > ./test
$ git add ./test
$ git commit -v

grumphp will fail.

Fails with version 0.15.2

PR #722, opened 2020-01-02
fix: should not fail on long body when using git --verbose

mblarsen:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #656 

This commit looks for a known marker in the git commit message when using `git --verbose` when the marker is seen it will ignore all lines after that (it will assume them to be comments).

The specs will test that long commit messages still fails above the marker.

# New Task Checklist:

### DIFF
--- a/src/Task/Git/CommitMessage.php
+++ b/src/Task/Git/CommitMessage.php
@@ -294,9 +294,14 @@ class CommitMessage implements TaskInterface
     private function getCommitMessageLinesWithoutComments(string $commitMessage): array
     {
         $lines = preg_split('/\R/u', $commitMessage);
+        $everything_below_will_be_ignored = false;
 
-        return array_values(array_filter($lines, function ($line) {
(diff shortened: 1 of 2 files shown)


src/Task/Git/CommitMessage.php
```diff
@@ -294,9 +294,14 @@ private function subjectIsSingleLined(ContextInterface $context): bool
     private function getCommitMessageLinesWithoutComments(string $commitMessage): array
     {
         $lines = preg_split('/\R/u', $commitMessage);
+        $everything_below_will_be_ignored = false;
```
oallain: Can you use Camel Case ?

mblarsen: Sure, committed the style change f732427

veewee:
Thanks for the PR!

PR #727, opened 2020-01-16
Add a footnote on using finder config in intersection mode

Taluu:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

Add a footnote that warns about setting the `config_contains_finder` to `true` (or leaving the default which is `true`) when using php-cs-fixer2 task on a lot of files. 

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4745 : on a docker+osx environment, leaving the intersection with the list of files given by grumphp can go into a timeout, if configured (otherwise it runs for a long time, gave up after 2min).

We didn't have this problem because we were making our own php_cs config for grumphp, which had a finder... but with only the result of the `git diff --staged` command (and a few pipes and greps). Removing this check from the config showed me something was wrong.

We should even probably thing of setting this to `false` by default, but that would probably be a bc break (not sure how to handle a real intersection)

### DIFF
--- a/doc/tasks/phpcsfixer2.md
+++ b/doc/tasks/phpcsfixer2.md
@@ -103,8 +103,11 @@ Cache is supported only for tool downloaded as phar file or installed via compos
 
 Intersection mode can only be used when you have a configuration file which contains a Finder.
 This mode works best since only files that are being commit and are in your configuration will be checked.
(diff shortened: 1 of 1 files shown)


veewee:
Thanks!

PR #739, opened 2020-02-12
Added support for phar executables

BrentRobert:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

Not every project allows to install psalm as a dependency.
I could not run the psalm task as I had installed the phar version.
I added support for the phar version of psalm in this PR by adding a new config option.

### DIFF
--- a/src/Locator/ExternalCommand.php
+++ b/src/Locator/ExternalCommand.php
@@ -38,23 +38,18 @@ class ExternalCommand
 
     public function locate(string $command): string
     {
-        $executable = false;
         foreach ($this->suffixes as $suffix) {
             $cmdName = $command . $suffix;
             // Search executable:
             $executable = $this->executableFinder->find($cmdName, null, [$this->binDir]);
 
             if ($executable) {
-                break;
+                return $executable;
             }
         }
 
-        if (!$executable) {
-            throw new RuntimeException(
-                sprintf('The executable for "%s" could not be found.', $command)
-            );
-        }
-
-        return $executable;
+        throw new RuntimeException(
+            sprintf('The executable for "%s" could not be found.', $command)
+        );
     }
 }
--- a/src/Locator/ExternalCommand.php
+++ b/src/Locator/ExternalCommand.php
@@ -10,6 +10,8 @@ use Symfony\Component\Process\ExecutableFinder;
 
 class ExternalCommand
 {
+    private $suffixes = ['', '.phar'];
+
     /**
      * @var string
      */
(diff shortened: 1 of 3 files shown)


veewee:
Thanks @BrentRobert  for the PR.
Doesn't GrumPHP detect phar's by default?
See config:

```yml
    executable_finder:
        class: Symfony\Component\Process\ExecutableFinder
        configurator: ['@GrumPHP\Configuration\Configurator\ExecutableFinderConfigurator', 'configure']
        calls:
            - { method: 'setSuffixes', arguments: [['.phar', '.exe', '.bat', '.cmd', '.com']] }
```

Will test it out soon.
I want to avoid adding a "phar" option for every task. So if a more generic solution works, that would be nice.

BrentRobert:
Hi Toon, I noticed the following things:

- The ExecutableFinder does not resolve the executable with the given suffixes, it only does this on Windows environments?
```
     $suffixes = [''];
        if ('\\' === \DIRECTORY_SEPARATOR) {
            $pathExt = getenv('PATHEXT');
            $suffixes = array_merge($pathExt ? explode(PATH_SEPARATOR, $pathExt) : $this->suffixes, $suffixes);
        }
```

- The second thing I noticed is the 'bin_dir' configuration option in the grumphp.yml seems to unused and ignored? Is the new way to set this using a server variable?

Thanks for the help :)!

veewee:
Hi @BrentRobert,

Never knew it was a windows only feature, but it might make sense.
So in this case it might be an option to add a thin layer in `\GrumPHP\Locator\ExternalCommand` that tries to detect the executable and tries to fallback to a phar file instead. Would that make sense?

Indeed : we removed the `bin_dir` and `git_dir` params from the config in 0.17.
We try to auto detect all of these settings. However, if GrumPHP is not able to auto detect the correct configuration, you can adjust it by using composer settings or env variables.
(https://github.com/phpro/grumphp/blob/master/doc/installation/exotic.md)

src/Locator/ExternalCommand.php
```diff
@@ -36,8 +38,17 @@ public static function loadWithPaths(Paths $paths, ExecutableFinder $executableF
...
     {
-        // Search executable:
-        $executable = $this->executableFinder->find($command, null, [$this->binDir]);
+        $executable = false;
+        foreach ($this->suffixes as $suffix) {
+            $cmdName = $command . $suffix;
+            // Search executable:
+            $executable = $this->executableFinder->find($cmdName, null, [$this->binDir]);
+
+            if ($executable) {
```
veewee: Instead of break, you could return it right away.
That makes the code less confusing and easier to read.
In that case, you can drop the executable assignment above the loop and throw the exception underneath the loop without the if statement.

src/Locator/ExternalCommand.php
```diff
@@ -10,6 +10,8 @@
 
 class ExternalCommand
 {
+    private $suffixes = ['', '.phar'];
```
veewee: This means we can drop the phar configure option for the symfony executable finder.
This list could be configurable. Don't know if that would add any value ATM though.

BrentRobert: We don't necessarily need to drop it. As for windows the executable would be found an iteration faster.

veewee: No, you are right. it's ok

veewee:
Thanks!

PR #743, opened 2020-02-21
fixes Phpcs task ignoring 'triggered_by' option when crafting the act…

giuspe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

the `phpcs` task allows to pass a list of file extension to consider (`triggered_by`);
although this parameter is considered in the `vendor/phpro/grumphp/src/Task/Phpcs.php` task class to build the file list (to check if *actually* have some file to process, and exit otherwise) the parameter is **not** eventually used in the method building the actual phpcs command (would translate into the `--extensions=` parameter). 

this patch simply remedies that with an extra line in `vendor/phpro/grumphp/src/Task/Phpcs.php@addArgumentsFromConfig`

```
$arguments->addOptionalCommaSeparatedArgument('--extensions=%s', $config['triggered_by']);
```

### DIFF
--- a/src/Task/Phpcs.php
+++ b/src/Task/Phpcs.php
@@ -120,6 +120,7 @@ class Phpcs extends AbstractExternalTask
         $arguments->addOptionalCommaSeparatedArgument('--sniffs=%s', $config['sniffs']);
         $arguments->addOptionalCommaSeparatedArgument('--ignore=%s', $config['ignore_patterns']);
         $arguments->addOptionalCommaSeparatedArgument('--exclude=%s', $config['exclude']);
+        $arguments->addOptionalCommaSeparatedArgument('--extensions=%s', $config['triggered_by']);
 
         return $arguments;
     }


giuspe:
hi there, PR submitter here.

my bad, but I don't understand why CI checks fail: something I can fix in the branch?

(thanks)

oallain:
Hi @giuspe 

You need to update unit tests in file https://github.com/phpro/grumphp/blob/master/test/Unit/Task/PhpcsTest.php
For example insert '--extensions=php' in line https://github.com/phpro/grumphp/blob/master/test/Unit/Task/PhpcsTest.php#L143 and https://github.com/phpro/grumphp/blob/master/test/Unit/Task/PhpcsTest.php#L157 
You can see it at https://travis-ci.org/phpro/grumphp/jobs/653445594?utm_medium=notification&utm_source=github_status

Thanks !

giuspe:
hi @oallain,

thanks! will fix and update the PR.

giuspe:
turns out the issue was fixed in v0.17.2 on 29/12/2019, I was still using the bugged v0.16.2.

removing the PR.

PR #745, opened 2020-03-12
Fixed exception message in constructor

MarkVaughn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

If you add phpparser to your tasks but forget to require php-parser in composer the exception thrown in the constructer hits an undefined getName() method. This could be added in the interface or in this abstract class, I chose the later.

**constructor code**
```

    public function __construct(ParserInterface $parser)
    {
        $this->configuration = new EmptyTaskConfig();
        $this->parser = $parser;

        if (!$parser->isInstalled()) {
            throw new RuntimeException(
                sprintf('The %s can\'t run on your system. Please install all dependencies.', $this->getName())
            );
        }
    }
````

### DIFF
--- a/src/Task/AbstractLinterTask.php
+++ b/src/Task/AbstractLinterTask.php
@@ -64,7 +64,10 @@ abstract class AbstractLinterTask implements TaskInterface
     {
         if (!$this->linter->isInstalled()) {
             throw new RuntimeException(
-                sprintf('The %s can\'t run on your system. Please install all dependencies.', $this->getName())
+                sprintf(
+                    'The %s can\'t run on your system. Please install all dependencies.',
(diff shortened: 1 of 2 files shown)


veewee:
Thanks for reporting! We recently moved the `getName()` method into the `config` property. Can you fetch it from there?

MarkVaughn:
Hope that is what you wanted, I do not think the failures have to do with my PR.

veewee:
Sorry for the late response, was on vacation. Will pick it up somewhere this week.
The build failure is due to a change in the latest version of `symfony/process`. (Will need to check this out first)

PR #765, opened 2020-05-08
Psalm phar support

mwolff-fn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | n/a

When using the phar package of Psalm as advertised in the task docs:
`composer require --dev psalm/phar`
...then the Psalm binary will be installed as "psalm.phar". Unfortunately, when executing the external command, the Psalm task would always use the binary name without the .phar extension, resulting in the error message `The executable for "psalm" could not be found.`on the console.

This PR introduces a new configuration option "phar" to the Psalm task that will use "psalm.phar" as the binary name when set to true.

Tests and docs have been updated to reflect the new option. There should be no side-effects.

I ticket "bug fix" instead of feature because the possibility of using the phar package was advertised in the task docs but does not work out-of-the-box when following the installation instructions.

### DIFF
--- a/doc/tasks/psalm.md
+++ b/doc/tasks/psalm.md
@@ -22,11 +22,12 @@ parameters:
             config: psalm.xml
             ignore_patterns: []
             no_cache: false
-            report: ~ 
+            report: ~
             output_format: null
             threads: 1
             triggered_by: ['php']
             show_info: false
+            phar: false
 ```
 
 
@@ -57,7 +58,7 @@ With this parameter you can run Psalm without using the cache file.
 
 *Default: null*
 
-With this path you can specify the path your psalm report file 
+With this path you can specify the path your psalm report file
 
 
 **output_format**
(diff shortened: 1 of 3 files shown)


veewee:
Hello @mwolff-fn

Are you using latest grumphp version?
It should detect phars without adding flags on every task.
More context:
https://github.com/phpro/grumphp/pull/739

mwolff-fn:
> Are you using latest grumphp version?
> It should detect phars without adding flags on every task.

Hey @veewee , thanks for the quick reply. I did indeed not see that since the project I was trying to use Grumphp on still has an old PHP version constraint set in the composer.json, so I only got to version 0.16.*.

The solution introduced in 0.18.0 solves the problem and is also a lot better since it applies to all tasks - thanks for the hint!!

Will close the PR.

PR #766, opened 2020-05-08
[phpmd] Allow setting output format to ansi

MarkVaughn:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

Since [2.8](https://phpmd.org/#phpmd-2-8-0-2019-12-18-) phpmd has a new, subjectively nicer, report format that is ansi

**ansi**
Screen Shot 2020-05-08 at 1 07 25 PM
**text (current)**
image

I did not set the default to ansi for this PR, as the renderer is not available in older versions of phpmd

### DIFF
--- a/doc/tasks/phpmd.md
+++ b/doc/tasks/phpmd.md
@@ -19,6 +19,7 @@ parameters:
         phpmd:
             whitelist_patterns: []
             exclude: []
+            report_format: text
             ruleset: ['cleancode', 'codesize', 'naming']
             triggered_by: ['php']
 ```
@@ -40,6 +41,13 @@ whitelist_patterns:
 *Default: []*
 
 This is a list of patterns that will be ignored by phpmd. With this option you can skip directories like tests. Leave this option blank to run phpmd for every php file.
+
+**report_format**
+
+*Default: text*
+
+This sets the output [renderer](https://phpmd.org/documentation/#renderers) of phpmd.
+Available formats: ansi, text.
 
 **ruleset**
 
--- a/src/Task/PhpMd.php
+++ b/src/Task/PhpMd.php
@@ -22,12 +22,15 @@ class PhpMd extends AbstractExternalTask
         $resolver->setDefaults([
             'whitelist_patterns' => [],
             'exclude' => [],
+            'report_format' => 'text',
             'ruleset' => ['cleancode', 'codesize', 'naming'],
             'triggered_by' => ['php'],
         ]);
 
         $resolver->addAllowedTypes('whitelist_patterns', ['array']);
         $resolver->addAllowedTypes('exclude', ['array']);
+        $resolver->addAllowedTypes('report_format', ['string']);
+        $resolver->addAllowedValues('report_format', ['text', 'ansi']);
         $resolver->addAllowedTypes('ruleset', ['array']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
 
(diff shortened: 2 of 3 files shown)


MarkVaughn:
the linked doc doesn't mention the reporter but I made a PR for that on phpmd https://github.com/phpmd/phpmd/pull/779

doc/tasks/phpmd.md
```diff
@@ -19,6 +19,7 @@ parameters:
         phpmd:
             whitelist_patterns: []
             exclude: []
+            report_format: text
```
veewee, line 22: Is it possible to set this option the null and let it fall back to the defaults inside the CLI tool?

test/Unit/Task/PhpMdTest.php
```diff
@@ -27,6 +27,7 @@ public function provideConfigurableOptions(): iterable
             [
                 'whitelist_patterns' => [],
                 'exclude' => [],
+                'report_format' => 'text',
```
veewee, line 30: Can you add a case for baseline in `provideExternalTaskRuns` as well?

The provider is added to validate that given the task config (first argument), the command line tool is being called with the correct CLI arguments (last argument).

For every config option, we require a task to make sure that the CLI is triggered correctly.

doc/tasks/phpmd.md
MarkVaughn: the report format is required by phpmd and previously the task here has set it to text. 
As seen here https://github.com/phpro/grumphp/pull/766/files#diff-a5512ab2de18407a205f9c4a15f17427L67
```
        $arguments->add('text');	
```
I'm not sure what you are asking of me.

test/Unit/Task/PhpMdTest.php
MarkVaughn: Will do later

doc/tasks/phpmd.md
veewee: ok, thanks for the feedback. That seems like the best choice then!

test/Unit/Task/PhpMdTest.php
```diff
@@ -143,5 +144,18 @@ public function provideExternalTaskRuns(): iterable
                 'cleancode',
             ]
         ];
+
+        yield 'report_formats' => [
+            [
+                'report_format' => ['ansi'],
```
veewee, line 150: This should be type string.

MarkVaughn:
I'm sorry for the back and forth, I'm not able to run the tests locally because seemingly no tests work  on my environment (osx, php7.4.2) getting lots of deprecation errors when running tests e.g

```
12) GrumPHPTest\Uni\Task\PhpMdTest::it_passes_on_stuff with data set "exitCode0" (array(), Double\GrumPHP\Task\Context\RunContext\P9 Object (...), Closure Object (...))
Function ReflectionType::__toString() is deprecated
```

src/Task/PhpMd.php
```diff
@@ -64,7 +66,8 @@ public function run(ContextInterface $context): TaskResultInterface
 
         $arguments = $this->processBuilder->createArgumentsForCommand('phpmd');
         $arguments->addCommaSeparatedFiles($files);
-        $arguments->add('text');
+        $format = in_array($config['report_format'], ['text', 'ansi']) ? $config['report_format'] : 'text';
```
veewee, line 69: It's better to do these kind of normalizations inside the options resolver directly:
https://symfony.com/doc/current/components/options_resolver.html#option-normalization

veewee:
Thanks for your work!

Issue #756, opened 2020-04-10
Composer 2.0.0 compatibility


veewee:
| Q               | A
| --------------- | ---
| Version         | 0.18
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

Since composer is almost going to release 2.0.0, we'll have to make sure GrumPHP is compatible as well!

To inspect: Upgrade info

https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-integrators-and-plugin-authors

veewee:
Will be fixed in #768

PR #768, opened 2020-05-16
Composer v2 compatibility

JeppeKnockaert:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #756 

This pull request adds support for Composer v2. I've also added the version of composer to the test matrix, so both versions are tested.

- Wrong namespaces don't get autoloaded anymore in Composer v2, so I fixed that as well.
- Paratest has not created a version supporting composer v2 yet, this is why I included the dev requirement
- friendsofphp/php-cs-fixer has an open PR for composer v2 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4935), but doesn't even seem to be needed anymore in the dev requirements, so I left it out. The alternative would be to include a phar version of php-cs-fixer for Travis.
- hirak/prestissimo should not be installed anymore for composer v2 (this is now built-in), see https://github.com/hirak/prestissimo/issues/218
- I don't think we need to use deactivate/uninstall (for now), because cleanup is already done by using events. Maybe things can be cleaned up / optimized after composer v1 has been dropped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ php:
 
 env:
   matrix:
-    -
-    - DEPENDENCIES=--prefer-lowest
+    - COMPOSER_VERSION=1
+    - COMPOSER_VERSION=2
+    - COMPOSER_VERSION=1 DEPENDENCIES=--prefer-lowest
 
 cache:
   directories:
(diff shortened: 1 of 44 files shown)


veewee:
Thanks for the PR! It looks really nice.

The error is related to a broken openssl on appveyor, will have to take a look at why it fails before I merge it in.

JeppeKnockaert:
Thank you @veewee. It seems every recent PR has this issue. I believe it is linked to this PHP issue, which was fixed in 7.4.6RC1 according to the comments in that issue (https://bugs.php.net/bug.php?id=79589).

veewee:
yeah .. maybe best to hardcode it to a previous 7.4 version for now?

veewee:
Thanks again!

PR #776, opened 2020-06-22
paratest: Add support for '--coverage-clover' and '--coverage-php'

LompiSystems:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

Now for paratest task grumphp support only `--coverage-html` and `--coverage-xml` agrument. I added support also for `--coverage-clover` and `--coverage-php`.
Currently paratest don't support generating clover coverage base on phpunit logging configuration, so without that change we can't use paratet to generate correct raport to `clover_coverage` task.

### DIFF
--- a/doc/tasks/paratest.md
+++ b/doc/tasks/paratest.md
@@ -25,8 +25,10 @@ grumphp:
             configuration: null
             runner: null
             debugger: null
-            coverage-xml: null
+            coverage-clover: null
             coverage-html: null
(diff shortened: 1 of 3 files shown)


veewee:
Thanks for adding the options! Looks good!

Issue #779, opened 2020-06-27
Incorrect escapting of EXEC_GRUMPHP_COMMAND


1ed:
| Q               | A
| --------------- | ---
| Version         | 0.19.1
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | #733

**My configuration**
```yaml
grumphp:
    git_hook_variables:
        EXEC_GRUMPHP_COMMAND: 'docker-compose run --rm --no-deps -u $(id -u):$(id -g) php'
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer init --no-interaction && composer require --dev phpro/grumphp

# Your actions
# Please add the steps on how to reproduce the issue here.
vendor/bin/grumphp git:init -vvv && cat .git/hooks/pre-commit

# Run GrumPHP:
git add -A && git commit -m"Test"
# or
./vendor/bin/grumphp run
```

**Result:**
```
Run a one-off command on a service.

For example:

    $ docker-compose run web python manage.py shell

By default, linked services will be started, unless they are already
running. If you do not want to start linked services, use
`docker-compose run --no-deps SERVICE COMMAND [ARGS...]`.

Usage:
    run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] [-l KEY=VALUE...]
        SERVICE [COMMAND] [ARGS...]

Options:
    -d, --detach          Detached mode: Run container in the background, print
                          new container name.
    --name NAME           Assign a name to the container
    --entrypoint CMD      Override the entrypoint of the image.
    -e KEY=VAL            Set an environment variable (can be used multiple times)
    -l, --label KEY=VAL   Add or override a label (can be used multiple times)
    -u, --user=""         Run as specified username or uid
    --no-deps             Don't start linked services.
    --rm                  Remove container after run. Ignored in detached mode.
    -p, --publish=[]      Publish a container's port(s) to the host
    --service-ports       Run command with the service's ports enabled and mapped
                          to the host.
    --use-aliases         Use the service's network aliases in the network(s) the
                          container connects to.
    -v, --volume=[]       Bind mount a volume (default [])
    -T                    Disable pseudo-tty allocation. By default `docker-compose run`
                          allocates a TTY.
    -w, --workdir=""      Working directory inside the container
```

[...]

Issue #733, opened 2020-02-04
Use double quotes instead of simple ones for git_hook_variables argument escaping


Maxooo:
| Q               | A
| --------------- | ---
| Version         | GrumPHP 0.17.2
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

When using `git_hook_variables` parameter, you may also need to use variables inside your configured values. This seems to be impossible to do as the replaced `$(EXEC_GRUMPHP_COMMAND)` variable in the hooks template is replaced by parts escaped with simple quotes (`'`) instead of the double ones (`"`).

**My configuration**

This use case can be found in the [parameters documentation section](https://github.com/phpro/grumphp/blob/master/doc/parameters.md) with the docker implementation. So please use the following configuration : 

```yaml
parameters:
    git_hook_variables:
        EXEC_GRUMPHP_COMMAND: 'docker run --rm -it -v $(pwd):/grumphp -w /grumphp webdevops/php:alpine'
``` 

**Steps to reproduce:**

```sh
# Generate an empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
# Please copy paste the configuration (e.g. previous section) and run the following command
pbpaste > grumphp.yml
composer require --dev phpro/grumphp:0.17.2
```
At this state you do not need much more to anticipate the error. Please open one of the  `.git/hooks/pre-commit` or `.git/hooks/commit-msg` hook files and notice the simple escaping on the GrumPHP running command (last line).

```bash
# .git/hooks/pre-commit

# Run GrumPHP
(cd "./" && printf "%s\n" "${DIFF}" | 'docker' 'run' '--rm' '-it' '-v' '$(pwd):/grumphp' '-w' '/grumphp' 'webdevops/php:alpine' 'vendor/phpro/grumphp/bin/grumphp' 'git:pre-commit' '--skip-success-output')
```
The second file is more interesting as it use variables in the GrumPHP command part (`$GIT_USER`, `$GIT_EMAIL` and `$COMMIT_MSG_FILE`) and they are both well escaped with double quotes (`"`).

```bash
# .git/hooks/commit-msg

# Run GrumPHP
[...]

PR #780, opened 2020-06-27
Fix command escaping

1ed:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #779, #733

The first commit what I need, the others are optional. I removed PhpCsFixerFormatter::formatSuggestion() which was public and not internal so I consider it as a BC break.


src/Console/Command/Git/InitCommand.php
```diff
@@ -133,7 +133,8 @@ protected function parseHookBody(string $hook, string $templateFile): string
         ];
 
         foreach ($this->hooksConfig->getVariables() as $key => $value) {
-            $replacements[sprintf('$(%s)', $key)] = ProcessUtils::escapeArgumentsFromString($value);
+            $process = is_array($value) ? new Process($value) : Process::fromShellCommandline($value);
```
veewee, line 136: Wouldn't it be better to add a configuration normalizer that converts a string into an array?
That way we can always assume we have an array, even if the `grumphp.yaml` contains a string.

I am not sure how this solves your issue though: it still escapes the configured CLI argument.
If you are on windows, it escapes with double quotes, so it might work.
However, if you are on linux: it escapes with single quotes, meaning it won't work.

Not sure if it's a good idea, but maybe we shouldn't escape the variables and leave that up to the one configurating grumphp? That way you can do something like:


```yaml
grumphp:
    git_hook_variables:
        EXEC_GRUMPHP_COMMAND: ['docker-composer --rm --no-deps -u "$(id -u):$(id -g)"', "'php'" ]
```

Which will result in:

```
docker-composer --rm --no-deps -u "$(id -u):$(id -g)" 'php'
```


Meaning you can choose if you want to escape or not from the config file.
However: it means we won't escape by default anymore.... 

(Assuming it is a developer tool, this won't be a big security issue IMO.)

test/E2E/AbstractE2ETestCase.php
```diff
@@ -95,7 +95,7 @@ protected function initializeComposer(string $path): string
     {
         $process = new Process(
             [
-                $this->executableFinder->find('composer'),
+                realpath($this->executableFinder->find('composer', 'composer')),
```
veewee, line 98: Can you tell me what the issue is exactly? 
Why are you defaulting to the string 'composer' if the file cannot be found?
I try to avoid realpath as much as possible in this package, since it already caused a lot of issues.

src/Formatter/PhpCsFixerFormatter.php
```diff
@@ -36,16 +35,6 @@ public function format(Process $process): string
         return $this->formatJsonResponse($json);
     }
 
-    public function formatSuggestion(Process $process): string
```
veewee, line 39: Agree we can get rid of this method. However, we should still suggest the fix command. (See further in CR)

test/E2E/GitHookParametersTest.php
```diff
@@ -41,14 +41,14 @@ function it_can_specify_hook_exec_command_with_additional_arguments()
...
-                    'EXEC_GRUMPHP_COMMAND' => $php.' -d date.timezone=Europe/Brussels',
+                    'EXEC_GRUMPHP_COMMAND' => [$php, '-d date.timezone=Europe/Brussels'],
                 ],
             ],
         ]);
 
         $this->installComposer($this->rootDir);
 
-        $hookPattern = '{[\'"]?'.preg_quote($php, '{').'[\'"]? [\'"]?-d[\'"]? [\'"]?date\.timezone=Europe/Brussels[\'"]?}';
+        $hookPattern = sprintf('{[\'"]%s[\'"] [\'"]-d date\.timezone=Europe/Brussels[\'"]}', $php);
```
veewee, line 51: better to keep `$php` preg_quote-ed

src/Task/PhpCsFixer.php
```diff
@@ -102,7 +102,7 @@ public function run(ContextInterface $context): TaskResultInterface
 
         if (!$process->isSuccessful()) {
             $messages = [$this->formatter->format($process)];
-            $fixerCommand = $this->formatter->formatSuggestion($process);
+            $fixerCommand = $process->getCommandLine();
```
veewee, line 105: Instead of creating the suggestion in the formatter, we can add arguments to the `$arguments` variable as we do in the eslint task:
https://github.com/phpro/grumphp/blob/master/src/Task/ESLint.php#L92-L105

This way, we are still suggesting the fix command to the end user in case of a detected fixable item.

src/Console/Command/Git/InitCommand.php
1ed: In this way it's up to the developer if the command will be escaped or not. If someone uses a string it won't be escaped automatically, but if someone uses an array it will be escaped by the process component.
Won't doing automatic escaping would be the other option, but I think it's more convenient this way. It's up to the user to decide.

If it would be normalized to an array how could we decide if it needs escaping or not?

In my case I can use `EXEC_GRUMPHP_COMMAND: 'docker-compose run --rm --no-deps -u $(id -u):$(id -g) php'`

test/E2E/AbstractE2ETestCase.php
1ed: I have `...:vendor/bin:...` in the `$PATH` so the executable finder finds the first and tries to run that with a relative path.

src/Task/PhpCsFixer.php
1ed: Fixed, thanks!

test/E2E/GitHookParametersTest.php
1ed: added back

src/Composer/DevelopmentIntegrator.php
```diff
@@ -60,6 +59,6 @@ private static function fixInternalComposerProcessVersion(
             return $process;
         }
 
-        return new Process(ProcessUtils::escapeArguments($commandlineArgs->getValues()));
+        return new Process($commandlineArgs->getValues());
```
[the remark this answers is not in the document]
1ed: Fixed

spec/Process/ProcessBuilderSpec.php
```diff
@@ -74,7 +73,7 @@ function it_outputs_the_command_when_run_very_very_verbose(IOInterface $io)
         $io->isVeryVerbose()->willReturn(true);
 
         $command = '/usr/bin/grumphp';
-        $io->write([PHP_EOL . 'Command: ' . ProcessUtils::escapeArgument($command)], true)->shouldBeCalled();
+        $io->write([PHP_EOL . sprintf('Command: %1$s%2$s%1$s', '\\' === DIRECTORY_SEPARATOR ? '"' : "'", $command)], true)->shouldBeCalled();
```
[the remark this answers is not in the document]
1ed: `Process::fromShellCommandline` not escapes the command. I added a check if the call to `$io->write` contains the command string, maybe that's enough

src/Console/Command/Git/InitCommand.php
veewee: sounds good!

test/E2E/AbstractE2ETestCase.php
veewee: @janvernieuwe  : Does this fix the issue you had while running paratest as well? It might be related

Issue #769, opened 2020-05-29
Use stand-alone tools instead of composer dependencies


veewee:
Currently we are adding dev-dependencies to the project.
However, the dependency tree is getting quite complex with a combination of tools like phpunit, phpspec and paralel all installed together.

It is e.g. also not possible to install psalm on composer 2 on versions lower then 7.4
However, we do added psalm config in #741
(Still todo : fix level 3 reported issues)

Issue #767, opened 2020-05-11
TestCase compatibility with recent versions of phpunit


Taluu:
| Q               | A
| --------------- | ---
| Version         | 0.18.0
| Bug?            | kinda
| New feature?    | no
| Question?       | kinda
| Documentation?  | no
| Related tickets | ~

As a task developper, I'd like to use the `AbstractTaskTestCase` class to make my tests on the tasks.

The thing is, it doesn't seem that the classes `AbstractTaskTestCase` (and probably the external one) are supporting phpunit > 7, especially with the setUp method that changed its signature I think it was on PHPUnit 8, and some assert helpers have been dropped (don't know which ones though).

Maybe using symfony's bridge would be an option ?

Taluu:
Here's the error I have :

```bash
$ composer req --dev phpunit/phpunit
Using version ^9.1 for phpunit/phpunit
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 30 installs, 0 updates, 0 removals
  - Installing sebastian/version (3.0.0): Downloading (100%)
  - Installing sebastian/type (2.0.0): Downloading (100%)
  - Installing sebastian/resource-operations (3.0.0): Downloading (100%)
  - Installing sebastian/recursion-context (4.0.0): Downloading (100%)
  - Installing sebastian/object-reflector (2.0.0): Downloading (100%)
  - Installing sebastian/object-enumerator (4.0.0): Downloading (100%)
  - Installing sebastian/global-state (4.0.0): Downloading (100%)
  - Installing sebastian/exporter (4.0.0): Downloading (100%)
  - Installing sebastian/environment (5.1.0): Downloading (100%)
  - Installing sebastian/diff (4.0.1): Downloading (100%)
  - Installing sebastian/comparator (4.0.0): Downloading (100%)
  - Installing sebastian/code-unit (1.0.2): Downloading (100%)
  - Installing phpunit/php-timer (3.1.4): Downloading (100%)
  - Installing phpunit/php-text-template (2.0.0): Downloading (100%)
  - Installing phpunit/php-invoker (3.0.0): Downloading (100%)
  - Installing phpunit/php-file-iterator (3.0.1): Downloading (100%)
  - Installing theseer/tokenizer (1.1.3): Downloading (100%)
  - Installing sebastian/code-unit-reverse-lookup (2.0.0): Downloading (100%)
  - Installing phpunit/php-token-stream (4.0.1): Downloading (100%)
[...]

PR #789, opened 2020-07-15
Standalone tools

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes: types + php
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #769, #767

❗  Dropped support for PHP 72 (because of dependencies and since it is in security fixes only mode untill the end of the year)

* This PR upgrades phpunit, phpspec, paratest, ...
* Phpcs, psalm and composer-normalize are placed inside the repository as standalone tools (through phive)
* Psalm level 3 issues are fixed with type detection of 96%

Phpunit could not be used as a standalone tool, since we require the prophecy bridge which is also linked to phpspec.


veewee:
CI issues are related to https://github.com/vimeo/psalm/issues/2858

veewee:
Awaiting #790 so that we can use composer-normalize as standalone tool.

PR #790, opened 2020-07-17
Add a way to call the standalone composer-normalize command

prudloff-insite:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

composer-normalize is also available as a stand-alone PHAR (https://github.com/ergebnis/composer-normalize#phar).
This PR adds a new option to call this stand-alone command instead of `composer normalize`.

### DIFF
--- a/src/Task/ComposerNormalize.php
+++ b/src/Task/ComposerNormalize.php
@@ -23,6 +23,7 @@ class ComposerNormalize extends AbstractExternalTask
             'indent_size' => null,
             'indent_style' => null,
             'no_update_lock' => true,
+            'use_standalone' => false,
             'verbose' => false,
         ]);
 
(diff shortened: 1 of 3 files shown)


test/Unit/Task/ComposerNormalizeTest.php
```diff
@@ -25,6 +25,7 @@ public function provideConfigurableOptions(): iterable
         yield 'defaults' => [
             [],
             [
+                'use_standalone' => false,
```
veewee, line 28: Can you add a case in which you test the `use_standalone` option?

veewee:
Thanks for the PR. Looks good but requires an additional test case.

src/Task/ComposerNormalize.php
```diff
@@ -48,6 +49,11 @@ public function run(ContextInterface $context): TaskResultInterface
 
         $arguments = $this->processBuilder->createArgumentsForCommand('composer');
```
veewee, line 50: If you use standalone and "composer" could not be found in your path, this crashes.
Better to call the `createArgumentsForCommand` method with the real tool name based on the `use_standalone` param

PR #793, opened 2020-07-20
Use no-interaction in Git hook

prudloff-insite:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

When using `fix_by_default: true`, the fix is not applied automatically on commit.
That seems to be because `QuestionHelper::ask()` thinks the input is interactive. 

This PR calls the Git hook with `--no-interaction` so that the question helper knows the input is not interactive and correctly returns the default answer.

### DIFF
--- a/resources/hooks/local/pre-commit
+++ b/resources/hooks/local/pre-commit
@@ -12,4 +12,4 @@ DIFF=$(git -c diff.mnemonicprefix=false --no-pager diff -r -p -m -M --full-index
 export GRUMPHP_GIT_WORKING_DIR=$(git rev-parse --show-toplevel)
 
 # Run GrumPHP
-(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | $(EXEC_GRUMPHP_COMMAND) $(HOOK_COMMAND) '--skip-success-output')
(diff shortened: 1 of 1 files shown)


resources/hooks/local/pre-commit
```diff
@@ -12,4 +12,4 @@ DIFF=$(git -c diff.mnemonicprefix=false --no-pager diff -r -p -m -M --full-index
 export GRUMPHP_GIT_WORKING_DIR=$(git rev-parse --show-toplevel)
 
 # Run GrumPHP
-(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | $(EXEC_GRUMPHP_COMMAND) $(HOOK_COMMAND) '--skip-success-output')
+(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | $(EXEC_GRUMPHP_COMMAND) $(HOOK_COMMAND) '--skip-success-output' '--no-interaction')
```
veewee, line 15: It is better to move this additional parameter (and possibly the `--skip-success-output` as well to the hook builder:

https://github.com/phpro/grumphp/blob/2ab78e5bdc08d4a579d8e7c12575a3203c88a95e/src/Console/Command/Git/InitCommand.php#L159-L175

This way, all hooks contain the same parameters.

veewee:
Just found out it is fixed by PR https://github.com/phpro/grumphp/pull/788.

Issue #803, opened 2020-08-22
Using strict matching in git blacklist task how can i avoid `dd` keyword matching when some text comes before it?


morloderex:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 0.20.0`
| Bug?            | Need to require more information (so maybe)
| New feature?    | unsure
| Question?       | yes
| Documentation?  | no
| Related tickets | None

Given I have this git blacklist configuration.

```yaml
git_blacklist_php:
  keywords:
    - "var_dump("
    - "die("
    - "die;"
    - "dd("
    - "exit("
    - "exit;"
  triggered_by: [php]
  regexp_type: E
  metadata:
     task: git_blacklist
```

If i now add some text like `function add($task)` in my file, the task will see this as a failure, because it does not do excat matches. Is there a way that I can make it so that `add(` does not trigger a task failure but `dd(` does?

I have a situation where i want to filter out any calls to a function named `dd` but anything before it is alright. I have already been tring to do it using a regex start character `^` in front if the keyword but that does not help at all.

**My configuration**
```yaml
Already pasted a snippet of my configuration file above.
# Please add a copy of your grumphp.yml file.
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer require --dev phpro/grumphp

# Your actions
Add a any php file with some text having anything before `dd(` like a function name for instance.

# Run GrumPHP:
git add -A && git commit -m"Test"
# or
./vendor/bin/grumphp run
```

**Result:**
```
git_blacklist_php
=================

GrumPHP detected a pre-commit command.
GrumPHP is sniffing your code!

Running tasks with priority 0!
==============================

Running task 1/6: git_blacklist_php... ✘
Aborted ...

git_blacklist_php
=================

You have blacklisted keywords in your commit:
-- filename -- (omitted for security reasons)
90:    public function authorizedToAdd(NovaRequest $request, $model)

```

Notice the function name it ends with `dd(` which correctly matches the keyword but i would like to allow for this because it's has additional characters behind it it's not en excat match.

PR #804, opened 2020-08-22
Add `match_word` option to git blacklist task

morloderex:
Added a match_word option to git blacklist.

Closes: #803

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #803

So I figured my issue #803 out by myself, as it turns out `git grep` command has an option called word `--word-regexp` or `-w` for sorthand.

This has been taken from the [Git scm documentation](https://git-scm.com/docs/git-grep)

```
Match the pattern only at word boundary 

(either begin at the beginning of a line, or preceded by a non-word character; end at the end of a line or followed by a non-word character).
```

I have added the possiblity to opt-in to this functionallity by using the new `match_word` option on the task. However I am fully open to changing this option on the task to something else. Some alternative naming conventions considered `strict_keywords`, `match_keywords`, `match_using_boundary`

All tests passes however let me know if I need to address anything before this can be merged I am new to contributing here.

### DIFF
--- a/doc/tasks/git_blacklist.md
+++ b/doc/tasks/git_blacklist.md
@@ -15,6 +15,7 @@ grumphp:
             whitelist_patterns: []
             triggered_by: ['php']
             regexp_type: G
+            match_word: true
 ```
 
 **keywords**
@@ -49,3 +50,13 @@ You can overwrite this option to whatever filetype you want to validate!
 *Default: G*
 
 This option allows you to choose the type of regexp you want to use for patterns (can be G for POSIX basic, E for POSIX extended or P for Perl Compatible).
+
+**match_word**
+
+*Default: false*
+
+This option allows you to choose how the keywords is found.
+
+For instance let's say you have a keyword looking like `"dd("` by default this task would also find any
+text before or after the keyword meaning this: `function add($someTask)` would still be considered invalid.
+This configuration option allows you to get around that issue.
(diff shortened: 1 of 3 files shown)


src/Task/Git/Blacklist.php
```diff
@@ -81,6 +84,11 @@ public function run(ContextInterface $context): TaskResultInterface
         $arguments->add('-n');
         $arguments->add('--break');
         $arguments->add('--heading');
+
+        if (true === $matchWord) {
```
veewee, line 88: To make the task more readable, you should go for:

```php
$arguments->addOptionalArgument('--word-regexp, $config['match_word']);
```

(1 line instead of 6)

doc/tasks/git_blacklist.md
```diff
@@ -15,6 +15,7 @@ grumphp:
             whitelist_patterns: []
             triggered_by: ['php']
             regexp_type: G
+            match_word: true
```
veewee, line 18:
```suggestion
            match_word: false
```

veewee (changes requested):
Thanks for the PR, it looks good!
I've added 2 small remarks.

src/Task/Git/Blacklist.php
morloderex: Oh yeah, of cause. 
At first i could not figure out how to get the tests to pass with a boolean (i kept getting an error that the default value was incorrect) So i just forgot to remove the if.

I will make that change.

morloderex: This has been committed

veewee:
Thanks for your work!

Issue #817, opened 2020-09-21
grumphp is really verbose in a CI


prudloff-insite:
| Q               | A
| --------------- | ---
| Version         | 0.22.0
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | 

When running in a CI, grumphp is really verbose because the `overwrite()` method of `ConsoleSectionOutput` is not able to replace the text.

It would be nice to have an option to disable this behaviour and simply display the result at the end.

**My configuration**
```yaml
---
grumphp:
  ascii:
    failed: ~
    succeeded: ~
  tasks:
    composer: ~
    jsonlint: ~
    yamllint: ~
    xmllint: ~
    phpcs:
      standard: tests/phpcs.xml
      triggered_by:
        - php
        - module
        - inc
        - install
        - test
        - profile
        - theme
        - info
        - txt
        - md
    phpmd:
      ruleset:
        - tests/phpmd.xml
    git_blacklist:
      keywords:
        - "kint("
        - "|raw"
        - "autoescape"
      triggered_by:
        - twig
```

**Steps to reproduce:**
You can simulate this behaviour like this:
```sh
./vendor/bin/grumphp run | cat
```

**Result:**
```
GrumPHP is sniffing your code!

Running tasks with priority 0!
==============================

Running task 1/6: composer... 
Running task 2/6: jsonlint... 
Running task 3/6: yamllint... 
Running task 4/6: xmllint... 
Running task 5/6: phpcs... 
Running task 6/6: phpmd... 
Running task 1/6: composer... 
Running task 2/6: jsonlint... 
Running task 3/6: yamllint... 
Running task 4/6: xmllint... ✔
Running task 5/6: phpcs... 
Running task 6/6: phpmd... 
Running task 1/6: composer... 
Running task 2/6: jsonlint... ✔
Running task 3/6: yamllint... 
Running task 4/6: xmllint... ✔
Running task 5/6: phpcs... 
Running task 6/6: phpmd... 
Running task 1/6: composer... 
Running task 2/6: jsonlint... ✔
Running task 3/6: yamllint... ✔
Running task 4/6: xmllint... ✔
Running task 5/6: phpcs... 
Running task 6/6: phpmd... 
Running task 1/6: composer... ✔
Running task 2/6: jsonlint... ✔
Running task 3/6: yamllint... ✔
Running task 4/6: xmllint... ✔
Running task 5/6: phpcs... 
[...]

PR #818, opened 2020-09-22
Improve no-ansi reporting

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #817

This PR creates a less verbose output on non-ansi systems like Github actions, Travis, ...
You can switch the behaviour by adding `--ansi` or `--no-ansi` flag.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "doctrine/collections": "^1.6.7",
         "gitonomy/gitlib": "^1.0.3",
         "monolog/monolog": "~1.16 || ^2.0",
+        "ondram/ci-detector": "^3.5",
         "opis/closure": "^3.5",
         "psr/container": "^1.0",
         "seld/jsonlint": "~1.1",
--- a/resources/config/runner.yml
+++ b/resources/config/runner.yml
@@ -110,4 +110,12 @@ services:
         arguments:
             - '@grumphp.io'
             - '@GrumPHP\Runner\MemoizedTaskResultMap'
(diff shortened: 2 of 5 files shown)


veewee:
This doesn't solve the issue on travis, since they do run in ANSI mode - but have a different way for displaying the results.
Maybe a better idea to determine if we are running in a known CI?

Psalm does it this way : https://github.com/vimeo/psalm/blob/37a2f8a33dc38300a7f19cbaafa1a68d4f890f1a/src/psalm.php#L380-L386

drupol:
Maybe we can use https://github.com/OndraM/ci-detector just like infection/infection.

prudloff-insite:
This does not fix the issue on GitLab CI either.

veewee:
This works for me.
I added the CI package that @drupol suggested, which means you don't have to manually add `--no-ansi` to the cli command + it still prints in green color.

On non-ascii situations, you will first see a list of all tasks that will be executed. Once all tasks are done, you will see the result of all tasks. This limits the amount of messages to 2 blocks. 

Also tried a more linear approach, but it didnt work well in combination with parallel task execution.
For me, this is good enough at the moment.

Feel free to improve or point out your remarks.

Example on appveyor:

Screenshot 2020-10-30 at 10 25 47

drupol:
Very nice :-)

PR #822, opened 2020-10-01
PHPStan task : search automatically phpstan.neon

DuboisS:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets |

It is now no longer necessary to specify its phpstan configuration file if it has the default name.

### DIFF
--- a/src/Task/PhpStan.php
+++ b/src/Task/PhpStan.php
@@ -73,6 +73,14 @@ class PhpStan extends AbstractExternalTask
             ->notPaths($config['ignore_patterns'])
             ->extensions($config['triggered_by']);
 
+        if (!($config['configuration'])) {
+            if (file_exists('phpstan.neon')) {
+                $config['configuration'] = 'phpstan.neon';
+            } elseif(file_exists('phpstan.neon.dist')) {
+                $config['configuration'] = 'phpstan.neon.dist';
+            }
+        }
+
         if (!empty($config['force_patterns'])) {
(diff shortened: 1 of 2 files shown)


src/Task/PhpStan.php
```diff
@@ -73,6 +73,14 @@ public function run(ContextInterface $context): TaskResultInterface
             ->notPaths($config['ignore_patterns'])
             ->extensions($config['triggered_by']);
 
+        if (!($config['configuration'])) {
```
veewee, line 76: https://phpstan.org/config-reference#config-file

> If you do not provide a config file explicitly, PHPStan will look for files named phpstan.neon or phpstan.neon.dist in current directory.

Why don't we let the tool decide? It detects these files by default - no need to maintain it in here.
If you want to specify a custom location, you can always do so in your configuration.
If you want to use the defaults, you can just net set the configuration option (which will fallback to the phpstan defaults.)

veewee (commented):
Thanks for the PR.
Can you take a look at my question?

src/Task/PhpStan.php
DuboisS: Sorry, I was wrong about that. My mistake. 
After further testing, it does indeed take the phpstan.neon file well even if we don't specify it.

veewee: No problem. :)

PR #824, opened 2020-10-01
Consistently use yml instead of yaml extension in documentation

verbruggenalex:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets |

Even though I can see that grumphp allows for both yml and yaml configuration files like it is supposed to:

https://github.com/phpro/grumphp/blob/0e6287a3ae70a8730d11a239449e991b3fc78f99/src/Locator/GuessedPathsLocator.php#L109-L112

I think semantically it's best to stay consistent through out the documentation to only reference one of the allowed extensions. A quick search showed that yml was the prefered extension:

- 11 results in 6 files for grumphp.yaml
- 148 results in 71 files for grumphp.yml

This pull request changes the extension in the documentation only. No code is touched. Maybe we could document that the GuessedPathsLocator allows for both yml and yaml extension on the grumphp configuration files. I just don't really know where to put that.

### DIFF
--- a/doc/installation/exotic.md
+++ b/doc/installation/exotic.md
@@ -116,7 +116,7 @@ If you want to support multiple bin directories, you can prepend them to your pa
 
 ### Environment variables
 
-It is also possible to set some of the environment variables above inside the `grumphp.yaml` file directly:
+It is also possible to set some of the environment variables above inside the `grumphp.yml` file directly:
 
 ```yaml
 grumphp:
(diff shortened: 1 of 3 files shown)


veewee:
Thanks! Makes sense to make it consistent. There isn't really a preferred extension at the moment.
So it's ok to stick with `.yml`, since it is the first in the list and will be autocreated by GrumPHP.

Issue #835, opened 2020-11-06
"No match for regexp /diff --git (a\/.*) (b\/.*)\n/" if diff.noprefix is true


sascha-egerer:
| Q               | A
| --------------- | ---
| Version         |  0.22.0
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | #164 

```sh
$ git commit --amend

In ParserBase.php line 67:

  No match for regexp /diff --git (a\/.*) (b\/.*)\n/ Upcoming: diff --git README.md README.md

git:pre-commit [--skip-success-output] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-c|--config CONFIG] [--] <command>
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer require --dev phpro/grumphp

# Your actions
# Please add the steps on how to reproduce the issue here.

git config diff.noprefix=true

# Run GrumPHP:
git add -A && git commit -m"Test"
```

**Result:**
```
In ParserBase.php line 67:

  No match for regexp /diff --git (a\/.*) (b\/.*)\n/ Upcoming: diff --git README.md README.md


git:pre-commit [--skip-success-output] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-c|--config CONFIG] [--] <command>
```

veewee:
Thanks! Continuing in the PR.

tiagodevweb:
In my case I solved it by deleting the build folder:

**git -c diff.mnemonicprefix=false -c diff.noprefix=false --no-pager diff -r -p -m -M --full-index --no-color --staged -- . ':(exclude)build/*' | cat**

PR #836, opened 2020-11-06
Make "git diff" behave more predictably #835

sascha-egerer:
| Q             | A
| ------------- | ---
| Branch        | *
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #835

Make gumphp work with git config diff.noprefix=true


veewee:
Thanks for the PR, looks good!
It would be better if the diff parser from gitonomy/gitlib supported more modes though :)

PR #838, opened 2020-11-07
add Page Views

xdvrx1:
When we add this service, it will count every hit of this repo. And this will guide us more about the visitors each day and will indicate the total views. For me, this is very helpful both for us and for those will view this repo seeing this page views. If there is the website built from this repo, it can be simply added there too.

Screenshot (1643)


veewee:
Hello @xdvrx1,

Thanks for the PR. However, we are not interested in tracking our users.

PR #841, opened 2020-11-14
Implement refactored formatter options for Deptrac >= v0.9.0

jerowork:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Documented?   | yes
| Fixed tickets | none

# Description
Deptrac has updated formatter options in `v0.9.0`. They enabled new formatters (e.g. baseline, table) and refactored existing formatters (e.g. Graphviz). This means that the previous options are not working anymore. New options are introduced for existing formatters.

This PR adds all new formatters with its new options to the `DeptracTask`.

_See commit descriptions for further information._

## Resources
- https://github.com/sensiolabs-de/deptrac#formatters
- https://github.com/sensiolabs-de/deptrac/blob/master/CHANGELOG.md#090---2020-10-30

### DIFF
--- a/src/Task/Deptrac.php
+++ b/src/Task/Deptrac.php
@@ -21,19 +21,19 @@ class Deptrac extends AbstractExternalTask
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'depfile' => null,
-            'formatter_graphviz' => false,
-            'formatter_graphviz_display' => true,
-            'formatter_graphviz_dump_image' => null,
-            'formatter_graphviz_dump_dot' => null,
-            'formatter_graphviz_dump_html' => null,
+            'formatter' => [],
+            'graphviz_display' => true,
+            'graphviz_dump_image' => null,
+            'graphviz_dump_dot' => null,
+            'graphviz_dump_html' => null,
         ]);
 
         $resolver->addAllowedTypes('depfile', ['null', 'string']);
-        $resolver->addAllowedTypes('formatter_graphviz', ['bool']);
-        $resolver->addAllowedTypes('formatter_graphviz_display', ['bool']);
-        $resolver->addAllowedTypes('formatter_graphviz_dump_image', ['null', 'string']);
-        $resolver->addAllowedTypes('formatter_graphviz_dump_dot', ['null', 'string']);
-        $resolver->addAllowedTypes('formatter_graphviz_dump_html', ['null', 'string']);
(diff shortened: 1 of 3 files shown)


src/Task/Deptrac.php
```diff
@@ -34,6 +42,14 @@ public static function getConfigurableOptions(): OptionsResolver
         $resolver->addAllowedTypes('formatter_graphviz_dump_image', ['null', 'string']);
         $resolver->addAllowedTypes('formatter_graphviz_dump_dot', ['null', 'string']);
         $resolver->addAllowedTypes('formatter_graphviz_dump_html', ['null', 'string']);
+        $resolver->addAllowedTypes('formatter', ['array']);
```
veewee, line 45: string[]

src/Task/Deptrac.php
```diff
@@ -54,11 +70,47 @@ public function run(ContextInterface $context): TaskResultInterface
...
+            $arguments->addOptionalArgument(
+                '--formatter-graphviz-dump-html=%s',
+                $config['formatter_graphviz_dump_html']
+            );
+        }
+
+        // Deptrac versions >= v0.9.0
+        $arguments->addArgumentArray('--formatter=%s', $config['formatter']);
+
+        if (in_array('graphviz', $config['formatter'], true) === true) {
```
veewee, line 95: Would it make sense to normalize the input options instead of adding all these in_arrays?
E.g.: if graphiz is not included in the formatter, the graphiz config option is set to null?

That would keep the task flow a bit more straight forward..

veewee (commented):
Hello,

Thanks for the PR!
This configuration is becoming quite hard for the end user.
Since the tool is in `0.X` versioing state, people know it is quite experimental.

I'dd suggest only adding the new types without providing BC grumphp configuration.

What do you think?

jerowork:
> I'dd suggest only adding the new types without providing BC grumphp configuration.
> 
> What do you think?

It certainly makes it easier indeed. As it's below v1.0, I agree with you. I will remove the pre-v0.9 options completely 👍

src/Task/Deptrac.php
jerowork: I now see that  `addOptionalArgument` is actually doing what was needed :). Updated!

veewee (approved):

veewee:
Thanks for changing and adding the options!

PR #842, opened 2020-11-18
Check config before skipping

dhaarbrink:
This adds a check on the 'use_grumphp_paths' config before skipping so this task runs even though there are no changed files when `use_grumphp_paths` is set to false.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | no
| Fixed tickets | -

# New Task Checklist:

### DIFF
--- a/src/Task/PhpStan.php
+++ b/src/Task/PhpStan.php
@@ -78,7 +78,7 @@ class PhpStan extends AbstractExternalTask
             $files = $files->ensureFiles($forcedFiles);
         }
 
-        if (0 === \count($files)) {
+        if (0 === \count($files) && $config['use_grumphp_paths']) {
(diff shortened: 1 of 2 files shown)


veewee:
Thanks for the PR! It makes sense to do it like this.

Can you also add a unit test covering this case? 
There is already a test for when there are files available, so it should be a small addition:
https://github.com/phpro/grumphp/blob/master/test/Unit/Task/PhpStanTest.php#L192

Landerstraeten:
@veewee Can you check if I have added the correct unit test?

Landerstraeten:
@dhaarbrink Thank you!

PR #855, opened 2020-12-14
Catch ExecutableNotFoundException

prudloff-insite:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

Right now, only `CommandNotFoundException` is caught but `ProcessBuilder::createArgumentsForCommand()` can throw `ExecutableNotFoundException`.

This means that if `phpcbf` is not available, it will fail with only:
> The executable for "phpcfbf" could not be found.

Instead of showing the errors + suggesting to install `phpcbf`.

### DIFF
--- a/src/Task/Phpcs.php
+++ b/src/Task/Phpcs.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace GrumPHP\Task;
 
 use GrumPHP\Collection\ProcessArgumentsCollection;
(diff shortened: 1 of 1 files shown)


veewee:
That's much better indeed! Thanks for the PR!

Issue #866, opened 2021-01-22
git_branch_name fails if any whitelist rules don't pass


qooplmao:
| Q               | A
| --------------- | ---
| Version         | `vendor/bin/grumphp -V`
| Bug?            | no
| New feature?    | no
| Question?       | yes
| Documentation?  | no
| Related tickets | PR #867 

`git_branch_name` fails if any of the items in the list fail.

My understanding that a whitelist is an explicit allow list so anything in the list that passes, providing it's not also in the black, should be an instant pass. With the following configuration I would expect a branch called `feature/new-feature` to pass but, at the moment, it would fail as it doesn't also fit `fix/*` .

I know that I could create a single line regex but it seems to defeat the purpose of having a "list".

```yaml
# grumphp.yml
grumphp:
    tasks:
        git_branch_name:
            whitelist:
                - 'feature/*'
                - 'fix/*'
            blacklist:
                - 'develop'
                - 'master'
```

veewee:
Thanks for reporting! Lets continue the discussion in #867.

PR #867, opened 2021-01-22
Test should not fail because it didn't match everything in the whitelist

qooplmao:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #866

`git_branch_name` fails if any of the items in the list fail.

PR changes so that any matching item in the whitelist will return a success unless already featured in the blacklist.

### DIFF
--- a/src/Task/Git/BranchName.php
+++ b/src/Task/Git/BranchName.php
@@ -105,7 +105,7 @@ class BranchName implements TaskInterface
             $additionalModifiersArray = array_filter(str_split($config['additional_modifiers']));
             array_map([$regex, 'addPatternModifier'], $additionalModifiersArray);
             
-            if (preg_match((string) $regex, $name) {
+            if (preg_match((string) $regex, $name)) {
                 if ($isBlacklisted) {
                     continue;   
                 }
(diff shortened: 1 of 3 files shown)


src/Task/Git/BranchName.php
```diff
@@ -95,17 +96,24 @@ public function run(ContextInterface $context): TaskResultInterface
...
             $regex = new Regex($rule);
 
             $additionalModifiersArray = array_filter(str_split($config['additional_modifiers']));
             array_map([$regex, 'addPatternModifier'], $additionalModifiersArray);
-
-            if (!preg_match((string) $regex, $name)) {
-                $errors[] = sprintf('Whitelist rule not matched: %s', $rule);
+            
+            if (preg_match((string) $regex, $name)) {
+                if ($isBlacklisted) {
```
veewee, line 109: Can you add some tests that show what happens if there is a mix of whitelist and blacklist and how it resolves a conflict in between these regexes?

veewee (commented):
It indeed makes more sense to result OK on *any* whitelist item.
Thanks!

src/Task/Git/BranchName.php
qooplmao: I've added a test with a conflict and additional output.  
Does that cover what you are thinking?

Landerstraeten (approved):

veewee:
Thanks!

PR #891, opened 2021-04-28
Fix extremelly minor typos.

drupol:
This is what is happening when looking for documentation and then you find something that can be fixed inline with Github in 10 seconds.


veewee:
Thanks

drupol:
Thanks!

PR #904, opened 2021-04-30
Add Stylelint task

yguedidi:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

Add a task to support [Stylelint](https://stylelint.io/) tool.
Heavily inspired from the existing ESLint task.
Supports relevant CLI supported arguments, based on the `--help` output, the inspiration for the sorting.
Parameter description is comming from [the official documentation](https://stylelint.io/user-guide/usage/cli)

Extra commit for a small CS fix in ESLint test.

# New Task Checklist:

### DIFF
--- a/doc/tasks/stylelint.md
+++ b/doc/tasks/stylelint.md
@@ -1,0 +1,195 @@
+# Stylelint ![fixer](https://img.shields.io/badge/-fixer-informational)
+
+[Stylelint](https://stylelint.io/) is a modern linter that helps you avoid errors and enforce conventions in your styles.
+
+## NPM
+If you'd like to install it globally:
+```bash
+npm -g stylelint
+```
+
+If you'd like install it as a dev dependency of your project:
+```bash
+npm install --save-dev stylelint
+```
+
+Done. See the Stylelint [Getting Started](https://stylelint.io/user-guide/get-started) guide for more info.
+
+## Config
+It lives under the `stylelint` namespace and has the following configurable parameters:
+
+```yaml
+# grumphp.yml
+grumphp:
+    tasks:
+        stylelint:
+            bin: node_modules/.bin/stylelint
+            triggered_by: [css, scss, sass, less, sss]
+            whitelist_patterns:
+                - /^resources\/css\/(.*)/
+            config: ~
+            config_basedir: ~
+            ignore_path: ~
+            ignore_pattern: ~
+            syntax: ~
+            custom_syntax: ~
+            ignore_disables: ~
+            disable_default_ignores: ~
+            cache: ~
+            cache_location: ~
+            formatter: ~
+            custom_formatter: ~
+            quiet: ~
+            color: ~
+            report_needless_disables: ~
+            report_invalid_scope_disables: ~
+            report_descriptionless_disables: ~
+            max_warnings: ~
+            output_file: ~
+```
+
+**bin**
+
+*Default: null*
+
+The path to your stylelint bin executable. Not necessary if stylelint is in your $PATH. Can be used to specify path to project's stylelint over globally installed stylelint.
+
+> **Note:** It is possible to add `node_modules/.bin` to the grumphp paths in order for it to be automatically discovered as well:
+> 
+> ```yaml
+> grumphp:
+>     environment:
+>         paths:
+>             - node_modules/.bin
+> ```
+
+**triggered_by**
+
+*Default: [css, scss, sass, less, sss]*
+
+This is a list of extensions which will trigger the Sylelint task.
+
+
+**whitelist_patterns**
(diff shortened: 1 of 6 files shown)


yguedidi:
@veewee hello, any missing thing here? if any don't hesitate to tell me!

doc/tasks/stylelint.md
```diff
@@ -0,0 +1,187 @@
...
+            max_warnings: ~
+            output_file: ~
+```
+
+**bin**
+
+*Default: null*
+
+The path to your stylelint bin executable. Not necessary if stylelint is in your $PATH. Can be used to specify path to project's stylelint over globally installed stylelint.
+
```
veewee, line 56: Maybe we could add here that it is possible to add node_modules/.bin to the grumphp paths in order for it to be automatically discovered as well:

```yaml
grumphp:
    environment:
        paths:
            - node_modules/.bin
```

src/Task/Stylelint.php
```diff
@@ -0,0 +1,140 @@
...
+        return ($context instanceof GitPreCommitContext || $context instanceof RunContext);
+    }
+
+    public function run(ContextInterface $context): TaskResultInterface
+    {
+        $config = $this->getConfig()->getOptions();
+
+        $files = $context
+            ->getFiles()
+            ->paths($config['whitelist_patterns'] ?? [])
```
veewee, line 90: maybe better to always make it an array instead of nullable

src/Task/Stylelint.php
```diff
@@ -0,0 +1,140 @@
...
+        $arguments->addOptionalBooleanArgument('--%s', $config['color'], 'color', 'no-color');
+        $arguments->addOptionalArgument('--report-needless-disables', $config['report_needless_disables']);
+        $arguments->addOptionalArgument('--report-invalid-scope-disables', $config['report_invalid_scope_disables']);
+        $arguments->addOptionalArgument(
+            '--report-descriptionless-disables',
+            $config['report_descriptionless_disables']
+        );
+        $arguments->addOptionalIntegerArgument('--max-warnings=%d', $config['max_warnings']);
+        $arguments->addOptionalArgument('--output-file=%s', $config['output_file']);
+        $arguments->addFiles($files);
```
veewee, line 123: Should files always be added? Or should we only add it during pre-commit?
The full list can be quite big, making it crash on windows-like systems which have a smaller max CLI input length

veewee (commented):
Thanks for the PR!
It looks good at first sight. I've added some minor remarks

src/Task/Stylelint.php
yguedidi: should I update the ESlint task too? I think it's a BC break for those who use `null` already

yguedidi:
@veewee updated with fixup commits, tell me if it's OK for you so I squash them before merge

src/Task/Stylelint.php
veewee: For eslint this might a BC break indeed.
We could add a deprecation warning when using the `null` value there:
https://symfony.com/doc/current/components/options_resolver.html#deprecating-the-option

But that seems out of scope for this PR.

src/Task/Stylelint.php
veewee: In that case, it seems like a better solution to use stdin instead?

Example:
https://github.com/phpro/grumphp/blob/master/src/Task/PhpLint.php#L63-L70

You'll also need to provide an stdin for the fix process.

yguedidi: I tried to run `echo "path_to_an_scss_file.scss" | node_modules/.bin/stylelint` and I got a `CssSyntaxError`, so it seems that Stylelint support direct SCSS code as input with stdin, not a list of file like in the example (if I understood correctly).
I'm completely fine to add the list of files only in the precommit context, as it's there where it's the most useful, and for the run context tools like stylelint usually have a config file
is it ok for you?

src/Task/Stylelint.php
yguedidi: I created https://github.com/phpro/grumphp/issues/911 so this is not forgotten

src/Task/Stylelint.php
yguedidi: @veewee updated with adding files only on precommit

veewee:
Looks good, thanks for the PR and sorry for the long time of silence in between :)

yguedidi:
@veewee it's a pleasure! if I knew you planned to merge it, I would have squashed the fixup commit, for a cleaner git history :)

PR #907, opened 2021-05-25
adds order-by argument to phpunit and phpunitBridge tasks

alejgarciarodriguez:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

# New Task Checklist:

### DIFF
--- a/src/Task/Phpunit.php
+++ b/src/Task/Phpunit.php
@@ -21,12 +21,22 @@ class Phpunit extends AbstractExternalTask
             'testsuite' => null,
             'group' => [],
             'always_execute' => false,
+            'order' => 'default',
         ]);
 
         $resolver->addAllowedTypes('config_file', ['null', 'string']);
         $resolver->addAllowedTypes('testsuite', ['null', 'string']);
         $resolver->addAllowedTypes('group', ['array']);
         $resolver->addAllowedTypes('always_execute', ['bool']);
+        $resolver->addAllowedValues('order', [
+            'default',
+            'defects',
+            'duration',
+            'no-depends',
+            'random',
+            'reverse',
+            'size',
+        ]);
 
         return $resolver;
     }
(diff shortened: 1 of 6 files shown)


src/Task/Phpunit.php
```diff
@@ -21,12 +21,22 @@ public static function getConfigurableOptions(): OptionsResolver
             'testsuite' => null,
             'group' => [],
             'always_execute' => false,
+            'order' => 'default',
```
veewee, line 24: This will always set the order to default. What happens if you specified an order in your phpunit task? Will it fall back to that one or use the default of phpunit?
Maybe better to set it to `null` by default so that you can overwrite when it makes sense to do so?

src/Task/Phpunit.php
```diff
@@ -21,12 +21,22 @@ public static function getConfigurableOptions(): OptionsResolver
...
         ]);
 
         $resolver->addAllowedTypes('config_file', ['null', 'string']);
         $resolver->addAllowedTypes('testsuite', ['null', 'string']);
         $resolver->addAllowedTypes('group', ['array']);
         $resolver->addAllowedTypes('always_execute', ['bool']);
+        $resolver->addAllowedValues('order', [
```
veewee, line 31: I am not sure wether to enforce these values from grumphp. I assume that the list will grow in time and needs to keep up-to date.
If phpunit fails on invalid order, it might make sense to let phpunit do the validation instead of grumphp?
That way we don't have to keep this list in sync with phpunit?

veewee (commented):
Thanks for the PR. I've added some questions to the code. Can you take a look at it? Thanks!

src/Task/Phpunit.php
alejgarciarodriguez: done

Issue #919, opened 2021-08-05
TwigCS task does not check the diff, preventing commit on unchanged files


jklmnop:
| Q               | A
| --------------- | ---
| Version         | 1.3.2
| Bug?            | no?
| New feature?    | yes
| Question?       | yes
| Documentation?  | yes/no
| Related tickets | n/a

We have a GrumPHP setup that utilizes the TwigCS task, but the task runs on the entire project instead of just the diff on pre-commit. This means that no commit will be passing until ALL twig issues are fixed.

It works as intended when run manually in a specific folder, though.

Should be noted that we're using lando for this project, but I do not suspect that is part of the issue. All other tasks work fine.

I've created a simple test project to illustrate this issue.

**Steps to reproduce:**

```sh
# Get the test project
$ git clone git@github.com:jklmnop/twigcstest.git

# Install dev dependencies
$ composer install

# Check what's broken on root
$ ./vendor/bin/grumphp run
```

Fix either twig file in `src/templates`, but not both. I fixed `two.twig` by adding `{{ something }}` to the file.

```sh
# Run GrumPHP:
git add -A && git commit -m"Test"
```

**Result:**

You should see TwigCS is unnecessarily checking `one.twig` which is unchanged and prevents the commit.

PR #920, opened 2021-08-06
[919] Add logic to scan the current directory and add any files that …

jklmnop:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #919 

Unfortunately, this will not work until friendsoftwig/twigcs changes how they exclude files, but there is a PR for it.

see: https://github.com/friendsoftwig/twigcs/issues/183
see: https://github.com/friendsoftwig/twigcs/pull/184

This bit of code will compare the files that have changed with the files in the project and add non-matching file paths to the exclude config. 

# New Task Checklist:

### DIFF
--- a/src/Task/TwigCs.php
+++ b/src/Task/TwigCs.php
@@ -57,6 +57,33 @@ class TwigCs extends AbstractExternalTask
         $arguments->addOptionalArgument('--ruleset=%s', $config['ruleset']);
         $arguments->addOptionalArgument('--ansi', true);
 
+        // Get a list of all changed files, and prepare them for comparison.
+        $changedFiles = array_map(function($item) {
+            return preg_quote($item->getPathName(), '/');
+        }, $files->toArray());
+
+        // Regexp for exclude config.
+        $excludePattern = '/('. implode('|', $config['exclude']) .')/';
+        // Regexp for triggered_by config.
+        $extensionPattern = '/^(.*?)\.('. implode('|', $config['triggered_by']) .')$/';
+        // Regexp for all changed files.
+        $changedFilesPattern = '/('. implode('|', $changedFiles) .')$/';
+
+        // Scans entire current directory.
+        $projectFiles = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator('.'));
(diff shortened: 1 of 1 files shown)


veewee:
Since twigcs does not support this yet, wouldn't it make sense to work with an include list instead of an exclude?
That way we could just pass the list of all changed twig files instead of adding some dark magic exclude logic :)

jklmnop:
@veewee Absolutely, but TwigCS does not support StdIn or an "include" list at this moment, so I'm just trying to work with what they do have. 🤦‍♀️

I do not think this is the best solution for the problem, but as of now, I think it's the ONLY solution with the options they provide (and my patch to their project). I don't think it should even make it into master.

At least there is a commit someone can reference with the dependency in composer.json in the event they need this problem solved too. Failing commits for unchanged files is a huge bummer and would mean we couldn't reliably use the tool without a bunch of unnecessary noise and false positive failures and `git commit -n`.

I'm hoping they'll at least change the strategy for exclusion, and better yet, maybe someone much more knowledgable than me could implement a StdIn/include strategy a la PHPCS. Best I could do right now is change which method they're calling to scan for excludes in TwigCS coupled with the "dark magic 🧙‍♂️" in this commit for GrumPHP.

veewee:
Hello @jklmnop 

Thanks for investigating this.
I am closing this PR for now. Feel free to provide a new PR if TwigCs decides to support this and you got a working PR.

jklmnop:
Thank you so much! Eventually, I realized that a patch might be a better option since I was losing out on any updates to master without maintaining this PR. (Here's [that patch](https://gist.githubusercontent.com/jklmnop/8b35c6428e2ffdb16892c1fcccd8848e/raw/585f4abd62738c0f371bc9349ffb23a0aeb5d928/grumphp-twigcs-git-diff-list.patch) for anyone who finds themselves here)

I might take a crack at making TwigCs accept a piped list of files a la PHP_CodeSniffer and other projects. Hoping if they do eventually support it, there won't be a need to update GrumPHP!

PR #925, opened 2021-09-10
fix(docs): commit message yml sample

sr2ds:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

This is PR to fix documentation sample to git commit message hooks with type scope correct works.

In the actual yml sample, don't work rules, is need change specification as I do.


veewee:
Thanks!

PR #928, opened 2021-09-22
Insufficient condition to switch interfaces for event interface.

ansergeyg:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

Latest version of grumphp uses a workaround to switch Event class implementation depending on the symfony component's versions (for example: event-dispatcher and dependency-injection component versions).

But it doesn't consider the case when, some other contributed components use this new component **directly**: [Event-dispatcher-contracts:](https://github.com/symfony/event-dispatcher-contracts)

Symfony event-dispatcher component uses it starting from the version ^4.3. 
Now, imagine that some other component uses symfony event dispatcher ^3.4 and installs another contributed component that directly uses https://github.com/symfony/event-dispatcher-contracts). It means that now the event dispatcher will be an old one, but the grumphp latest version will bring this workaround and the Event class will be switched to the `SymfonyEventContract`  and the the conflict occurs.

This PR fixes this issue.

# New Task Checklist:

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -6,9 +6,11 @@ namespace GrumPHP\Event;
 
 use Symfony\Contracts\EventDispatcher\Event as SymfonyEventContract;
 use Symfony\Component\EventDispatcher\Event as SymfonyLegacyEvent;
-
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 // @codingStandardsIgnoreStart
-if (class_exists(SymfonyEventContract::class)) {
+if (class_exists(SymfonyEventContract::class) &&
+    in_array('EventDispatcherInterface', class_implements(EventDispatcher::class))) {
     class Event extends SymfonyEventContract
     {
     }


veewee:
Do you have an example configuration in which it breaks?
I am not sure I am following the explanation above.

This package requires:

```
        "symfony/event-dispatcher": "~4.4 || ~5.0",
```

Meaning you will get conflicts if another package requires 3.4 right?


If the statements about the versions above are correct, we might as well drop the legacy event fallback all together?

ansergeyg:
Hi @veewee ,
Sorry it's not the latest version of grumphp, but v0.17.2:

https://github.com/phpro/grumphp/blob/ecfc01920fb5733193a41da292e8abb702b6d55d/composer.json#L27

Sure, if you can drop it, it also works, but if clients use symfony components ^3.4, it means that they are not ready yet to upgrade to symfony ^4.3, but they still might want to use grump.

Regards, Sergey

veewee:
Hello @ansergeyg,

Sorry, We are not actively supporting the 0.X branch anymore.
If you want to avoid having old dependency conflicts, you might want to use our dependency-less version:
https://github.com/phpro/grumphp-shim

veewee:
I think it is safest to keep this event fallback system until we drop support for Symfony 4.4 next year.

Issue #926, opened 2021-09-14
stylelint task only adds files in GitPreCommitContext


leonexcc:
Is there a reason this was implemented like this:
https://github.com/phpro/grumphp/blob/4248487a8f34188aec5d48b767e02ccea432a028/src/Task/Stylelint.php#L122-L127

It will produce an error if i run grumphp run outside of a git commit (since stylelint need files to run on).

I comparison the eslint task doesn't do this and will always add the files:
https://github.com/phpro/grumphp/blob/4248487a8f34188aec5d48b767e02ccea432a028/src/Task/ESLint.php#L80-L82

leonexcc:
I just saw this was discussed in https://github.com/phpro/grumphp/pull/904#discussion_r646086110. But i don't understand how stylelint would work without files:

```
  Usage: stylelint [input] [options]

  Input: Files(s), glob(s), or nothing to use stdin.

    If an input argument is wrapped in quotation marks, it will be passed to
    globby for cross-platform glob support. node_modules are always ignored.
    You can also pass no input and use stdin, instead.
```

I also didn't find any possibility to define files or folders in the .stylelintrc file

veewee:
I'm not using the tool myself. Maybe @yguedidi can elaborate?

leonexcc:
We made a copy of the task with the same behaviour as in the ESLint task and it works for us fine. If you want, i can make a pull request with the change.

There is probably a problem with projectes with many css or scss files. But you have probably the same problem with eslint.

veewee:
Feel free to suggest a PR. We can continue the discussion from that point in.

leonexcc:
Hello @yguedidi, i think a `path` option would be great.

PR #932, opened 2021-09-28
Always call addFiles in Stylelint task (like in ESLint)

leonexcc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #926 

The current Stylelint task does not add files to the command if run outside of a git commit. This results in an error because stylelint expects a list of files or a glob.

This pull request changes the behaviour of Stylelint task to the one of ESLint, to fix the error.

There was some discussion about the current implementation in https://github.com/phpro/grumphp/pull/904#discussion_r646086110. But i couldn't find a way to specify files or a globs in the stylelint config file.


veewee:
Taken a look at the cli options of stylelint and agree that always adding the files is the best option for the tool at this moment.
Thanks for pointing it out!

Issue #944, opened 2021-11-04
Add support to specify --no-check-lock option for composer_normalize task


RoSk0:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 1.5.0`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | #595 , #596 , #645 

composer-normalize introduced `--no-check-lock ` option in [2.7.0](https://github.com/ergebnis/composer-normalize/releases/tag/2.7.0) release back in August  2020. 

This option is very useful when you work on a  library project and you only concerned about the `composer.json` file but not that much about the lock file that only exists in the development environment for the library itself.

It would be awesome to see it available via configuration for the `composer_normalize` task.

PR #645, opened 2019-06-03
Add composer normalize task

Landerstraeten:

PR #596, opened 2019-02-16
Add task composer normalize

drupol:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
# New Task Checklist:

- [x] Is the README.md file updated?
- [x] Are the dependencies added to the composer.json suggestions?
- [x] Is the doc/tasks.md file updated?
- [x] Are the task parameters documented?
- [x] Is the task registered in the tasks.yml file?
- [x] Does the task contains phpspec tests?
- [x] Is the configuration having logical allowed types?
- [x] Does the task run in the correct context?
- [x] Is the `run()` method readable?
- [x] Is the `run()` method using the configuration correctly?
- [x] Are all CI services returning green?

Issue #595, opened 2019-02-16
Add a new tasks "composer normalize" using package from localheinz/composer-normalize


drupol:
The goal of this issue is to track the development of a new Grumphp task that aim to provide a new "task" using the package [localheinz/composer-normalize](https://packagist.org/packages/localheinz/composer-normalize).

| Q               | A
| --------------- | ---
| Version         | `grumphp -V`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no

Ping @localheinz

veewee:
Will be handled in the PR

PR #947, opened 2021-11-04
Add --no-check-lock option support

RoSk0:
| Q               | A
| --------------- | ---
| Version         | `GrumPHP 1.5.0`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | #595 , #596 , #645 , #944, #945

composer-normalize introduced `--no-check-lock ` option in [2.7.0](https://github.com/ergebnis/composer-normalize/releases/tag/2.7.0) release back in August  2020. 

This option is very useful when you work on a  library project and you only concerned about the `composer.json` file but not that much about the lock file that only exists in the development environment for the library itself.

It would be awesome to see it available via configuration for the `composer_normalize` task.

### DIFF
--- a/doc/tasks/composer_normalize.md
+++ b/doc/tasks/composer_normalize.md
@@ -22,6 +22,7 @@ grumphp:
         composer_normalize:
             indent_size: ~
             indent_style: ~
+            no_check_lock: false
             no_update_lock: true
             verbose: false
 ```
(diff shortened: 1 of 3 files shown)


RoSk0:
The only thing that is missing is tests - unfortunately I haven't figured out this part. `ComposerNormalizeSpec` added with this task originally gone form the code base...

veewee:
Looks nice! Thanks for the addition.

Just to answer your question : we transitioned the task system at some point in time. 
The spec tests for the tasks weren't really helpful. That is why we added a new way of testing tasks through phpunit.

Looks like you found that during your PR, so that's nice! :)

PR #955, opened 2021-11-24
No vagrant

verbruggenalex:
Just a patch to first check if vagrant is available. Otherwise always use local preset.

### DIFF
--- a/src/Configuration/Model/HooksConfig.php
+++ b/src/Configuration/Model/HooksConfig.php
@@ -41,7 +41,8 @@ class HooksConfig
 
     public function getPreset(): string
     {
-        return $this->preset;
+      exec('vagrant -v > /dev/null 2>&1', $output, $novagrant);
+      return $novagrant ? 'local' : $this->preset;
     }
 
     public function getVariables(): array


verbruggenalex:
I get annoyed by my setup. Team members commit from outside of vagrant box, I commit from inside.
They have default config set in composer.json so I can not use grumphp.yml and grumphp.yml.dist.
This check avoids using vagrant setup when  you don't have vagrant.

veewee:
Sorry, I don't think this is a solution for the problem.
I have vagrant installed but also have other projects that don't use vagrant.

Can you open up a discussion instead in which we can search for a better solution?

You might have to share some more details about the setup though.

verbruggenalex:
Before I open an issue. Can you explain to me what the problem would be if Vagrant was installed and your setup was saying to use the local git hook?

For me there only seems to be a problem if Vagrant is not installed. Because then grumphp does not work for me with install hooks when the type is set to 'vagrant'.

Anyway. I just dropped the PR here to let people know this is a possible solution. I also think the setup where I'm at is wrong because we don't use grumphp.yml and grumphp.yml.dist properly.

veewee:
Looks like I misread the PR a bit yesterday. Thought it worked the other way around.

However it is still possible that someone create their own preset in a an extension or convention checker.
So it  if vagrant is not installed, overwriting the preset internally to local doesnt seem like a good idea.

This preset is coming directly from the config (or from a default value), so it would be strange to swap it without any notice.
Maybe an alternative approach would be to introduce a "conditional preset" system, in which you can specify rules to determine if a preset is valid for use and optionally a way to specify a fallback preset. This system could inform the user about this, so that nothing "strange" happens.

PR #962, opened 2021-12-06
Fix link to composer.json suggest section

paigedwight:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

Fixes a link in the README. Perhaps it would be better to just inline this as a snippet, or make a separate `composer.json.example` (or some such) file that won't get out of sync with the link here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ You will see following message in the composer logs:
 *Watch out! GrumPHP is sniffing your commits!*
 
 GrumPHP will never push you into using a specific task. You can choose the tasks that fit your needs, and activate or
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for the PR!
I'm gonna merge this one as-is. Not sure how we can keep the docs fully in sync with the composer file at this moment.
The file doesn't change that often anymore either. So maybe it isn't that big of an issue.

PR #972, opened 2021-12-24
Add support for DDEV environments.

bserem:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes/no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

This PR adds support for running grumphp in DDEV environments. https://github.com/drud/ddev is a very popular (and rising in popularity) dockerized environment.

I am very new to Grum but I wanted this inside DDEV, so here it is. I'll be glad to hear your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ grumphp:
     git_hook_variables:
          VAGRANT_HOST_DIR: .
          VAGRANT_PROJECT_DIR: /var/www
+         DDEV_HOST_DIR: .
+         DDEV_PROJECT_DIR: /var/www/html
          EXEC_GRUMPHP_COMMAND: exec
          ENV: {}
     stop_on_failure: false
--- a/doc/parameters.md
+++ b/doc/parameters.md
@@ -8,6 +8,8 @@ grumphp:
     git_hook_variables:
         VAGRANT_HOST_DIR: .
         VAGRANT_PROJECT_DIR: /var/www
+        DDEV_HOST_DIR: .
+        DDEV_PROJECT_DIR: /var/www/html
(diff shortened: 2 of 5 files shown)


bserem:
Apparently this is not getting much attention and in the meanwhile I've found an easier way to achieve this. I can alter the `EXEC_GRUMPHP_COMMAND` variable, although this will only make the script work outside of ddev, not both outside and inside the container.

```
grumphp:
  tasks:
    git_commit_message: ~
  git_hook_variables:
    EXEC_GRUMPHP_COMMAND: ddev exec php
```

veewee:
Haven't found the time to look into ddev yet, sorry.
Maybe it can start out as an extension and if there is a lot of traction, it can become GrumPHP core?

bserem:
@veewee no need to be sorry! Grum works fine with ddev, the only issue is that it needs different configuration if you work outside the container and inside the container.
I started from the Vagrant "recipe", but I now feel that the local "recipe" is a better start.

My previous comment is the easiest fix for those who want to work outside the container.

veewee:
@bserem As mentioned above, the suggested way here is to use:

```yaml
grumphp:
  git_hook_variables:
    EXEC_GRUMPHP_COMMAND: ddev exec php
```

We also use `EXEC_GRUMPHP_COMMAND` for running it on docker.
The hook you provided is basically the same as the local hook. 
So it is best to change the exec command instead. That way we dont have to maintain all these different presets.


Feel free to add this specific ddev setup in our docs : https://github.com/phpro/grumphp/blob/master/doc/parameters.md
-> git_hook_variables

Issue #957, opened 2021-11-26
PHP 8.1 compatibility


veewee:
| Q               | A
| --------------- | ---
| Version         | `vendor/bin/grumphp -V`
| Bug?            | no
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets |

We added PHP 8.1 compatibility in #956.
However, not all downstream packages are fully compatible yet.
Therefore, you might still see some deprecation warnings.
The tool itself should be working though!

Awaiting deps:

- [x] opis/closure: https://github.com/opis/closure/pull/105
- [x] opis/closure:  https://github.com/opis/closure/issues/59
- [x] laravel/closure if amphp/parallel-functions wants to use that: https://github.com/amphp/parallel-functions/pull/29

Awaiting dev deps
- [x] phpspec: https://github.com/phpspec/phpspec/pull/1397

Awaiting shim deps (patched locally)

- [x] humbug/box & scoper: https://github.com/box-project/box/issues/562
- [x] scoper issue with serializable closure: https://github.com/humbug/php-scoper/issues/642
- [x] scoper issue with union return types https://github.com/humbug/php-scoper/issues/643
- [x] New shim build setup : https://github.com/phpro/grumphp-shim/pull/17

Broken Tests

- [ ] ~Fix skipped E2E tests~ - currently won't do. They seem buggy edge cases anyways. If there is any need for them to still work, time can be invested

Carpenter0100:
Hi, 
thank you for this.
Is it possible to update "psr/container" too?

dvdknaap:
@veewee please also add psr/container there are packages that can't be upgrade because grumphp is still using v1

veewee:
Idd surely accept a PR for that..

dvdknaap:
> Idd surely accept a PR for that.. care to give it a try?

Please check #968

ctrl-f5:
Are you guys willing to switch to the https://github.com/laravel/serializable-closure fork?

Since I agree with the remarks on opis/closure that depending on FFI is not a very good idea.

veewee:
I am willing to switch to whatever works best.
However, `amphp/parallel-functions` is still using `opis/closure`, so I don't know if we can rule it out completely:

https://github.com/amphp/parallel-functions/blob/master/composer.json#L29

Lets keep an eye on this issue as well:
https://github.com/amphp/parallel-functions/pull/29

tm1000:
[...]

PR #982, opened 2022-02-04
replace opis/closure with laravel/serializable-closure

ctrl-f5:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

replace opis/closure with laralvel/serialzable-closure to support php 8.1

partially solves https://github.com/phpro/grumphp/issues/957

# New Task Checklist:


composer.json
```diff
@@ -22,9 +22,9 @@
         "amphp/parallel-functions": "1.0",
```
veewee, line 22: Maybe best to upgrade amp/paralllel-functions to 1.1 here?
Otherwise opis will still be used ?

ctrl-f5: I had some issues running the tests locally, had the same errors as in the pipelines on my local machine, even on master. So I assumed something was wrong with my machine.

I can try upping `amphp/parallel-functions` to 1.1, but then we're back on the `The given data could not be serialized` error :confused:

veewee: Yes, that's an error we need to get resolved in  this PR. It should work for parallel functions ^1.1

Some of the pipelines were indeed failing earlier today as well.
but in order to get this released in a trustworthy way, all actions should be green again

ctrl-f5:
This issue was preventing us from moving to php8.1, but now we've also started seeing issues with grumphp on 8.0. (the same issue as in the checks in this MR.)

We have decided to just stop using grumphp on all our projects and packages and move the functionality to our `make` workflow, which already handled all other automations.

As a result I will not be putting any more effort in this MR.

veewee:
Sad to hear @ctrl-f5!
Especially since it is all caused by dependencies and is pretty much out of our control :)

Anyways, Good luck with your projects and packages!

Issue #990, opened 2022-03-10
PhpMnd: Remove --non-zero-exit-on-violation option


kukukk:
| Q               | A
| --------------- | ---
| Version         | 1.8.1
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets |

In the master version of PhpMnd the `--non-zero-exit-on-violation` option was removed: https://github.com/povils/phpmnd/commit/028e0e0d1e9ed73d9468b8b724453401e9a7400c

**My configuration**
```yaml
grumphp:
  tasks:
    phpmnd:
      directory: './app'
      whitelist_patterns: []
      exclude: ['Providers']
      exclude_name: []
      exclude_path: []
      extensions: []
      hint: false
      ignore_funcs: []
      ignore_numbers: []
      ignore_strings: []
      strings: false
      triggered_by: ['php']
```

**Steps to reproduce:**
```sh
./vendor/bin/grumphp run
```

**Result:**
```
phpmnd
======

phpmnd version 2.5.0 by Povilas Susinskas

  The "--non-zero-exit-on-violation" option does not exist.
```

veewee:
Thanks for reporting @kukukk.
Care to create a PR for this? It's a quite easy pick.

PR #991, opened 2022-03-11
Remove --non-zero-exit-on-violation option for PhpMnd

kukukk:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #990 

In the master version of PhpMnd the --non-zero-exit-on-violation option was removed. This PR removes this option from the arguments list.


veewee:
@kukukk 
Thanks for the PR.
The only thing missing is an update of the phpmnd tasks unit test.
You can remove the attribute in there as well.

veewee:
Thanks!

PR #995, opened 2022-03-23
Fix broken tests

danepowell:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Deprecations? | yes/no
| Documented?   | yes/no

Tests are broken in HEAD, caused by some Symfony components recently upgrading from 5 to 6.

I fixed the issues in Psalm per advice in these upstream issues:

- https://github.com/vimeo/psalm/issues/7738
- https://github.com/symfony/symfony/issues/45609

The bigger issue is a Symfony EventDispatcher deprecated constructor that apparently went unnoticed and was recently completely removed. I'm unable to grok the documentation on how to fix it:

- https://github.com/symfony/event-dispatcher/commit/0bc2e0d8ba8a6eb353a42d19890f57a3dee410a5
- https://symfony.com/doc/current/components/event_dispatcher.html

Have you ever considered implementing deprecation scanning or committing your composer.lock and using Dependabot or Violinist to update dependencies? This wouldn't avoid these problems, but it would at least catch them earlier and pinpoint the cause without breaking tests in HEAD.

# New Task Checklist:

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -21,4 +21,12 @@
     <stubs>
         <file name="stubs/Amp.php"/>
     </stubs>
+
+    <issueHandlers>
+        <UndefinedDocblockClass>
+            <errorLevel type="suppress">
+                <referencedClass name="UnitEnum" />
+            </errorLevel>
+        </UndefinedDocblockClass>
+    </issueHandlers>
 </psalm>


danepowell:
Sorry, it looks like the errors I mentioned are actually _not_ happening in automated tests. I have no idea why. I get them when running Psalm locally.

The breakage in automated tests must be due to something else. Although I think the issues I mentioned are still valid and _will_ become a problem.

veewee:
Thanks for reporting! Will take care of it. 

Noticed the event dispatcher thing indeed;
It got removed here without any notice https://github.com/symfony/symfony/pull/40468

About the psalm issue : it looks phpunit related, but we are currently not running psalm on the tests. Not sure if it requires fixing.

PR #998, opened 2022-03-31
Infection: Add missing `show_mutations` (incl. verbose mode)

jerowork:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

Infection does have cli options to directly show the escaped mutations on run (apart from the logfiles). 

However, by default, these are disabled, but can be enabled with the option `--show-mutations`. In verbose mode (`-v`), the non-covered escaped mutations will be shown as well.

@see https://infection.github.io/guide/command-line-options.html#show-mutations-or-s

### DIFF
--- a/doc/tasks/infection.md
+++ b/doc/tasks/infection.md
@@ -21,6 +21,7 @@ grumphp:
             test_framework: ~
             only_covered: false
             show_mutations: false
+            verbose: false
             configuration: ~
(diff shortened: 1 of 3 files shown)


veewee:
Looks good, thanks!

PR #1003, opened 2022-04-28
Bugfix/git absolute path

bart-jaskulski:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #936 

This is a follow up to PR which has been stalled: https://github.com/phpro/grumphp/pull/936

I've fixed behavior for git worktree environment as described in the attached PR according to @veewee advices about keeping the code in `GitRepositoryDirLocator` and adding some tests for this.


veewee:
Looks awesome 👍
Thanks for helping out.

PR #1021, opened 2022-06-30
Remove Monolog v3 as it is not currently compatible with GrumPHP

### LINKED ISSUE #1020: PHP8.1: Monolog\Handler\NullHandler::$level must not be accessed before initialization (git_commit_message)
| Q               | A
| --------------- | ---
| Version         | `1.13.0`
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | no

I get the error `Monolog\Handler\NullHandler::$level must not be accessed before initialization` when I try to use the `git_commit_message` check. Other checks pass.

**My configuration**
(PHP 8.1)
```yaml
parameters:
  bin_dir: "./vendor/bin"
  git_dir: "."

grumphp:
  stop_on_failure: true
  ignore_unstaged_changes: false
  tasks:
    git_commit_message:
      enforce_no_subject_trailing_period: true
      enforce_single_lined_subject: true
      max_subject_width: 250
      matchers:
        Must contain issue number: /(XYZ-\d+)|(Merge)/
```

**Steps to reproduce:**
```sh
cd /tmp
mkdir test-grumphp
cd test-grumphp
php -v # use PHP 8.1!
git init
composer init
composer require phpro/grumphp
vim grumphp.yml # use my grumphp.yml
git add .
git commit -m 'XYZ-1 Test'
```

**Result:**
```
git_commit_message
==================

Typed property Monolog\Handler\NullHandler::$level must not be accessed before initialization
```

With GrumPHP 1.12.0 (and therefore Monolog 2 instead of 3) the check works (see #1014).

sippsolutions:
Fixes #1020

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "doctrine/collections": "^1.6.8",
         "gitonomy/gitlib": "^1.3",
         "laravel/serializable-closure": "^1.1",
-        "monolog/monolog": "^2.0 || ^3.0",
+        "monolog/monolog": "^2.0",
         "ondram/ci-detector": "^4.0",
         "psr/container": "^1.1 || ^2.0",
         "seld/jsonlint": "~1.8",


veewee:
Thanks for reporting.
Would you mind fixing the error for both v2 and v3 instead of removing v3?

sippsolutions:
@veewee it seems Monolog v3 only got integrated with no visible reason or testing. This PR fixes the issue so users can use grumphp again with the newest version. If anyone got more time than me feel free to fix the issue with v3 :)

PR #1023, opened 2022-07-06
Add 'cache_file' parameter to 'deptrac' task.

TravisCarden:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

Deptrac provides a `--cache-file` option to control where its cache file will be stored. This PR adds support for it with a new `grumphp.tasks.deptrac.cache_file` option in `grumphp.yml`, i.e.:

```
grumphp:
    tasks:
        deptrac:
            cache_file: ~
```

I updated the corresponding unit test. I didn't see any functional tests, but I manually tested it and it worked great.

### DIFF
--- a/test/Unit/Task/DeptracTest.php
+++ b/test/Unit/Task/DeptracTest.php
@@ -25,6 +25,7 @@ class DeptracTest extends AbstractExternalTaskTestCase
         yield 'defaults' => [
             [],
             [
+                'cache_file' => null,
                 'depfile' => null,
                 'formatter' => null,
                 'output' => null,
--- a/test/Unit/Task/DeptracTest.php
+++ b/test/Unit/Task/DeptracTest.php
@@ -99,6 +99,17 @@ class DeptracTest extends AbstractExternalTaskTestCase
                 'analyse',
             ]
         ];
+        yield 'cache-file' => [
+            [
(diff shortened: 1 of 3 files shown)


test/Unit/Task/DeptracTest.php
```diff
@@ -25,6 +25,7 @@ public function provideConfigurableOptions(): iterable
         yield 'defaults' => [
             [],
             [
+                'cache_file' => null,
```
yguedidi, line 28: I think a test with the value set should be added, in `provideExternalTaskRuns` I think

yguedidi (changes requested):

test/Unit/Task/DeptracTest.php
TravisCarden: Sure! Like this, @yguedidi? https://github.com/phpro/grumphp/pull/1023/commits/eedbe734b5021468d1484ec8e06e921ddffc5106

yguedidi: yes!

veewee:
Looks good, thanks!

PR #1024, opened 2022-07-08
Check if PHP is available before trying to run Grumphp (wip)

drupol:
Hi,

Do you think such minor update in the hooks are justified in Grumphp?

Thanks

| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

# New Task Checklist:

### DIFF
--- a/resources/hooks/local/commit-msg
+++ b/resources/hooks/local/commit-msg
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if ! command -v $(EXEC_GRUMPHP_COMMAND) &> /dev/null
+then
+    echo "PHP could not be found... Skipping commit-msg hook."
+    exit
+fi
+
 #
 # Run the hook command.
 # Note: this will be replaced by the real command during copy.
--- a/resources/hooks/local/pre-commit
+++ b/resources/hooks/local/pre-commit
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if ! command -v $(EXEC_GRUMPHP_COMMAND) &> /dev/null
+then
+    echo "PHP could not be found... Skipping pre-commit hook."
+    exit
+fi
+
 #
 # Run the hook command.
 # Note: this will be replaced by the real command during copy.


resources/hooks/local/commit-msg
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if ! command -v $(EXEC_GRUMPHP_COMMAND) &> /dev/null
```
veewee, line 3: The exec command can also be a command with cli options and attributes.
For example:

```
grumphp:
    git_hook_variables:
        EXEC_GRUMPHP_COMMAND: '/usr/local/bin/php72'
        EXEC_GRUMPHP_COMMAND: 'lando php'
        EXEC_GRUMPHP_COMMAND: 'fin exec php'
        EXEC_GRUMPHP_COMMAND: ['php', '-c /custom/config.ini']
        EXEC_GRUMPHP_COMMAND: ['docker-compose', 'run', '-T', '--rm', '--no-deps', 'php']
        EXEC_GRUMPHP_COMMAND: 'docker run --rm -it -v $(pwd):/grumphp -w /grumphp webdevops/php:alpine'
        EXEC_GRUMPHP_COMMAND: 'ddev exec php'
```

This won't work in that situation

veewee:
Why would you want to skip the hooks in that case?
Doesn't it sound more reasonable to fail so that the root cause can be fixed by the owner of the machine instead?

I don't see a big benefit of adding this check at this moment.  In what situation would you use it?

drupol:
Right, I understand that sometimes this is not really feasible using `command` in bash, I don't see any better option right now.

> I don't see a big benefit of adding this check at this moment. In what situation would you use it?

**Context:** I'm using Nix locally for PHP development.
**Issue:** Sometimes, I need to make a change in a file without entering an environment containing PHP and all my dependencies (`nix shell github:loophp/nix-shell#env-php82-nts --impure`).
When I do that, `git` refuses to commit stuff because the hook is failing, because PHP doesn't exist. This is precisely the reason why I wanted to make this PR.

veewee:
I'dd suggest to either

1. Commit with `-n` flag in those cases
2. Create your own git hook templates

I don't think this is something we should add to this repo nevertheless.

drupol:
You're right, thanks for the enlightenment.

Have a good day !

PR #1025, opened 2022-07-09
Add color on phpcsfixer

oallain:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

Hello, 
A picture is worth a thousand words, I let you see the result, which is so much more pleasant.

Before
before

After
after

The inspiration comes from ECS.
https://github.com/phpro/grumphp/blob/3ec61c1678c4c370f02b05fef606fd561d923c8e/src/Task/Ecs.php#L71

### DIFF
--- a/src/Task/PhpCsFixer.php
+++ b/src/Task/PhpCsFixer.php
@@ -28,6 +28,7 @@ class PhpCsFixer extends AbstractExternalTask
             'allow_risky' => null,
             'cache_file' => null,
             'config' => null,
+            'format' => 'txt',
             'rules' => [],
             'using_cache' => null,
             'config_contains_finder' => true,
@@ -39,6 +40,7 @@ class PhpCsFixer extends AbstractExternalTask
         $resolver->addAllowedTypes('allow_risky', ['null', 'bool']);
         $resolver->addAllowedTypes('cache_file', ['null', 'string']);
         $resolver->addAllowedTypes('config', ['null', 'string']);
+        $resolver->addAllowedTypes('format', ['null', 'string']);
         $resolver->addAllowedTypes('rules', ['array']);
         $resolver->addAllowedTypes('using_cache', ['null', 'bool']);
         $resolver->addAllowedTypes('config_contains_finder', ['bool']);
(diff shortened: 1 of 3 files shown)


src/Task/PhpCsFixer.php
```diff
@@ -71,11 +71,12 @@ public function run(ContextInterface $context): TaskResultInterface
         $this->formatter->resetCounter();
 
         $arguments = $this->processBuilder->createArgumentsForCommand('php-cs-fixer');
-        $arguments->add('--format=json');
+        $arguments->add('--format=txt');
```
yguedidi, line 74: I think this task is using a dedicated formatter (GrumPHP\Formatter\PhpCsFixerFormatter), that this change will break

yguedidi (changes requested):

src/Task/PhpCsFixer.php
veewee: Indeed, this will bypass the custom formatter.
However, the formatter is quite old and was used to propose a way of fixing the code.
Now that we have our fixer configuration - which used the process to build de fixer process - it seems safe to nuke that custom formatter. It doesn't really provide any benefits anymore + it seems better to rely on the official php-cs-fixer formatters.

@oallain : Can you:

- remove the custom formatter and it's implementation
- provide a configurable `format` string option with a default of `txt`
- provide a configurable `ansii` boolean option with a default of `true`.
- add documentation for these options?

oallain:
Hello @veewee 

I tried to do what you asked.
Tell me if I've forgotten or misunderstood anything.

As for the "--ansi" yes, if not, it doesn't work.

veewee:
@oallain 

> I tried to do what you asked. Tell me if I've forgotten or misunderstood anything.> 

I probably did not describe it well enough :)
Currently the task relies on `PhpCsFixerFormatter`. This one basically requires the format to be JSON in order to work.
If it's not, it falls back to the raw value - which might work in your `text` case.

However, I don't think we need the custom `PhpCsFixerFormatter` anymore.
That one was introduces to suggest the fix command.
Since we introduced a Fixer based on a `Process` command, I don't think we need that custom formatter anymore.

Instead we could use the raw php-cs-fixer output and print the additionel  `Process` arguments as string in case of failure.

Hope this makes it a bit more clear. Feel free to ask questions if not :)

veewee:

PR #1038, opened 2022-08-26
Arkitect support

JTubex:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #1036

# New Task Checklist:

### DIFF
--- a/src/Task/PhpArkitect.php
+++ b/src/Task/PhpArkitect.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GrumPHP\Task;
+
+use GrumPHP\Runner\TaskResult;
+use GrumPHP\Runner\TaskResultInterface;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * PHPArkitect task.
+ */
+class PhpArkitect extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): OptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'config' => null,
+            "target_php_version" => null,
+            'stop_on_failure' => null,
+        ]);
+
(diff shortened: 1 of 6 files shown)


src/Task/PhpArkitect.php
```diff
@@ -0,0 +1,67 @@
...
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'config' => null,
+            "target_php_version" => null,
+            'stop_on_failure' => null,
+        ]);
+
+        $resolver->addAllowedTypes('config', ['null', 'string']);
+        $resolver->addAllowedTypes('target_php_version', ['null', 'string']);
+        $resolver->addAllowedTypes('stop_on_failure', ['null', 'boolean']);
```
veewee, line 30: Maybe better to default to false and only accept bools?

src/Task/PhpArkitect.php
```diff
@@ -0,0 +1,67 @@
...
+    public function run(ContextInterface $context): TaskResultInterface
+    {
+        $config = $this->getConfig()->getOptions();
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('phparkitect');
+        $arguments->add('check');
+        $arguments->addOptionalArgument('--config=%s', $config['config']);
+        $arguments->addOptionalArgument('--target-php-version=%s', $config['target_php_version']);
+        $arguments->addOptionalArgument('--stop-on-failure', $config['stop_on_failure']);
+        $arguments->add('--no-ansi');
```
veewee, line 55: Would it make sense not to provide this argument and let the system figure out if it needs to use ansi or not?
(the tool is based on symfony/console, so it should be able to detect and display ansi as wel)

veewee (commented):
Thanks for the PR. Added some small things to the review.

src/Task/PhpArkitect.php
JTubex: I based this class on the PhpStan task. The `--no-ansi` is added there by default. I wasn't aware symfony/console detects this.

veewee:
Nice! Thanks.

JTubex:
I was checking the MR again and forgot to update the default value in the docs 🙈

Created #1039

PR #1039, opened 2022-08-26
Update phpArkitect default value for stop_on_failure to false

JTubex:
Follow-up of #1038

the default value of the stop_on_failure option is no longer null, but false.


veewee:
Woops - missed that one as well :)
Thanks!

PR #1047, opened 2022-09-28
Update ParaTest documentation

DennisdeBest:
The debugger option [got removed](https://github.com/phpro/grumphp/commit/ac14a1d8d12d4ad953764bf6dd3e9a6e10510677) 

However the option is still in the documentation


veewee:
Thanks

PR #1053, opened 2022-11-25
Support PHP 82 - Drop PHP 74

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

Following the PHP cycle:
https://www.php.net/supported-versions.php

Following the Symfony cycle:
https://symfony.com/releases


pelletiermaxime:
Is there a reason to drop support for 7.4 in a minor release ? The package uses symfony 5.4 components and in the linked page we see that symfony 5.4 supports php >= 7.2.5, so I am not sure why 7.4 would be dropped in grumphp. The supported php versions is irrelevant as most distribution will keep supporting 7.4 for 10+ years.

veewee:
Hello @pelletiermaxime,

PHP 74 will not be supported anymore by PHP internals in 3 days:
https://www.php.net/supported-versions.php

Both Symfon, PHP Internals and *most distributions* have a major amount of people working on that. I am pretty much maintaining this package on my own and don't have room for dealing with more than **3 active PHP versions** and **2 major Symfony versions**.

You can  use an older version of GrumPHP on 74 for the next 10+ years, but I am not actively planning on maintaining that version. Don't worry: GrumPHP is very stable at this moment. You only won't receive any new goods on PH 74.

Since nothing has changed in the API of GrumPHP, this is not a breaking change and a minor release is OK. You just won't be able to fetch the latest version because of your individual PHP constraints.

PR #1054, opened 2022-11-25
Add Rector Task

alessandroaussems:
| Q             | A
| ------------- | ---
| Branch        | rector
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

This PR adds a task to run Rector with grumphp

# New Task Checklist:

### DIFF
--- a/src/Task/Rector.php
+++ b/src/Task/Rector.php
@@ -22,17 +22,17 @@ class Rector extends AbstractExternalTask
             'config' => 'rector.php',
             'triggered_by' => ['php'],
             'ignore_patterns' => [],
-            'clear-cache' => true,
-            'no-progress-bar' => true,
-            'no-diffs' => false,
+            'clear_cache' => true,
+            'no_progress_bar' => true,
+            'no_diffs' => false,
         ]);
 
         $resolver->addAllowedTypes('config', ['null', 'string']);
         $resolver->addAllowedTypes('triggered_by', ['array']);
         $resolver->addAllowedTypes('ignore_patterns', ['array']);
-        $resolver->addAllowedTypes('clear-cache', ['bool']);
-        $resolver->addAllowedTypes('no-progress-bar', ['bool']);
-        $resolver->addAllowedTypes('no-diffs', ['bool']);
+        $resolver->addAllowedTypes('clear_cache', ['bool']);
+        $resolver->addAllowedTypes('no_progress_bar', ['bool']);
+        $resolver->addAllowedTypes('no_diffs', ['bool']);
 
         return $resolver;
     }
@@ -61,9 +61,9 @@ class Rector extends AbstractExternalTask
         $arguments->add('--ansi');
 
         $arguments->addOptionalArgument('--config=%s', $config['config']);
-        $arguments->addOptionalArgument('--clear-cache', $config['clear-cache']);
-        $arguments->addOptionalArgument('--no-progress-bar', $config['no-progress-bar']);
-        $arguments->addOptionalArgument('--no-diffs', $config['no-diffs']);
+        $arguments->addOptionalArgument('--clear-cache', $config['clear_cache']);
+        $arguments->addOptionalArgument('--no-progress-bar', $config['no_progress_bar']);
+        $arguments->addOptionalArgument('--no-diffs', $config['no_diffs']);
 
         if ($context instanceof GitPreCommitContext) {
             $arguments->addFiles($files);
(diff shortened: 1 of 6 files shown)


src/Task/Rector.php
```diff
@@ -0,0 +1,89 @@
...
+class Rector extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): OptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'config' => 'rector.php',
+            'triggered_by' => ['php'],
+            'ignore_patterns' => [],
+            'clear-cache' => true,
```
veewee, line 25: grumphp uses snake_case for all its options.
This makes the configuration of multiple tasks more consistent

src/Task/Rector.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('rector');
+        $arguments->add('process');
+        $arguments->add('--dry-run');
+        $arguments->add('--ansi');
+
+        $arguments->addOptionalArgument('--config=%s', $config['config']);
+        $arguments->addOptionalArgument('--clear-cache', $config['clear-cache']);
+        $arguments->addOptionalArgument('--no-progress-bar', $config['no-progress-bar']);
```
veewee, line 65: Not sure if it makes sense to specify if you want a progress bar or not.
Since this progress is not being displayed to the end user, it makes more sense to me to hardcode it in the arguments rather than making it configurable. Wdyt?

src/Task/Rector.php
```diff
@@ -0,0 +1,89 @@
...
+            ->extensions($config['triggered_by']);
+
+        if (0 === \count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('rector');
+        $arguments->add('process');
+        $arguments->add('--dry-run');
+        $arguments->add('--ansi');
```
veewee, line 61: You should be able to drop --ansi. The internal code is smart enough to decide to enable it or not.

src/Task/Rector.php
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        $process = $this->processBuilder->buildProcess($arguments);
+        $process->run();
+
+        if (!$process->isSuccessful()) {
+            return FixableProcessResultProvider::provide(
+                TaskResult::createFailed($this, $context, $this->formatter->format($process)),
+                function () use ($arguments): Process {
+                    $arguments->add('--clear-cache');
```
veewee, line 79: Any specific reason why you want to add --clear-cache possibly twice?
Wouldn't it make sense to use the cache here, since you just ran it without or decided to run without cache to start with?

src/Task/Rector.php
```diff
@@ -0,0 +1,89 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Process\Process;
+
+class Rector extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): OptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'config' => 'rector.php',
```
veewee, line 22: Is config required by rector? If it automatically detects the config file, it might make more sense to make this null?

veewee (changes requested):
Thanks for your work!
I provided some pointers to make it even more awesome. :)

src/Task/Rector.php
alessandroaussems: If i drop the --ansi the output looks like: https://shot.alssndro.be/xjTXzC with the flag the output looks like: https://shot.alssndro.be/QW1th7 which is IMO a lot nicer.

src/Task/Rector.php
alessandroaussems: Refactored the options to use snake_case 🐍

src/Task/Rector.php
alessandroaussems: Since the option is offered by Rector I added it to the list. I wanted to give the end user as much customization options as possible but maybe this one doesn't make any sense indeed. For now I removed it.

src/Task/Rector.php
alessandroaussems: Rector does indeed automatically use rector.php if you don't specify the config file. I made this null by default.

src/Task/Rector.php
alessandroaussems: Indeed. Removed the option when fixing!

alessandroaussems:
@veewee I processed & answered your remarks!

veewee (approved):
Looks good! Thanks!

PR #1059, opened 2022-12-16
Adding PHPCS specific rule (-s) to display which rule triggered the e…

MeCapron:
### Summary

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | **yes**
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

This PR adds the `-s` which displays the rule that triggered the error when running phpcs. 

As an example : 

image

### Configuration

The configuration is set to `true` by default so that developers can more easily find the rule that blocks the validation of `phpcs`.

As `s` was not clear enough, the rule was renamed to `show_sniffs_error_path` inside `grumphp` configuration.

### Reference

The reference of the `phpcs` configuration can be found here : https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#printing-full-and-summary-reports

### DIFF
--- a/doc/tasks/phpcs.md
+++ b/doc/tasks/phpcs.md
@@ -30,7 +30,7 @@ grumphp:
             sniffs: []
             triggered_by: [php]
             exclude: []
-            
+            show_sniffs_error_path: true
 ```
 
 **standard**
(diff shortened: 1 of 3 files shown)


veewee:
Looks good, thanks!

Issue #1069, opened 2023-02-17
Support grumphp.dist.yml


jreklund:
| Q               | A
| --------------- | ---
| Version         | `1.15.0`
| Bug?            | no
| New feature?    | yes
| Question?       | yes
| Documentation?  | yes
| Related tickets | 

Before I start coding a solution to add support for `grumphp.dist.yml` - is this something you want to support?
Currently I'm in the works of changing all configurations to end with the actual file extension e.g. `yml` to simplify our `.editorconfig` and to get syntax highlight (without modifications).

Currently we need to add both `yml` and `yml.dist` as an supported file extension. It works, but without any syntax highligt.
```
# YAML
[*.{yml,yml.dist}]
indent_size = 2
```

`grumphp.yml`

```yaml
grumphp:
  ascii: ~
  environment:
    paths:
      - node_modules/.bin
  tasks:
    phpcsfixer:
    phpstan:
      memory_limit: "-1"
    stylelint:
      triggered_by: [scss]
    eslint:
  fixer:
      fix_by_default: true
```
`grumphp.yml.dist`

```
grumphp:
  ascii: ~
  environment:
    paths:
      - node_modules/.bin
  tasks:
    phpcsfixer:
    phpstan:
      memory_limit: "-1"
    stylelint:
      triggered_by: [scss]
    eslint:
  fixer:
      fix_by_default: true
```

---

I can see your projects `grumphp.yml.dist` uses 2 spaces for indention but according to your `.editorconfig` you let the users editor decide the number of spaces. My editor will break yours as it defaults to 4 spaces. You also would need two file-extensions in your `.editorconfig` (or default into 2 spaces).

---

Some examples in the documentation uses 2 spaces but most are using 4 spaces. I think that needs to be standardized. Is this something you want help with?

veewee:
Thanks for reporting.
We currently didn't keep that specific issue in mind - but I don't mind having it normalized.
Feel free to provide a PR for that!

On top of the regular `yml` and `yml.dist`, We also allow `.yaml` and `.yaml.dist`.

It might even be better if we added a task that checks the code with the editor config as well. 
That way we can either adjust the code or the editor config accordingly.

jreklund:
I took a quick look at an `editorconfig-checker` and it may fail on some markdown files. Currently there are `yaml`, `json` and `php` tags in the documentation and they have a different indentation style. There are also some paragraphs getting indented with 4 or more spaces.

PR #1071, opened 2023-02-17
Normalize yaml code blocks in doc

jreklund:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #1069 

- [x] Changed all `yaml` indentations to 4 spaces
- [x] Renamed all `yml` to `yaml` code blocks
- [x] Updated .editorconfig
- [x] Added `grumphp.yml` to code blocks starting with `grumphp:`
- [x] Removed `grumphp.yml` from some code blocks to reflect the rest of the doc 
- [x] Fixed "tester" task documentation
- [x] Add missing `:` to parameters doc

One thing I'm not 100% certain about are my `phpcsfixer2` into `phpcsfixer` rename. I _think_ that was a typo so I corrected it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,11 +9,6 @@ end_of_line = lf
 insert_final_newline = true
 
 # 4 space indentation
-[*.php]
+[*.{php,md}]
 indent_style = space
 indent_size = 4
-
-# 2 space indentation
-[*.md]
-indent_style = space
-indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,3 +12,8 @@ insert_final_newline = true
 [*.php]
 indent_style = space
 indent_size = 4
+
+# 2 space indentation
+[*.md]
+indent_style = space
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -72,33 +72,33 @@ You can specify a custom config filename and location in `composer.json` or in t
 ```yaml
 # grumphp.yml
 grumphp:
-    hooks_dir: ~
-    hooks_preset: local
-    git_hook_variables:
-         VAGRANT_HOST_DIR: .
-         VAGRANT_PROJECT_DIR: /var/www
-         EXEC_GRUMPHP_COMMAND: exec
-         ENV: {}
-    stop_on_failure: false
-    ignore_unstaged_changes: false
-    hide_circumvention_tip: false
-    process_timeout: 60
-    ascii:
-        failed: grumphp-grumpy.txt
-        succeeded: grumphp-happy.txt
-    parallel:
-        enabled: true
-        max_workers: 32
-    fixer:
-        enabled: true
-        fix_by_default: false
-    environment:
-        files: []
-        variables: {}
-        paths: []
-    tasks: {}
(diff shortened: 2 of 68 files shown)


.editorconfig
```diff
@@ -12,3 +12,8 @@ insert_final_newline = true
 [*.php]
 indent_style = space
 indent_size = 4
+
+# 2 space indentation
```
veewee, line 16: This applies to all code-blocks in markdown files or is there a way to tell editorconfig to lookup the corresponding type instead? For example: php code blocks -> look at *.php config?

veewee:
Looks good! I've got one question about the editor config.

> One thing I'm not 100% certain about are my phpcsfixer2 into phpcsfixer rename. I think that was a typo so I corrected it.

That is correct! Previously we had both phpcsfixer(1) and phpcsfixer2 tasks. Now that v1 is deprecated, we removed it and renamed it to phpcsfixer.

.editorconfig
jreklund: It applies to the whole markdown document and you cant have mixed values. So you need to either default to 2 or 4.

2 = You will need to double indent php and other types of code blocks. Also affects the number of indents needed to create "git_hook-variables" in the doc/parameters.

4 = You need to press space two times for all yaml code blocks and tab (4 spaces) for the rest.

Currently there are more 2 space indentation than 4, but there are no perfect solution. :(

veewee: I'm a bit confused by this explanation to be honest.

Would it make sense to set the defaults to always be 4 spaces as indentation?

That way it would be compatible with how php is written (psr-12) and with the defaults of the symfony yaml dumper: https://github.com/symfony/yaml/blob/6.2/Dumper.php#L30
JSON would also be indented by 4 spaces.

That way, we might be able to automatically run an editor config checker on this project?

jreklund: Sorry about that! If you want to automatically correct it using .editorconfig it needs to be set to 4 or PHP-code blocks won't match PSR-12. The official specification of YAML specify `zero or more` for their [indentation](https://yaml.org/spec/1.2.2/#61-indentation-spaces). Both `2` or `4` are correct. Most .yml files I have come across uses 2 spaces, like the one used in this project.  

The debate on how many spaces you should indent your code with are "tab vs spaces" once again. `package.json` are a valid JSON and that one are 2 spaces by default.

veewee: OK, thanks for clarifying.

I'dd say lets go for 4 spaces then to conclude this debate.

Reasoning:

* .php 4 by PSR-12
* .yaml 4 is the symfony convention
* composer.json -> 4 by default

Sounds like the most logical and straight forward choice.

veewee:
Thanks @jreklund . Looks good to me!
Or can you think of anything else that needs to be done in thisPR?

jreklund:
I'm finished on my end. Better to merge it so that other features don't end up with merge conflict for holding this one.

veewee:
Thanks again! :)

PR #1074, opened 2023-03-21
Add grouping of output in GitHub Actions

jackbentley:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | ?
| Deprecations? | no
| Documented?   | ?
| Fixed tickets |

This will add [groups](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) to the output when running in GitHub Actions.

### DIFF
--- a/resources/config/services.yml
+++ b/resources/config/services.yml
@@ -37,15 +37,16 @@ services:
                 logger: '@grumphp.logger'
                 inherit_environment_variables: true
 
-    grumphp.io:
-        alias: grumphp.io.console
-
-    grumphp.io.console:
-        class: GrumPHP\IO\ConsoleIO
+    GrumPHP\IO\IOFactory:
         arguments:
-          - '@console.input'
-          - '@console.output'
-        public: true
+            - '@OndraM\CiDetector\CiDetector'
+
+    grumphp.io:
+        class: GrumPHP\IO\IOInterface
+        factory: ['@GrumPHP\IO\IOFactory', 'create']
+        arguments:
+            - '@console.input'
+            - '@console.output'
 
     grumphp.logger:
         class: Monolog\Logger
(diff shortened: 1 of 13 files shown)


veewee:
Hello,

Thanks for the PR - it looks quite nice at first sight - yet it seems to fail CI. 
Can you take a look at whats going wrong?

jackbentley:
Should be good now. I was passing a string instead of array to `->write` in the GitHub Actions specific class

resources/config/services.yml
```diff
@@ -37,15 +37,16 @@ services:
...
-        alias: grumphp.io.console
+    GrumPHP\IO\IOFactory:
+        arguments:
+            - '@OndraM\CiDetector\CiDetector'
 
-    grumphp.io.console:
-        class: GrumPHP\IO\ConsoleIO
+    grumphp.io:
+        class: GrumPHP\IO\IOInterface
+        factory: ['@GrumPHP\IO\IOFactory', 'create']
```
veewee, line 46: Since this service is being build from the factory - wouldnt it make sense to inject the `IOInterface` directly instead of the IOFactory?

veewee (changes requested):
Cool feature, thanks!
Could you add some additional tests in the newly added IO classes?

veewee:
@jackbentley It's a nice feature... Care to finish up this PR? :)

resources/config/services.yml
jackbentley: Suppose we could 🤔 I think the reason I injected the factory is because the command previously built the class it's self using the provided `$input, $output` args

veewee (approved):

veewee:
Looks good, thanks :)

PR #1076, opened 2023-03-22
Ignore untracked files with ignore_unstaged_changes parameter

### LINKED ISSUE #1075: Allow to also ignore untracked files
| Q               | A
| --------------- | ---
| Version         | 1.15.0
| Bug?            | yes/no
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | N/A

I faced an issue for which ignoring unstaged changes was not sufficient: running a unit test suite as precommit with an untracked unit test file. it wasn't expected to commit that file at that moment, but it existed already.
And so being on the file system it was taken into account by PHPUnit.

I was about to quickly contribute a new `ignore_untracked_files` option, but I realized that it may not be good to have it as a separate configuration, as it would mean that it could be used alone, without ignore_unstaged_changes.

Another solution is to simply add `--ignore-untracked` to the stach command, but maybe this is a BC break?

So I'm questioning what would be the best way to solve the issue I faced.

Note: issue faced when doing lot of code changes then trying to do clean atomic commits 🙂

yguedidi:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Documented?   | yes

closes #1075

### DIFF
--- a/src/Event/Subscriber/StashUnstagedChangesSubscriber.php
+++ b/src/Event/Subscriber/StashUnstagedChangesSubscriber.php
@@ -97,14 +97,32 @@ class StashUnstagedChangesSubscriber implements EventSubscriberInterface
      */
     private function doSaveStash(): void
     {
-        $pending = $this->repository->getWorkingCopy()->getDiffPending();
-        if (!\count($pending->getFiles())) {
+        $hasPending = \count($this->repository->getWorkingCopy()->getDiffPending()->getFiles()) > 0;
+        $hasUntracked = \count($this->repository->getWorkingCopy()->getUntrackedFiles()) > 0;
+
+        if (!$hasPending && !$hasUntracked) {
             return;
         }
 
         try {
-            $this->io->write(['<fg=yellow>Detected unstaged changes... Stashing them!</fg=yellow>']);
-            $this->repository->run('stash', ['save', '--keep-index', uniqid('grumphp')]);
+            if ($hasPending && !$hasUntracked) {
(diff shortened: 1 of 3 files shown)


src/Event/Subscriber/StashUnstagedChangesSubscriber.php
```diff
@@ -97,14 +97,32 @@ public function handleErrors(): void
...
+        $hasUntracked = \count($this->repository->getWorkingCopy()->getUntrackedFiles()) > 0;
+
+        if (!$hasPending && !$hasUntracked) {
             return;
         }
 
         try {
-            $this->io->write(['<fg=yellow>Detected unstaged changes... Stashing them!</fg=yellow>']);
-            $this->repository->run('stash', ['save', '--keep-index', uniqid('grumphp')]);
+            if ($hasPending && !$hasUntracked) {
```
veewee, line 108: Does it make sense to add 3 different messages here?
Would 1 suffice as well?

veewee:
Looks good, added a small question.

src/Event/Subscriber/StashUnstagedChangesSubscriber.php
veewee: Ok, lets keep it like this then!

veewee:
Thanks!

yguedidi:
You're welcome! Thanks for the quick merge

PR #1080, opened 2023-04-04
Extra bin-dir lookup feature

greew:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | -

The Composer bin-dir path is usually changed by setting the composer config `bin-dir`, but another common way of doing it is, if the vendor directory itself is moved by changing the `vendor-dir` configuration, by which the bin dir is moved also.


veewee:
Thanks for the PR. It looks good, but before I can merge, could you please add a test-case covering this in https://github.com/phpro/grumphp/blob/master/spec/Util/ComposerFileSpec.php ?

src/Util/ComposerFile.php
```diff
@@ -29,6 +29,12 @@ public function getBinDir(): string
         if (null !== $binDir) {
             return (string) $binDir;
         }
+        
+        $binDir = $this->configuration['config']['vendor-dir'] ?? null;
```
veewee, line 33: `$vendorDir` feels more appropriate.

greew:
@veewee Done! :) 

Let me know if you have other requests :)

veewee:
Thanks, that looks nice @greew .
However, now you introduced a second test class for a single unit. There was already a spec test class (the one I linked to in previous comment.)

Would you mind adding to the spec test instead? 
Another option is to move the spec cases to the phpunit testcase.

greew:
I can do either? Which would you prefer?

(And yes, I saw your path to the file, but has never worked with PhpSpec, so didn't grasp it correctly - just thought you wrote the wrong path - sorry about that :smile: )

veewee:
No problem 
Maybe best to move to phpunit then.
We moved away from phpspec at Phpro anyways, so that would make sense.

greew:
There you go! All tests in the specified PhpSpec file has been moved to PhpUnit and the PhpSpec file has been deleted :)

veewee:
Thanks!

PR #1089, opened 2023-05-18
Add support for devenv

sneakyvv:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 

This PR adds support for usage of GrumPHP when developing with a [devenv](https://devenv.sh/) environment. 

## Problem

Devenv basically allows to run different PHP versions per project (amongst other services), but unlike Docker the services run natively.

In my setup, devenv is running in a WSL2 environment (Ubuntu). Doing commits from within WSL2 is not a problem, because direnv is automatically hooking into each bash command so that the correct PHP version is used for example. 

The problem is that my git client (Fork) is running in Windows itself (and is proxying its git commands to WSL2 via [wslgit](https://github.com/andy-5/wslgit)). Therefore when the `pre-commit` hook is running its PHP command, the devenv shell is not loaded, so it's using the wrong PHP version, resulting in Psalm errors (but there might be others of course due to the wrong environment being used).

## Solution

Alter the `pre-commit` hook install script so that the GrumPHP command is running inside the devenv shell

---
PS: I also wondered if the `commit-msg` hook should also be adapted, but since that's only using git afaics it's not necessary.

### DIFF
--- a/resources/hooks/local/pre-commit
+++ b/resources/hooks/local/pre-commit
@@ -12,5 +12,14 @@ DIFF=$(git -c diff.mnemonicprefix=false -c diff.noprefix=false --no-pager diff -
 $(ENV)
 export GRUMPHP_GIT_WORKING_DIR="$(git rev-parse --show-toplevel)"
 
-# Run GrumPHP
-(cd "${HOOK_EXEC_PATH}" && printf "%s\n" "${DIFF}" | $(EXEC_GRUMPHP_COMMAND) $(HOOK_COMMAND) '--skip-success-output')
+# Add path to devenv (if available)
+export PATH="$HOME/.nix-profile/bin:$PATH"
+
+# Check if devenv command is available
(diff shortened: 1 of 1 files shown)


veewee:
Hello,

Thanks for the pr.
I'm wondering why you can't just configure the command to run through devenv by configuring `EXEC_GRUMPHP_COMMAND`?
Similar to configuring docker and other tools:
https://github.com/phpro/grumphp/blob/master/doc/parameters.md

sneakyvv:
Thanks for the pointer. That will probably work! I'll test it asap.

sneakyvv:
For anyone looking, this worked:

```yml
  git_hook_variables:
    EXEC_GRUMPHP_COMMAND: $HOME/.nix-profile/bin/devenv shell
```

PR #1097, opened 2023-07-11
Add new nette tester options

hanisko:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | no

added new nette tester config options
- php_ini_configuration_path [documentation](https://tester.nette.org/en/running-tests#toc-c-path)
- default_php_ini_configuration [documentation](https://tester.nette.org/en/running-tests#toc-c)

### DIFF
--- a/src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
+++ b/src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
@@ -41,10 +41,12 @@ class ForbiddenStaticMethodCallsVisitor extends AbstractVisitor implements Confi
             return;
         }
 
-        $class = implode(
-            '\\',
-            method_exists($node->class, 'getParts') ? $node->class->getParts() : $node->class->parts
-        );
+        /**
+         * https://github.com/nikic/PHP-Parser/releases/tag/v4.16.0
+         * @psalm-suppress DeprecatedProperty
+         */
+        $class = implode('\\', $node->class->parts);
+
         $method = $node->name;
         $normalized = sprintf('%s::%s', $class, $method);
 
--- a/src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
+++ b/src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
@@ -41,7 +41,10 @@ class ForbiddenStaticMethodCallsVisitor extends AbstractVisitor implements Confi
             return;
         }
 
-        $class = implode('\\', $node->class->getParts());
+        $class = implode(
+            '\\',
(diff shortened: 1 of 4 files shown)


veewee:
Thanks for the PR. It looks like you still have a coding-style issue. Can you fix it?

src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
```diff
@@ -37,11 +37,11 @@ public function configure(array $options): void
      */
     public function leaveNode(Node $node): void
     {
-        if (!$node instanceof Node\Expr\StaticCall) {
+        if (!$node instanceof Node\Expr\StaticCall || !$node->class instanceof Node\Name) {
             return;
         }
 
-        $class = implode('\\', $node->class->parts);
+        $class = implode('\\', $node->class->getParts());
```
veewee: Thanks for looking into this as well!

This one seems to be the party pooper:
https://github.com/nikic/PHP-Parser/releases/tag/v4.16.0

Maybe a `method_exists` check could temper psalm on older versions of php-parser?

hanisko: wouldn't it be better solution to update the version of nikic/php-parser in composer instead of making exceptions in code?

from "nikic/php-parser": "~4.13" to "nikic/php-parser": "^4.16"

veewee: The problem is that the php-parser currently is a dev dependency (as it is opt-inable as task) and that by just bumping it, it might cause exceptions for users that are using an older version of php-parser.

Ideally, we should either extract the parser task into either a new extension (with it's own version requirements) or deprecate it completely.
Nowadays with psalm and phpstan, writing an additional extension for that is probably a better way then writing your own phpparser visitors. The visitors we provide can already be achieved by those package iirc.

src/Parser/Php/Visitor/ForbiddenStaticMethodCallsVisitor.php
```diff
@@ -37,11 +37,16 @@ public function configure(array $options): void
...
     public function leaveNode(Node $node): void
     {
-        if (!$node instanceof Node\Expr\StaticCall) {
+        if (!$node instanceof Node\Expr\StaticCall || !$node->class instanceof Node\Name) {
             return;
         }
 
+        /**
+         * https://github.com/nikic/PHP-Parser/releases/tag/v4.16.0
+         * @psalm-suppress DeprecatedProperty
```
veewee, line 46: lets do it like that for now.

veewee:
Thanks!

Issue #1099, opened 2023-07-18
Don't do commit message checks on merges?


delboy1978uk:
| Q               | A
| --------------- | ---
| Version         | latest
| Bug?            | no
| New feature?    | no
| Question?       | yes
| Documentation?  | no
| Related tickets | 

Is there a way of NOT doing the Git Commit Message checks when doing a merge? It works great for us on our ticket branches, and we also have a hook adding the ticket number automatically from the branch name, but when we are on develop or master, there is no ticket number, so when we merge it doesn't like it!

veewee:
That could be a nice addition to the task.
There is some sense of merge commit detection already, but we could add a flag to skip the task on merge commits.
Care to give it a try?

delboy1978uk:
@veewee sure, can you point me to the code I should be looking at?

veewee:
The code for this specific task is located here:
https://github.com/phpro/grumphp/blob/v2.x/src/Task/Git/CommitMessage.php

Currently, there is a merge matching mechanism for skipping validation on scope conventions:
https://github.com/phpro/grumphp/blob/d0e23de45dbea092b244894cf3ee97844c0b7256/src/Task/Git/CommitMessage.php#L366-L384

It might make sense to extract these merge patterns into a separate function so that it can be used to determine if the task needs to run at all. You could use that same check to skip running the scope conventions.

That specific task is becoming a bit messy because of historical reasons and is open for improvements  :)

delboy1978uk:
Thanks @veewee I'll have a look! 
BTW I see you're in Belgium, me too! (ik woon al 7 jaar in Leuven)

veewee:
Nice :) (Ik ben van de Kempen!)

delboy1978uk:
I'm now able to merge without a JIRA number on a merge! But let me know if I should move that is merge commit check further up or down the chain of events, I allowed it to still go through the first rules, like not empty message etc. Also, you mentioned a flag setting so we can have the merge skip enabled or disabled, how/where do we go about adding the setting?

delboy1978uk:
Of course, I have broken your tests in doing this :P But I'll wait for your feedback first before continuing work!

veewee:
Cool. Let's continue the discussion in the PR.

PR #1100, opened 2023-07-19
skip additional checks if a merge commit

delboy1978uk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | https://github.com/phpro/grumphp/issues/1099 

This PR makes Grump skip commit message checks (beyond the defaults) if it is a merge commit. This stops problems like JIRA ticket numbers not being in the main branches like master and develop.


oallain (approved):

src/Task/Git/CommitMessage.php
```diff
@@ -135,6 +138,9 @@ public function run(ContextInterface $context): TaskResultInterface
             );
         }
 
+        if ($this->isMergeCommit($commitMessage)) {
```
veewee: Would it make sense to add a configuration flag like `skip_on_merge_commit` (default `true`) and do the validation at the top of the task - before any other task runs?

That way people could opt-in on checking commit messages on merges (which might be a bit odd, but someone might use it.) Setting this by default to true is a bc break somehow, but I think it makes total sense to do so. People can easily re-activate it by config.

WDYT?

src/Task/Git/CommitMessage.php
```diff
@@ -135,6 +138,9 @@ public function run(ContextInterface $context): TaskResultInterface
...
+            return TaskResult::createSkipped($this, $context);
+        }
 
         if ($this->enforceTypeScopeConventions()) {
```
veewee: you could do the `isMergeCommit` check for merge commits here and skip the type-scope convetions when it's a merge commit. That could make that conventions logic more straight forward

veewee (commented):
Looks like a good starting point. I've added some additional pointers.

src/Task/Git/CommitMessage.php
delboy1978uk: Sounds like a sensible idea, although wouldn't a default of `false` be better, as then no behaviour will change unless you set it. Although as you say, using it on merges is a bit odd and probably the exception.

src/Task/Git/CommitMessage.php
```diff
@@ -56,6 +60,7 @@ public static function getConfigurableOptions(): OptionsResolver
         $resolver->addAllowedTypes('max_body_width', ['int']);
         $resolver->addAllowedTypes('max_subject_width', ['int']);
         $resolver->addAllowedTypes('case_insensitive', ['bool']);
+        $resolver->addAllowedTypes('skip_on_merge_commit', ['bool']);
```
veewee, line 63: Can you add this one to the documentation of the task?
https://github.com/phpro/grumphp/blob/v2.x/doc/tasks/git_commit_message.md

src/Task/Git/CommitMessage.php
```diff
@@ -135,8 +144,7 @@ public function run(ContextInterface $context): TaskResultInterface
             );
         }
 
-
-        if ($this->enforceTypeScopeConventions()) {
+        if (!$this->isMergeCommit($commitMessage) && $this->enforceTypeScopeConventions()) {
```
veewee: instead of checking this twice, you could as well remember it in a var like `$isMergeCommit`

veewee (commented):
Looks good. Think this is about all the code it requires, nice! :)
Can you take a look at the comments I made and at the failing testsuite? Looks like a missing type in your function signature

veewee:
Thanks for the addition

delboy1978uk:
@veewee Thanks so much for merging this feature! Can I also create a PR for v1.x ?

veewee:
I wasn't really planning of porting new features back to v1 because of the maintenance overhead.

delboy1978uk:
@veewee Understandable, I'll use a fork of my own until the rest of the servers where I'm working are on PHP 8.1 (which shouldn't be long now). I'm looking forward to the release! Thanks again :D

PR #1107, opened 2023-09-13
Add codeception XML/HTML reporting support to grumphp

petrisorciprian-vitals:
| Q             | A
| ------------- | ---
| Branch        | v2.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

Codeception allows generating HTML/XML reports after test runs, which is very useful for CI and/or sharing results.
Currently, this is not possible within grumphp, except through a shell script which invokes codeception manually.

This PR adds the capability for the Codeception task to output XML/HTML reports.

### DIFF
--- a/src/Task/Codeception.php
+++ b/src/Task/Codeception.php
@@ -60,6 +60,8 @@ class Codeception extends AbstractExternalTask
         $arguments->add('run');
         $arguments->addOptionalArgument('--config=%s', $config['config_file']);
         $arguments->addOptionalArgument('--fail-fast', $config['fail_fast']);
+        $arguments->addOptionalArgument('--xml', $config['xml']);
+        $arguments->addOptionalArgument('--html', $config['html']);
         $arguments->addOptionalArgument('%s', $config['suite']);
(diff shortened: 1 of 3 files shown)


src/Task/Codeception.php
```diff
@@ -60,6 +60,8 @@ public function run(ContextInterface $context): TaskResultInterface
         $arguments->add('run');
         $arguments->addOptionalArgument('--config=%s', $config['config_file']);
         $arguments->addOptionalArgument('--fail-fast', $config['fail_fast']);
+        $arguments->addOptionalArgument('--xml', $config['xml']);
```
veewee, line 67: You need to configure these available options in `getConfigurableOptions()` as well.

veewee:
Thanks for the PR.
You are almost there: only the options configuration is missing.

src/Task/Codeception.php
petrisorciprian-vitals: done!

veewee:
Looks good, thanks again! :)

PR #1108, opened 2023-09-14
Prints error message when rejecting commit

EwenQuim:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | what is BC?
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #1106 

Fixes the issue described in #1106 

# New Task Checklist:

- [x] Are the dependencies added to the composer.json suggestions? No
- [ ] Is the doc/tasks.md file updated? N/A
- [ ] Are the task parameters documented? N/A
- [ ] Is the task registered in the tasks.yml file? N/A
- [ ] Does the task contains phpunit tests? N/A
- [ ] Is the configuration having logical allowed types? N/A
- [ ] Does the task run in the correct context? N/A
- [ ] Is the `run()` method readable? N/A
- [ ] Is the `run()` method using the configuration correctly? N/A
- [x] Are all CI services returning green? Yes

### DIFF
--- a/src/Task/Git/CommitMessage.php
+++ b/src/Task/Git/CommitMessage.php
@@ -20,7 +20,7 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 
 class CommitMessage implements TaskInterface
 {
-    const MERGE_COMMIT_REGEX =
+    public const MERGE_COMMIT_REGEX =
         '(Merge branch|tag \'.+\'(?:\s.+)?|Merge remote-tracking branch \'.+\'|Merge pull request #\d+\s.+)';
 
     /**
@@ -93,6 +93,18 @@ class CommitMessage implements TaskInterface
         return $context instanceof GitCommitMsgContext;
     }
 
+    private static function withCommitMessage(string $errorMessage, string $commitMessage): string
+    {
+        return sprintf(
+            "%s%sOriginal commit message: %s%s",
+            $errorMessage,
+            PHP_EOL,
+            PHP_EOL,
+            $commitMessage
+        );
+    }
+
+
     public function run(ContextInterface $context): TaskResultInterface
     {
         assert($context instanceof GitCommitMsgContext);
@@ -110,7 +122,7 @@ class CommitMessage implements TaskInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Commit message should not be empty.'
+                self::withCommitMessage('Commit message should not be empty.', $commitMessage),
             );
         }
 
@@ -118,7 +130,7 @@ class CommitMessage implements TaskInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Subject should start with a capital letter.'
+                self::withCommitMessage('Subject should start with a capital letter.', $commitMessage)
             );
         }
 
@@ -126,7 +138,7 @@ class CommitMessage implements TaskInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Subject should be one line and followed by a blank line.'
+                self::withCommitMessage('Subject should be one line and followed by a blank line.', $commitMessage)
             );
         }
 
@@ -134,7 +146,7 @@ class CommitMessage implements TaskInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Please omit all punctuations from commit message subject.'
+                self::withCommitMessage('Please omit all punctuations from commit message subject.', $commitMessage)
             );
         }
 
@@ -142,7 +154,7 @@ class CommitMessage implements TaskInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Please omit trailing period from commit message subject.'
+                self::withCommitMessage('Please omit trailing period from commit message subject.', $commitMessage)
             );
         }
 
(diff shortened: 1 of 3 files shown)


veewee:
Thanks @EwenQuim 
It seems like the functional e2e tests pass in this pr.
The code looks good, but you need to alter the phpunit testcases for this task to match the new error messages.

EwenQuim:
Hi again @veewee .

It seems that the only difference between my changes and the existing version is that the newLines are replaced by a visible \n. Do you have any idea of what I did wrong ?

Also, I messed up amending my commits. Don't worry, it will be fixed when I'll be able to understand how to fix the tests. The PR should be OK after this, I think.

src/Task/Git/CommitMessage.php
```diff
@@ -93,6 +93,18 @@ public function canRunInContext(ContextInterface $context): bool
         return $context instanceof GitCommitMsgContext;
     }
 
+    private static function withCommitMessage(string $errorMessage, string $commitMessage): string
+    {
+        return sprintf(
+            "%s%sOriginal commit message:%s%s",
```
veewee, line 99: this one has no space after `:`.
This makes the tests fail.

See https://github.com/phpro/grumphp/pull/1108/files#diff-93ecaacb0c8dbc50a120fd5d5f863454fe131909bfe4ed88e4ea085772a9c03dL170

```diff
+ '%s%sOriginal commit message: %s%s',
- "%s%sOriginal commit message:%s%s"
```

src/Task/Git/CommitMessage.php
```diff
@@ -110,39 +122,39 @@ public function run(ContextInterface $context): TaskResultInterface
             return TaskResult::createFailed(
                 $this,
                 $context,
-                'Commit message should not be empty.'
+                self::withCommitMessage('Commit message should not be empty.', $commitMessage),
```
veewee: For every changed error message, you need to change the test-cases as well:
https://github.com/phpro/grumphp/blob/fb70da3dff23f55529ea2c13fd7c00a966278320/test/Unit/Task/Git/CommitMessageTest.php#L86-L94

Here you'll need to make sure the newly added 'original message' is applied to the error message so that it gets tested as well. (you could create a shortcut function for that as well inside that test class.

src/Task/Git/CommitMessage.php
EwenQuim: Wow, well observed

src/Task/Git/CommitMessage.php
EwenQuim: I don't think a shortcut function is a good thing to test another shortcut function, so I just tested with some text on two validations. Tell me what you think about it.

EwenQuim: The tests pass because your test framework only test is the text is contained, I don't know if it's relevant to apply these modifications to all the tests (harder to read), but I'll do it if you think it's important.

Thank you for your review !

test/Unit/Task/Git/CommitMessageTest.php
```diff
@@ -206,6 +219,32 @@ function () {
                 'Line 5 of commit message has > 10 characters.'
             )
         ];
+        yield 'enforce_text_with_regular_gives_back_message' => [
```
veewee, line 214: What about this instead?

It introduces a new method `buildFailureMessage` which takes the original commit message and the lines it should match on? This keeps te tests readable and gives feedback on the intention.

```php
        yield 'enforce_capitalized_subject' => [
            [
                'enforce_single_lined_subject' => false,
                'max_subject_width' => 10,
                'max_body_width' => 10,
            ],
            $this->mockCommitMsgContext($commitMsg = $this->buildMessage(
                'Subject 1234567891011',
                'Body 1234567891011',
                'Body ok',
                'Body 1110987654321'
            )),
            function () {
            },
            $this->buildFailureMessage(
                ['Please keep the subject <= 10 characters.',
                'Line 3 of commit message has > 10 characters.',
                'Line 5 of commit message has > 10 characters.',],
                $commitMsg,
            )
        ];
```

I wouldn't introduce 2 tests covering the capitalized subject just to check the original commit message. We want to test the logic and validate if the error message contains what what went wrong + the original commit message. All in 1 test-case

EwenQuim: As all the cases are already covered, I added two new test cases that overlap with some existing tests (that's why 2 tests cover the capitalized subject and 2 tests cover the subject/body width). We can delete the first one as you suggested, no problem.

Also, what should be the content of `buildFailureMessage`? Exactly the same as `withCommitMessage` that we defined above, doesn't it ? Doing so, the only thing we're testing is that the function is called. That's why I tested against a full string.

veewee: Yes, it would have similar contents to the `withCommitMessage`.
It would validate that the error message contains the expected original commit message.
This needs to be done at alle places where it is added, otherwise one could remove the original commit message without the tests resulting in failure.

EwenQuim: Added

veewee:
Thanks again! :)

EwenQuim:
Thank you too!

When do you think a release will be made ?

veewee:
Most likely at the end of the month.

PR #1110, opened 2023-09-20
Apply ternary to null operator

peter279k:
| Q             | A
| ------------- | ---
| Branch        | 2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

It should use the ternary null operator to process the variable is null.

# New Task Checklist:


veewee:
Thanks! :)

PR #1129, opened 2024-04-02
fix: XML and HTML reports as string paths instead of boolean options

petrisorciprian-vitals:
| Q             | A
| ------------- | ---
| Branch        | v2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

Converts the existing 'xml' and 'html' options for Codeception from booleans to strings.
Codeception supports specifying paths for these options.
If a plain option is passed (without the value, i.e. '--xml' or '--html'), Codeception uses defaults of 'report.xml' and 'report.html'.

If a valued option is passed (i.e. '--xml=my-custom-path/report.xml' or '--html=my-custom-path/report.html'), then this is used instead.

This MR just brings this possibility into GrumPHP

### DIFF
--- a/doc/tasks/codeception.md
+++ b/doc/tasks/codeception.md
@@ -20,8 +20,8 @@ grumphp:
             fail_fast: false
             suite: ~
             test: ~
-            xml: false
-            html: false
+            xml: ~
+            html: ~
 ```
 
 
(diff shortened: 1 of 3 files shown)


veewee:
Thanks for the PR @petrisorciprian-vitals 

So if I understand this option correctly, there are 3 ways to use these options?

* no option : no report
* `--xml`: Stores the report by default at `report.xml`
* `--xml=path`: The path where you want to store te report.


Would it make sense to find a way to support all 3 options in here as well, that can be used in a Backwards Compatible way from a grumphp configuration point of view?

I'm thinking:

Keep bool|string as supported types, where:

* `false` would result in no argument (no report)
* `true` would result in `--xml`
* string would result in a custom XML path: `--xml=path`

WDYT?

veewee:

PR #1131, opened 2024-05-14
feat: Add task for Twig-CS-Fixer

antoniovj1:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | 

This pull request adds a task for using https://github.com/VincentLanglet/Twig-CS-Fixer/tree/main

# New Task Checklist:

### DIFF
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -130,7 +130,7 @@ Every task has its own default configuration. It is possible to overwrite the pa
 - [Stylelint](tasks/stylelint.md)
 - [Tester](tasks/tester.md)
 - [TwigCs](tasks/twigcs.md)
-- [TwigCsFixer](tasks/twigcsfixer.md)
+- [Twig-CS-Fixer](tasks/twigcsfixer.md)
 - [XmlLint](tasks/xmllint.md)
 - [YamlLint](tasks/yamllint.md)
 
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -63,6 +63,7 @@ grumphp:
         stylelint: ~
         tester: ~
         twigcs: ~
+        twigcsfixer: ~
         xmllint: ~
         yamllint: ~
 ```
@@ -129,6 +130,7 @@ Every task has its own default configuration. It is possible to overwrite the pa
 - [Stylelint](tasks/stylelint.md)
 - [Tester](tasks/tester.md)
 - [TwigCs](tasks/twigcs.md)
+- [TwigCsFixer](tasks/twigcsfixer.md)
 - [XmlLint](tasks/xmllint.md)
 - [YamlLint](tasks/yamllint.md)
 
--- a/doc/tasks/twigcsfixer.md
+++ b/doc/tasks/twigcsfixer.md
@@ -1,4 +1,4 @@
-#  Twig CS Fixer
+#  Twig-CS-Fixer
 
 Check and fix Twig coding standard using [VincentLanglet/Twig-CS-Fixer](https://github.com/VincentLanglet/Twig-CS-Fixer).
 You can check config file [here](https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md).
--- a/doc/tasks/twigcsfixer.md
+++ b/doc/tasks/twigcsfixer.md
@@ -1,0 +1,37 @@
+#  TwigCs
+
+Check and fix Twig coding standard using [VincentLanglet/Twig-CS-Fixer](https://github.com/VincentLanglet/Twig-CS-Fixer).
+You can check config file [here](https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md).
+
+***Composer***
+
+```
+composer require --dev "vincentlanglet/twig-cs-fixer:>=2"
+```
+
+***Config***
+
+The task lives under the `twigcsfixer` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+grumphp:
+    tasks:
+        twigcsfixer:
+            path: '.'
+            triggered_by: ['twig']
+```
+
+**path**
+
+*Default: null*
+
+By default `.` (current folder) will be used.
+On precommit the path will not be used, changed files will be passed as arguments instead.
+You can specify an alternate location by changing this option. If the path doesn't exist or is not accessible an exception will be thrown.
+
+**triggered_by**
+
+*Default: [twig]*
+
+This option will specify which file extensions will trigger this task.
--- a/doc/tasks/twigcsfixer.md
+++ b/doc/tasks/twigcsfixer.md
@@ -17,35 +17,28 @@ The task lives under the `twigcsfixer` namespace and has following configurable 
 grumphp:
     tasks:
         twigcsfixer:
-            paths: ['.']
-            level: 'NOTICE'
+            paths: []
+            level: ~
             config: ~
             report: 'text'
-            fix: false
             no-cache: false
-            debug: false
-            quiet: false
-            version: false
-            ansi: false
-            no-ansi: false
-            no-interaction: false
             verbose: false
             triggered_by: ['twig']
 ```
 
 **paths**
 
-*Default: null*
+*Default: []*
 
-By default [`.`] (current folder) will be used.
+By default, current folder will be used.
 On precommit only changed files that live in the paths will be passed as arguments.
 
 
 **level**
 
-*Default: 'NOTICE'*
+*Default: 'notice'*
 
-The level of the messages to display (possibles values are : 'NOTICE', 'WARNING', 'ERROR').
+The level of the messages to display (possibles values are : 'notice', 'warning', 'error').
 
 **config**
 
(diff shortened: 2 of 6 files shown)


doc/tasks.md
```diff
@@ -129,6 +130,7 @@ Every task has its own default configuration. It is possible to overwrite the pa
 - [Stylelint](tasks/stylelint.md)
 - [Tester](tasks/tester.md)
 - [TwigCs](tasks/twigcs.md)
+- [TwigCsFixer](tasks/twigcsfixer.md)
```
VincentLanglet:
```suggestion
- [Twig-CS-Fixer](tasks/twigcsfixer.md)
```

antoniovj1: I don't know why I thought all the names used Camel Case, but checking it now there are several with spaces in the name. Do you prefer the name with hyphen or spaces?

VincentLanglet: I created my tool with PHP-CS-Fixer as inspiration, so let's go with `Twig-CS-Fixer` since `PHP-CS-Fixer` is written like this in these tasks.

doc/tasks/twigcsfixer.md
```diff
@@ -0,0 +1,37 @@
+#  Twig CS Fixer
```
antoniovj1:
```suggestion
#  Twig-CS-Fixer
```

doc/tasks.md
antoniovj1: Perfect!

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @extends AbstractExternalTask<ProcessFormatterInterface>
+ */
+class TwigCsFixer extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'path' => '.',
```
veewee: Shouldn't it rather be `paths` as in the actual CLI tool?

It seems we are missing some additional options as well:

https://github.com/VincentLanglet/Twig-CS-Fixer/blob/e9342648fd547758acbf515b4457bd1094ef4814/src/Command/TwigCsFixerCommand.php#L34-L80

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,81 @@
...
+        $arguments = $this->processBuilder->createArgumentsForCommand('twig-cs-fixer');
+        $arguments->add('lint');
+        $arguments->add('--report=text');
+
+        if ($context instanceof GitPreCommitContext) {
+            $arguments->addFiles($files);
+        }
+
+        if ($context instanceof RunContext) {
+            $arguments->add($config['path']);
```
veewee: Is it correct to assume paths is only being used in run context?
One could expect the pats to work as some kinf of filter in pre-commit context as well.
For example, you configure it as:

```yaml
paths:
  - templates/only-check-these-once
```

When committing `templates/unchecked-folder/file.twig` you don't expect the fixer to run on this?


Other tasks have things like:

- path black / white-listst
- the php-cs-fixer task has something like path intersection mode on the provided finder
- ...

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,81 @@
...
+        if (0 === \count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('twig-cs-fixer');
+        $arguments->add('lint');
+        $arguments->add('--report=text');
+
+        if ($context instanceof GitPreCommitContext) {
+            $arguments->addFiles($files);
```
veewee, line 66: This could be lots of files, causing CLI max line limits.
Is this the only way to pass files or is there a different way (through sdin, tmp file, ...) ?

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if ($context instanceof RunContext) {
+            $arguments->add($config['path']);
+        }
+
+        $process = $this->processBuilder->buildProcess($arguments);
```
veewee: You are building and running the process twice here

veewee (changes requested):
Thanks for the PR.
I've added some feedback to the code.

src/Task/TwigCsFixer.php
antoniovj1: Fixed! thanks

src/Task/TwigCsFixer.php
antoniovj1: I've been looking for other ways, but I think this is the only one. I have little experience with this kind of stuff, so maybe I'm missing something...

test/Unit/Task/TwigCsFixerTest.php
```diff
@@ -0,0 +1,156 @@
...
+            $this->mockContext(GitPreCommitContext::class, ['templates/hello.twig', 'templates/hello2.twig', 'other/hello2.twig']),
+            'twig-cs-fixer',
+            [
+                'lint',
+                'templates/hello.twig',
+                'templates/hello2.twig',
+                '--level=NOTICE',
+                '--report=text',
+            ]
+        ];
```
antoniovj1: Now paths are check at precommit

src/Task/TwigCsFixer.php
antoniovj1: Now I've used the array of paths to filter the files that will be considered in the precommit. For example, if the only valid path is 'templates/' and I modify the file src/file.twig, it will be discarded since it is not found in the valid path.

Does it make sense? 

https://github.com/phpro/grumphp/pull/1131/files#r1601843153

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * @extends AbstractExternalTask<ProcessFormatterInterface>
+ */
+class TwigCsFixer extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'paths' => ['.'],
```
veewee: Wouldn't it be better to provide an empty array here?
As far as I can tell from the tool, it will work that way as well.

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+            'level' => 'NOTICE',
```
veewee: maybe better to set level to null by default? That way it uses the tool's defaults instead whilst keeping it overwritable.

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+            'config' => null,
+            'report' => 'text',
+            'fix' => false,
+            'no-cache' => false,
+            'debug' => false,
+            'quiet' => false,
+            'version' => false,
+            'ansi' => false,
```
veewee: Tasks are being ran in sub-processes.
Having ansi / no-ansi configurable does not make a lot of sense for grumphp.

The same goes for no-interaction : it will have no interaction by default since it is being ran in a subprocess in parallel.

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+{
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'paths' => ['.'],
+            'level' => 'NOTICE',
+            'config' => null,
+            'report' => 'text',
+            'fix' => false,
```
veewee: Fix should not be an option here : failing tasks will be fixed based on the user's grumphp preferences

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+            'no-cache' => false,
+            'debug' => false,
```
veewee: debug, quiet and version are default symfony/console flags. I don't think it makes much sense to have them configurable from grumphp's end.

src/Task/TwigCsFixer.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $arguments = $this->processBuilder->createArgumentsForCommand('twig-cs-fixer');
+        $arguments->add('lint');
+
+        if ($context instanceof GitPreCommitContext) {
+            $arguments->addFiles($files);
+        }
+
+        if ($context instanceof RunContext) {
+            foreach ($config['paths'] as $path) {
```
veewee: It would be shorter to use `$arguments->addArgumentArray('%s', $config['paths'])` instead of having the foreach here.

test/Unit/Task/TwigCsFixerTest.php
```diff
@@ -0,0 +1,156 @@
...
+        yield 'no-files-in-paths' => [
+            ['paths' => ['src']],
+            $this->mockContext(RunContext::class, ['other/hello.twig']),
+            function () {
+            }
+        ];
+    }
+
+    public function provideExternalTaskRuns(): iterable
```
veewee, line 105: can you add tests for all possible options to make sure the CLI arguments are being built correctly?

veewee (changes requested):
Thanks for the changes. 
I've added some additional comments

src/Task/TwigCsFixer.php
antoniovj1: Done!

src/Task/TwigCsFixer.php
antoniovj1: Done!

src/Task/TwigCsFixer.php
antoniovj1: Done!

src/Task/TwigCsFixer.php
antoniovj1: Done!

src/Task/TwigCsFixer.php
antoniovj1: Done!

src/Task/TwigCsFixer.php
antoniovj1: Way better! Done!

test/Unit/Task/TwigCsFixerTest.php
antoniovj1: Sure! Done!

veewee:
Looks nice, thanks for your work !

antoniovj1:
> Looks nice, thanks for your work !

Thanks for your help and patience!

Issue #1135, opened 2024-05-18
Add support for git_pre_push


saidatom:
| Q               | A
| --------------- | ---
| Branch         | master for features and deprecations
| New feature?    | yes
| Related tickets | #252

As example I don't want to run codeception at every commit but only at every push but the hook git_pre_push is not available.

veewee:
Closing this issue, will continue in PRs

PR #1136, opened 2024-05-18
Pre push testsuites

saidatom:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Documented?   | yes
| Fixed tickets | #1135 

It is an experiment to get the https://github.com/phpro/grumphp/issues/1135 working, actually there is the problem that if we have a grumphp.yml like that:
```
grumphp:
  process_timeout: 480
  tasks:
    phpcs:
      standard: PSR2
      ignore_patterns:
        - "spec/*Spec.php"
        - "test/*.php"
        - "stubs/*.php"
    phpspec:
      format: progress
      verbose: true
  testsuites:
    git_pre_push:
      tasks: [phpcs, phpspec]
  environment:
    paths:
      - tools
```

# New Task Checklist:


src/Console/Command/Git/PrePushCommand.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * @var TestSuiteCollection
+     */
+    private $testSuites;
+
+    /**
+     * @var StdInFiles
+     */
+    private $stdInFilesLocator;
```
veewee, line 38: The STDIN files locator is not being used

src/Console/Command/Git/PrePushCommand.php
```diff
@@ -0,0 +1,115 @@
...
+    public static function getDefaultName(): string
+    {
+        return self::COMMAND_NAME;
+    }
+
+    protected function configure(): void
+    {
+        $this->setDescription('Executed by the pre-push hook');
+        $this->addOption(
+            'skip-success-output',
```
veewee, line 77: should output be ever skipped?
For pre-commit, this is the case because the commit-msg runs in the same git commit command, just behind the other hook.

src/Console/Command/Git/PrePushCommand.php
```diff
@@ -0,0 +1,115 @@
...
+            null,
+            InputOption::VALUE_NONE,
+            'Skips the success output. This will be shown by another command in the git commit hook chain.'
+        );
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $localRef = $this->changedFilesLocator->getLocalRef();
```
veewee, line 86: I think we need to introduce a new PushedFiles locator which contains all of this logic and the `getPushedFiles()` logic. Pollution the ChangedFiles locator doesn't seem like a good idea here.
By moving it to its own locator, you can keep these specific git details "hidden" and just use the pushed files from this command.

I'm not sure how it behaves exactly, but I'm sure test-cases will be added to show how it works exactly.

src/Locator/ChangedFiles.php
```diff
@@ -46,6 +46,66 @@ public function locateFromGitRepository(): FilesCollection
...
+    {
+        return $this->repository->run('rev-parse', ['HEAD']);
+    }
+    private function getRemoteRef(): ?string
+    {
+        return $this->repository->tryToRunWithFallback(
+            function (): ?string {
+                return $this->repository->run('rev-parse', ['--abbrev-ref', '--symbolic-full-name', '@{u}']);
+            },
+            'null'
```
veewee, line 63: having a `'null'` string doesn't seem like a good idea. Instead maybe use `null`?

src/Locator/ChangedFiles.php
```diff
@@ -46,6 +46,66 @@ public function locateFromGitRepository(): FilesCollection
...
+        );
+    }
+    
+    public function getRemoteSHA1(): ?string
+    {
+        if ($this->getRemoteRef() === 'null') {
+            return null;
+        }
+        /**
+         * @psalm-suppress PossiblyUndefinedArrayOffset
```
veewee, line 73: Maybe better to validate the format and return null if it's unexpected?

src/Locator/ChangedFiles.php
```diff
@@ -75,6 +135,6 @@ private function makeFileRelativeToProjectDir(File $file): SplFileInfo
             $file->isRename() ? $file->getNewName() : $file->getName()
         );
 
-        return new SplFileInfo($filePath, dirname($filePath), $filePath);
+        return new SplFileInfo($filePath);
```
veewee, line 138: any specific reason why this was changed?

src/Locator/ChangedFiles.php
```diff
@@ -46,6 +46,66 @@ public function locateFromGitRepository(): FilesCollection
...
+            function () use ($commits): ?string {
+                return $this->repository->run('diff', ['--name-only', trim($commits), '--oneline']);
+            },
+            'null'
+        );
+        if ($diff === 'null') {
+            return new FilesCollection($files);
+        }
+
+        foreach (explode("\n", $diff) as $file) {
```
veewee, line 101: would this always be `\n` or `\r\n` dependeing on the system?

src/Task/Ant.php
```diff
@@ -36,7 +37,9 @@ public static function getConfigurableOptions(): ConfigOptionsResolver
 
     public function canRunInContext(ContextInterface $context): bool
     {
-        return $context instanceof GitPreCommitContext || $context instanceof RunContext;
+        return $context instanceof GitPreCommitContext
```
veewee, line 40: By default, this will run the task in both pre-commit and pre-push scenario?
I think that might be a bit too much for a lot of people.

Would it make sense to have a new metadata field that can be used to change the context instead?
See https://github.com/phpro/grumphp/blob/v2.x/doc/tasks.md#metadata

Something like:

```yaml
grumphp:
    tasks:
        anytask:
            metadata:
                git_stage: ["pre-commit", "pre-push"]
```

Which can be used to specify at what stages in your git flow you want to run this task.
The canRunInContext will then by based on the metadata configuration, which will default to `["pre-commit"]`.

WDYT?

src/Locator/ChangedFiles.php
```diff
@@ -46,6 +46,66 @@ public function locateFromGitRepository(): FilesCollection
...
+         */
+        list($remoteName, $branchName) = explode('/', $this->getRemoteRef(), 2);
+        $output = $this->repository->run('ls-remote', [$remoteName, $branchName]);
+        return $output ? strtok($output, "\t") : null;
+    }
+
+    /**
+     * @return FilesCollection
+     */
+    public function locateFromGitDiff(string $fromSha, ?string $toSha): FilesCollection
```
veewee, line 84: What worries me a bit, is that there seem to be a lot of different states possible which require slightly different approaches / solutions.
How sure are you that all possible situations are covered?

Maybe the new PushedFiled locator should also contain details on how it works and what it covers exactly.

src/Console/Command/Git/PrePushCommand.php
```diff
@@ -0,0 +1,115 @@
...
+        return $results->isFailed() ? self::EXIT_CODE_NOK : self::EXIT_CODE_OK;
+    }
+
+    protected function getPushedFiles(?string $localRef, ?string $localSha, ?string $remoteSha): FilesCollection
+    {
+        if (empty($remoteSha)) {
+            return $this->changedFilesLocator->locateFromRawDiffInput('HEAD');
+        }
+
+        if ($localRef === '(delete)') {
```
veewee, line 109: should the task run if you delete a branch ?

veewee (commented):
Thanks @saidatom,

I've quickly skimmed through the code and added some pointers / questions.
Beside the review comments, A lot of tests are missing at the moment.

veewee:

Issue #1126, opened 2024-02-28
[Task] Add bin/console task for Symfony


veewee:
| Q               | A
| --------------- | ---
| Version         | `next`
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | 

Provide a task that makes it possible to run Symfony's `bin/console` with a specific command:

```yaml
tasks:
  symfony_console:
    command: [ "lint:container", "-vvv"]
```

This will run : `"./bin/console" "lint:container" "-vvv"`

If the command fails (non-zero exit code), the grumphp task fails.

Running multiple symfony console commands (in parallel) is already covered:

```yaml
tasks:
  lint-container:
    command: [ "lint:container", "-vvv"]
    metadata:
      task: symfony_console
```

veewee:
Closing for now, we'll continue in the PR.

PR #1138, opened 2024-05-31
1126 bin console task sf

bortefi:
| Q             | A
| ------------- | ---
| Branch        | v2.x 
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #1126 

Added Task: Symfony Console: execute symfony command
- :no_entry_sign:  ~php executable finder~
- [x] configurable bin path, defaults to project root `./bin/console`

# New Task Checklist:

- [ ] Are all CI services returning green? :warning: awaiting github ci runner

### Future Improvements (not included)
* :no_entry_sign: ~symfony/flex recipe for console component could be required to generate `bin/console` if not existing on the project~

### DIFF
--- a/doc/tasks/symfony_console.md
+++ b/doc/tasks/symfony_console.md
@@ -2,13 +2,9 @@
 
 Run a symfony console command.
 
-## Composer
+## Requirements
 
-Requires the Symfony Console component: [Console Component](https://symfony.com/components/Console)
-
-```bash
-composer require symfony/console
-```
+This task requires a Symfony application with console component.
 
 ## Config
 
--- a/doc/tasks/symfony_console.md
+++ b/doc/tasks/symfony_console.md
@@ -1,0 +1,47 @@
+# Symfony Console
+
+Run a symfony console command.
+
+## Composer
+
+Requires the Symfony Console component: [Console Component](https://symfony.com/components/Console)
+
+```bash
+composer require symfony/console
+```
+
+## Config
+
+The task lives under the `symfony_console` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+grumphp:
+    tasks:
+        symfony_console:
+            command: [ "lint:container", "-vvv" ]
+```
+
+**command**
+
+Specify the symfony command with defined options and arguments.  
+Verify the installed console component version for available commands `./bin/console list`
+
+## Note: Multiple Console command tasks
+
+[Run the same task twice with different configuration](../tasks.md#run-the-same-task-twice-with-different-configuration)
+
+Specific running multiple symfony console commands:
+
+```yaml
+# grumphp.yml
+grumphp:
+  lint-container:
+      command: [ "lint:container", "-vvv"]
+      metadata:
+        task: symfony_console
+  lint-yaml:
+      command: [ "lint:yaml", "path/to/yaml"]
+      metadata:
+        task: symfony_console
+```
--- a/doc/tasks/symfony_console.md
+++ b/doc/tasks/symfony_console.md
@@ -19,17 +19,62 @@ The task lives under the `symfony_console` namespace and has following configura
 grumphp:
     tasks:
         symfony_console:
+            bin: "./bin/console"
             command: [ "lint:container", "-vvv" ]
+            ignore_patterns: [],
+            whitelist_patterns: [],
+            triggered_by: ['php', 'yml', 'xml'],
+            run_always: false
 ```
-
-**Note:** GrumPHP will use the default Symfony console path `./bin/console` from the project root.
 
 ### Parameters
 
+**bin**
+
+*Default: `./bin/console`*
+
+Specify the path to the Symfony Console script. 
+
 **command**
+
+*Default: `[]`*
 
 Specify the symfony command with defined options and arguments.  
 Verify the installed console component version for available commands `./bin/console list`
+
+**ignore_patterns**
+
+*Default: []*
+
+This is a list of file patterns that will be ignored by the Symfony console task. 
+Leave this option blank to run the task for all files defined in the whitelist_patterns and or triggered_by extensions.
+
+**whitelist_patterns**
+
+*Default: []*
+
+This is a list of regex patterns that will filter files to validate. With this option you can skip files like tests. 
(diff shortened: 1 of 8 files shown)


src/Task/SymfonyConsole.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function canRunInContext(ContextInterface $context): bool
+    {
+        return ($context instanceof GitPreCommitContext || $context instanceof RunContext);
+    }
+
+    public function run(ContextInterface $context): TaskResultInterface
+    {
+        $config = $this->getConfig()->getOptions();
```
veewee, line 57: There is currently no way to tell when a task should be ran.
Maybe it's a good idea to include something like:

```php
        $files = $context->getFiles()
            ->extensions($config['triggered_by'])
            ->paths($config['whitelist_patterns'] ?? [])
            ->notPaths($config['ignore_patterns'] ?? []);

        if (0 === \count($files)) {
        }
```

It might be handy though, to have an `run_always` option for commands where if it makes no sense to filter the files?

src/Task/SymfonyConsole.php
```diff
@@ -0,0 +1,64 @@
...
+        if (0 === \count($context->getFiles())) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        if ('' === \implode('', $config['command'])) {
+            return TaskResult::createNonBlockingFailed(
```
veewee: maybe it makes more sense to validate this during option resolving?

```
$resolver->setAllowedValues('command', function ($value): bool {
    // return true or false
});
```

doc/tasks/symfony_console.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Config
+
+The task lives under the `symfony_console` namespace and has following configurable parameters:
+
+```yaml
+# grumphp.yml
+grumphp:
+    tasks:
+        symfony_console:
```
veewee, line 17: the bin option is missing here and in the parameters section ad well.

doc/tasks/symfony_console.md
```diff
@@ -0,0 +1,51 @@
...
+            command: [ "lint:container", "-vvv" ]
+```
+
+**Note:** GrumPHP will use the default Symfony console path `./bin/console` from the project root.
+
+### Parameters
+
+**command**
+
```
veewee, line 35: In other task documentations, the default values are being displayed as well.

veewee (commented):
Thanks for the PR.
I've listed a couple of possible improvements.

src/Task/SymfonyConsole.php
```diff
@@ -0,0 +1,64 @@
...
+
+        if (0 === \count($config['command'])) {
+            return TaskResult::createNonBlockingFailed(
+                $this,
+                $context,
+                'Missing "command" configuration for task "symfony_console".'
+            );
+        }
+
+        $arguments = $this->processBuilder->createArgumentsForCommand($config['bin']);
```
bortefi: add process lookup of PHP executable

src/Task/SymfonyConsole.php
bortefi: True, I had in mind to use the metadata  `enabled`,
but `white/ignore patterns` and `run_always` is more flexible

src/Task/SymfonyConsole.php
bortefi: Implemented, 
the only drawback i.m.h.o. with this approach is that the message is a bit less clear and the command fails blocking on config.

src/Task/SymfonyConsole.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $files = $context->getFiles()
+            ->extensions($config['triggered_by'])
+            ->paths($config['whitelist_patterns'] ?? [])
+            ->notPaths($config['ignore_patterns'] ?? []);
+        if (!$config['run_always'] && 0 === \count($files)) {
+            return TaskResult::createSkipped($this, $context);
+        }
+
+        $php = (new PhpExecutableFinder())->find();
```
veewee: Symfony's executable finder might add arguments to the command:
https://github.com/symfony/process/blob/7.1/PhpExecutableFinder.php#L90

This is not supported by grumphp's processbuilder and will fail.
Maybe easier to just skip the phpExectuableFinder and use this?

```php
$arguments = $this->processBuilder->createArgumentsForCommand('php')
```

doc/tasks/symfony_console.md
```diff
@@ -0,0 +1,96 @@
+# Symfony Console
+
+Run a symfony console command.
+
+## Composer
+
+Requires the Symfony Console component: [Console Component](https://symfony.com/components/Console)
+
+```bash
+composer require symfony/console
```
veewee: This is only valid if you are using flex:
https://github.com/symfony/recipes/tree/main/symfony/console/5.3

Maybe easier to tell that they should be operating in a symfony application?
The suggest in composer.json might be removed as well maybe? Since It's rather : have the framework installed than have a specific dependency.

src/Task/SymfonyConsole.php
```diff
@@ -0,0 +1,86 @@
...
+                ->addAllowedTypes('whitelist_patterns', ['array'])
+                ->addAllowedTypes('triggered_by', ['array'])
+                ->addAllowedTypes('run_always', ['bool'])
+                ->setAllowedValues(
+                    'bin',
+                    static fn(string $bin): bool => '' !== $bin
+                )
+                ->setAllowedValues(
+                    'command',
+                    static fn(array $command): bool => '' !== \implode('', $command)
```
veewee, line 43: why do you implode here?
You could as well just check the array directly

veewee (changes requested):

src/Task/SymfonyConsole.php
bortefi: good catch

src/Task/SymfonyConsole.php
bortefi: The implode would catch (some very bad config oversight) of `[""]` a non empty array with an empty string

Issue #1141, opened 2024-06-12
Composer validate task not running when composer.lock changes


vever001:
| Q               | A
| --------------- | ---
| Version         | 2.5.0 and 2.x
| Bug?            | (arguably) yes
| New feature?    | no
| Question?       | no
| Documentation?  | no
| Related tickets | -

Currently the composer task gets triggered/run only when `composer.json` changes.
In many cases, only the `composer.lock` gets changed (of course, if it is committed/versioned in the project) and the composer task should run, otherwise we can end up with an invalid composer.lock hash (for example).

For example: if we merge/rebase a branch and have `composer.lock` conflicts, and the rebase doesn't contain any change to `composer.json`, and forget to run `composer update --lock` to refresh the hash after fixing conflicts, then we end up with an invalid lock hash.

**Steps to reproduce:**
On a project with the `composer.lock` committed/versioned, and the composer task enabled.
1. Hack (change manually) the composer.lock hash.
2. Commit, and notice that the grumphp composer task doesn't run.
>> The composer lock is now invalid, grumphp could prevent that.

**Proposed resolution:**
Add the composer.lock to the detection.

Possibly add a setting to enable/disable this, in order to keep the current behavior?
But would that have any applications/added value?

PR #1142, opened 2024-06-12
Trigger composer task also when composer.lock changes.

vever001:
See https://github.com/phpro/grumphp/issues/1141


veewee:
Thanks!

PR #1144, opened 2024-06-28
Add gherkin_lint task

saidatom:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

`gherkin-lint-php` might be easier to set up and configure, especially if you only need basic Gherkin linting.

# New Task Checklist:


composer.json
```diff
@@ -52,6 +52,7 @@
         "brianium/paratest": "Lets GrumPHP run PHPUnit in parallel.",
         "codeception/codeception": "Lets GrumPHP run your project's full stack tests",
         "consolidation/robo": "Lets GrumPHP run your automated PHP tasks.",
+        "dantleech/ngherkin-lint-php": "Lets GrumPHP lint your Gherkin files.",
```
vever001: typo? dantleech/**n**gherkin-lint-php

src/Task/GherkinLint.php
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @extends AbstractExternalTask<ProcessFormatterInterface>
+ */
+class GherkinLint extends AbstractExternalTask
+{
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'directory' => 'features',
```
vever001, line 25: I noticed that if the specified directory doesn't exist, it fails with:
```
In FeatureFinder.php line 25:
  Provided path "/var/www/html/features" does not exist

lint <path>
Failed to run grumphp run --tasks=gherkin_lint: exit status 1
```

Is this expected ? or should we skip in such cases? 
How are other tasks handling these cases?

vever001 (commented):
I found a minor typo and have 1 question. 
Besides that it works great! Thanks

src/Task/GherkinLint.php
saidatom: Yes, it's expected. For example src/Task/Gherkin.php has the same configuration, and returns the same error.

src/Task/GherkinLint.php
```diff
@@ -0,0 +1,69 @@
...
+            'config' => null,
+        ]);
+
+        $resolver->addAllowedTypes('directory', ['string']);
+        $resolver->addAllowedTypes('config', ['null', 'string']);
+        $resolver->addAllowedValues('config', [null, 'text']);
```
veewee, line 31: shouldn't config be any string? Now you are limiting it to the literal `text` here ?

On second notice: there doesn't seem a way to specify the configuration file. It assumes a hardcoded path instead:

* https://github.com/dantleech/gherkin-lint-php?tab=readme-ov-file#configuration
* https://github.com/dantleech/gherkin-lint-php/blob/f3277c4945261d26a3098b5e003b51d815eaed4b/bin/gherkinlint#L20-L27
* https://github.com/dantleech/gherkin-lint-php/blob/f3277c4945261d26a3098b5e003b51d815eaed4b/src/Command/LintCommand.php#L27-L32

So I'dd say its better to remove the config option alltogether.

test/Unit/Task/GherkinLintTest.php
```diff
@@ -0,0 +1,102 @@
...
+        yield 'no-files-after-triggered-by' => [
+            [],
+            $this->mockContext(RunContext::class, ['notafeaturefile.txt']),
+            function () {}
+        ];
+    }
+
+    public function provideExternalTaskRuns(): iterable
+    {
+        yield 'defaults' => [
```
veewee, line 92: this provider should contain a test for every configurable option to make sure it behaves the correct way.

veewee (commented):
Thanks! Looks good in general. I've added a couple of small pointers.

veewee:

PR #1145, opened 2024-07-05
Support --skip-property-types of ValidateSchemaCommand

wucdbm:
Doctrine seems to have added property types validation in the aforementioned command This will allow passing the skip property types option where making changes to that is undesirable

| Q             | A
| ------------- | ---
| Branch        | v2.x (Or, IDK?)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no Because the current behaviour is property types are validated depending on your ORM version, this only allows switching that off
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

There seems to be an option `skip-property-types` in Doctrine's `ValidateSchemaCommand` which can't be turned off and so always blows up on my on-commit hook and is something I can't/won't fix in my project, but I'd still like the rest of what `ValidateSchemaCommand` does. I guess that's an issue with other people as well.

PS Added a test case, please review though

### DIFF
--- a/doc/tasks/doctrine_orm.md
+++ b/doc/tasks/doctrine_orm.md
@@ -10,6 +10,7 @@ grumphp:
         doctrine_orm:
             skip_mapping: false
             skip_sync: false
+            skip_property_types: false
             triggered_by: ['php', 'xml', 'yml']
 ```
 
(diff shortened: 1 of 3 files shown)


veewee:
Looks good, Thanks!

wucdbm:
@veewee Um, would you consider releasing this one, seeing as the other PR in the milestone is sort of dead for a while

I am currently stuck to the commit hash, so I wouldn't get any patch releases and/or bugfixes

veewee:
@wucdbm Tagged ;)

PR #1151, opened 2024-08-27
There was a double loading of the autoloader

dezer32:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Deprecations? | yes/no
| Documented?   | yes/no

# New Task Checklist:

### DIFF
--- a/bin/grumphp
+++ b/bin/grumphp
@@ -8,9 +8,11 @@ use Symfony\Component\Console\Output\ConsoleOutput;
 (function () {
     // First load current projects autoloader for smarter  global / phar installs ....
     // A bit opinionated at the moment though:
-    $autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
-    if (is_file($autoloaderInWorkingDirectory)) {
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for the PR.

Can you give me more details about what is going wrong exactly.
I'm not sure if adding an env var for this purpose is the way to go.
Maybe there are better solutions?

dezer32:
“Hi, sorry for the lack of information :)”
“GrumPHP is connected in the project, and the project runs locally through Docker using Docker-sync. An error occurs stating that the Composer autoload class is already included. I found out that this place includes the class, and since GrumPHP is installed only in the project and nowhere else, it tries to include it a second time in the loop below.”

veewee:
I don't fully grasp the issue nor the implementation.
I don't think it's a good idea to use an env variable to decide what to load or not.

Therefore, I will be closing this PR. Feel free to provide additional information, a reproducer, ...  if you want to continue on this. Thanks anyway!

PR #1155, opened 2024-10-19
Adds 'parallel' option support to phpcs task.

mrmishmash:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

This change adds support for 'parallel' option for the phpcs and phpcbf fixer. By default the option is set to null, which means it
runs normally, if it's a positive integer that's the number of processes that will be used. If 'auto' is used the number of processors on the machine will determine the number.

# New Task Checklist:

### DIFF
--- a/src/Task/Phpcs.php
+++ b/src/Task/Phpcs.php
@@ -65,7 +65,7 @@ class Phpcs extends AbstractExternalTask
         $resolver->addAllowedTypes('report_width', ['null', 'int']);
         $resolver->addAllowedTypes('exclude', ['array']);
         $resolver->addAllowedTypes('show_sniffs_error_path', ['bool']);
-        $resolver->addAllowedTypes('parallel', ['null', 'int', 'string']);
+        $resolver->addAllowedTypes('parallel', ['null', 'int']);
 
         return ConfigOptionsResolver::fromOptionsResolver($resolver);
     }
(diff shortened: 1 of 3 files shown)


src/Task/Phpcs.php
```diff
@@ -161,7 +164,76 @@ private function addArgumentsFromConfig(
...
     }
+
+    protected function getParallelValue(string|int|null $option): ?int
+    {
+        if ($option === null) {
+            return null;
+        }
+
+        if (is_string($option)) {
+            if ($option === 'auto') {
```
veewee: To keep the maintenance of this codebase manageable, we rely on the tools to provide the options for us.
Does phpcs have a way to automatically detect the number of cores? If not, it should probably be done in their rather than in this package.

Ideally, the implementation fo passing the cores options, should be as simple as:
`$arguments->addOptionalArgument('--parallel=%s', $config['parallel']);`

Relying on phpcs to do the proper validations of the int range / auto value - where null means full back to phpcs defaults.


Additional note here, is that GrumPHP itself also is running in parallel and possibly other tasks are running in parallel as well. If they all take up all cores, it might have an impact on the system.

veewee (commented):
Lookin' good!
I do have a remark about the auto CPU detection

mrmishmash:
Yeah I was also thinking about this, it just adds complexity to the task and there also wasn't a straightforward way to test it. Good point that this should be done in phpcs.

In any case developers can add detection of the number of cores in their build processes and set it as the value for the 'parallel' option.

So with that in mind I have removed the auto detection of the cpu cores. Thank you for the input.

veewee:
Thanks!

PR #1162, opened 2024-11-26
phpversion - updated support of not EOL versions

jwderoos:
| Q             | A
| ------------- | ---
| Branch        | v2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes

phpversion errors on php 8.1 as original support was set to end on the 25th of november 2024. Official support was extended. Also added php 8.3 and php 8.4


resources/config/util.yml
```diff
@@ -20,5 +20,7 @@ services:
               '7.3': '2021-12-06 23:59:59'
               '7.4': '2022-11-28 23:59:59'
               '8.0': '2023-11-26 23:59:59'
-              '8.1': '2024-11-25 23:59:59'
-              '8.2': '2025-12-08 23:59:59'
+              '8.1': '2025-12-31 23:59:59'
+              '8.2': '2027-12-31 23:59:59'
```
veewee: it looks like you swapped 8.2 and 8.3 release date?
At least according to https://www.php.net/supported-versions.php

veewee (commented):
Thanks, I've noticed a small misconfiguration.

resources/config/util.yml
jwderoos: Did I? Oh, yes looks like I did.

jwderoos: Fixed, sorry about that.

PR #1165, opened 2024-12-16
Allow laravel/serializable-closure 2

simoheinonen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "amphp/parallel": "^2.1",
         "doctrine/collections": "^1.6.8 || ^2.0",
         "gitonomy/gitlib": "^1.3",
-        "laravel/serializable-closure": "^1.1",
+        "laravel/serializable-closure": "^2.0",
         "monolog/monolog": "^2.0 || ^3.0",
         "ondram/ci-detector": "^4.0",
         "psr/container": "^1.1 || ^2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "amphp/parallel": "^2.1",
         "doctrine/collections": "^1.6.8 || ^2.0",
         "gitonomy/gitlib": "^1.3",
-        "laravel/serializable-closure": "^1.1",
+        "laravel/serializable-closure": "^1.1 || ^2.0",
```
veewee: Is it worth supporting both? Supporting 2.x seems like a better approach to me - but I'm not sure how heavily people depend on this dependency in Laravel. Any clue?

veewee (commented):
Thanks for the PR!

composer.json
simoheinonen: No idea :D 

I guess the biggest (or only) difference between 1 and 2 is they dropped PHP 7.x support, which GrumPHP doesn't support anyway so probably fine to go just with 2

mrunkel:
Any idea/plan when this will get "released" and available on packagist?

To add some context, this is preventing the installation of grumphp in laravel 11 project.

veewee:
@mrunkel Forgot about this one. I'll tag it somewhere this week.

veewee:
Tagged ! (Finally ...) :)

Issue #1157, opened 2024-10-31
Set path and/or full filename in triggered_by


evs-xsarus:
| Q               | A
| --------------- | ---
| Version         | 2.5.0
| Bug?            | no
| New feature?    | yes
| Question?       | no
| Documentation?  | no
| Related tickets | comma-separated list of related tickets

Due to the use of dockers, my grump config only uses shell tasks , example:

```yaml
grumphp:
    process_timeout: 512
    tasks:
        magento_composer:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./magento validate"]
            metadata:
                task: shell
        magento_stan-check:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./magento stan-check"]
            metadata:
                task: shell

        integrations_composer:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./integrations validate"]
            metadata:
                task: shell
        integrations_cs-check:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./integrations cs-check"]
            metadata:
                task: shell
```

Most of these tasks start properly when PHP files are changed as per documentation https://github.com/phpro/grumphp/blob/v2.x/doc/tasks/shell.md  for triggered_by .

Sadly, the `composer validate` is not triggered if only composer.json/composer.lock is changed. I can add these:

```yaml
        magento_composer:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./magento validate"]
            metadata:
                task: shell
            triggered_by: [json, lock]
```
But that means the task will run even if a non-related json or lock file is changed.

To avoid this, specifying a complete filename should help as in this example:

```yaml
            triggered_by: [composer.lock, composer.json]
```

or even with wildcards:

```yaml
            triggered_by: [composer.*]
```

As you can see in my example, I've duplicate tasks but for different folders. The path should be definable too in triggered by, full example:

```yaml
        magento_composer:
            scripts:
                - ["-c", "/usr/local/bin/composer2 --working-dir=./magento validate"]
            metadata:
                task: shell
            triggered_by: ['./magento/composer.*']
```

veewee:
Hello @evs-xsarus,

That would be possible. We do a similir thing for the `ComposerRequireChecker` task.
However : since we are now filtering `triggered_by` as extensions, it would be considered a BC break.
[...]

PR #1167, opened 2025-02-21
Ticket #1157: support specific filenames as trigger for Shell.

evs-xsarus:
| Q             | A
| ------------- | ---
| Branch        | [feature/1157-triggered_by_filenames](https://github.com/evs-xsarus/grumphp/tree/feature/1157-triggered_by_filenames)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | [comma-separated list of tickets fixed by the PR, if any](https://github.com/phpro/grumphp/issues/1157)

Question, I would like to be able to specify a pathname in the filename. This fails now because the pattern is checked against the filename without a path. How to fix?

# New Task Checklist:

### DIFF
--- a/doc/tasks/shell.md
+++ b/doc/tasks/shell.md
@@ -9,6 +9,8 @@ grumphp:
     tasks:
         shell:
             scripts: []
+            ignore_patterns: [],
+            whitelist_patterns: [],
             triggered_by: [php]
 ```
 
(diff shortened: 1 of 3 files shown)


src/Task/Shell.php
```diff
@@ -59,7 +61,12 @@ public function canRunInContext(ContextInterface $context): bool
     public function run(ContextInterface $context): TaskResultInterface
     {
         $config = $this->getConfig()->getOptions();
-        $files = $context->getFiles()->extensions($config['triggered_by']);
+        $files = new FilesCollection(
```
veewee: $files isn't really being used anywhere.
Instead of building the list of files, it might make more sense to count() the amount of files in either `trigger`.

That way we don't have to do all these lookups all of the time.

test/Unit/Task/ShellTest.php
```diff
@@ -27,6 +27,7 @@ public function provideConfigurableOptions(): iterable
             [
                 'scripts' => [],
                 'triggered_by' => ['php'],
+                'triggered_by_filenames' => [],
```
veewee: We'll also need a test-case that covers how the task behaves on the newly added configuration.

src/Task/AbstractParserTask.php
```diff
@@ -50,10 +50,12 @@ protected static function sharedOptionsResolver(): OptionsResolver
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'triggered_by' => [],
+            'triggered_by_filenames' => [],
```
veewee: not sure why this one is altered? It's not being used anywhere?

doc/tasks/shell.md
```diff
@@ -35,11 +37,30 @@ grumphp:
 
 *Note:* When using the `-c` option, the next argument should contain the full executable with all parameters. Be careful: quotes will be escaped!
 
+**ignore_patterns**
+
+*Default: []*
+
+This is a list of file patterns that will be ignored by the Symfony console task.
```
veewee: It's the shell task.

veewee (commented):
Looks good. A small copy/paste thing in the docs but good to go otherwise.

evs-xsarus:
@veewee the doc has been updated.

veewee:
Thanks!

PR #1171, opened 2025-04-15
Invoke GrumPHP executable using the PHP binary so it respects the PHP version Composer was invoked with.

### LINKED ISSUE #1170: Composer plugin tasks not executed using the PHP version Composer is invoked with
Given:
 - two PHP versions installed side by side - `8.2` and `8.3`
 - `/usr/bin/php` invokes PHP `8.2` (managed via `alternatives`), so that's the default version
 - Composer is explicitly invoked with PHP `8.3`, i.e. `/usr/bin/php8.3 ./composer.phar install`

In this setup GrumPHP `git:init` and `configure` scripts are executed using PHP `8.2` because the plugin calls `vendor/bin/grumphp` directly instead of using the PHP binary Composer is using for the current invocation.

I'm not familiar with Composer Plugin API enough to tell if Composer exposes internally the PHP binary it was invoked with. However it looks like there is `@php` syntax for scripts that does something similar to what `${MAKE}` would do in a Makefile, so Composer must maintain some sort of reference to the current PHP binary.

echernyavskiy:
| Q             | A
| ------------- | ---
| Branch        | invoke-grumphp-using-php-binary
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | -

Fixes #1170

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -108,9 +108,12 @@
     },
     "scripts": {
         "post-install-cmd": [
-            "GrumPHP\\Composer\\DevelopmentIntegrator::integrate"
+            "@integrate"
         ],
         "post-update-cmd": [
+            "@integrate"
+        ],
+        "integrate": [
             "GrumPHP\\Composer\\DevelopmentIntegrator::integrate"
         ]
     }
--- a/src/Composer/DevelopmentIntegrator.php
+++ b/src/Composer/DevelopmentIntegrator.php
@@ -8,6 +8,7 @@ use Composer\Script\Event;
 use GrumPHP\Collection\ProcessArgumentsCollection;
 use GrumPHP\Process\ProcessFactory;
 use GrumPHP\Util\Filesystem;
+use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
 use Symfony\Component\Process\ProcessBuilder;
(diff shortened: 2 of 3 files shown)


veewee:
@echernyavskiy Thanks for the PR.
This doesn't seem to be working on windows. Can you take a look at the failing appveyor tasks?

echernyavskiy:
@veewee, the `8.4` failures there look like they have to do with the environment:
```
php --version
php : PHP Warning:  'C:\windows\SYSTEM32\VCRUNTIME140.dll' 14.22 is not compatible with this PHP build linked with 14.43 in Unknown on line 0
```
For other PHP versions it looks like GrumPHP's `pre-commit` file is not getting created for some reason. I'm not sure how to investigate this - if someone running Windows is able to look into it, that'd be great.

veewee:
I don't own a Windows computer either. But can't merge this the way it is, since it breaks every windows installation :)
Previously, i've used appveyor to debug this kind of stuff directly from within the E2E tests and rerun. It's slow and tyring, but it gets the job done.

veewee:
Closing this one based on inactivity. If there is still interest, a new PR can be created

PR #1177, opened 2025-07-16
Change the task compiler logic to allow resolving of environment variables.

### LINKED ISSUE #1146: Env variables don't get expanded in task definitions
| Q               | A
| --------------- | ---
| Version         | GrumPHP 2.6.0 
| Bug?            | yes
| New feature?    | no
| Question?       | no
| Documentation?  | yes

Following the [docs on using environment variables and parameters](https://github.com/phpro/grumphp/blob/v2.x/doc/parameters.md#configuration) doesn't actually function as expected.
The example seems to work, however setting the variable to false still has the config enabled.

i've debugged it by using it in a task and dumping the task definition in the code. Which results in:
```
GrumPHP\Task\Config\Metadata^ {#659
  -metadata: array:5 [
    "blocking" => true
    "task" => ""
    "label" => ""
    "priority" => 0
    "enabled" => "env_19eeff9b12d8f08c_bool_GRUM_COMPOSER_NORMALIZE_ENABLED_b9f45633a4808b98cb74867bcd84c333"
  ]
}
```
instead of the expected bool value, it is a string identifying the variable to be resolved.
This feels similar to https://github.com/symfony/symfony/issues/45868

**My configuration**
```yaml
# grumphp.yml
parameters:
    env(GRUM_COMPOSER_NORMALIZE_ENABLED): 'false'
    env(GRUM_COMPOSER_NORMALIZE_VERBOSE): 'false'
    
grumphp:
    tasks:
        composer_normalize:
            metadata:
                enabled: '%env(bool:GRUM_COMPOSER_NORMALIZE_ENABLED)%'
            indent_size: 4
            indent_style: "space"
            no_update_lock: false
            verbose: '%env(bool:GRUM_COMPOSER_NORMALIZE_VERBOSE)%'
```

**Steps to reproduce:**
```sh
# Generate empty folder
mkdir tmp
cd tmp
git init
echo "vendor" > .gitignore
pbpaste > grumphp.yml
composer require --dev phpro/grumphp

# Run GrumPHP:
git add -A && git commit -m"Test"
# or
./vendor/bin/grumphp run
```

**Result:**
```
Composer normalize is still excecuted instead of not excecuting at all
```

veewee:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | #1146

Fixes  #1146

This allows for task configurations to resolve environment variables.
Example:

```yaml
grumphp:
    tasks:
        phpunit:
          testsuite: "%env(string:PHPUNIT_TESTSUITE)%"
          metadata:
            enabled: '%env(bool:PHPUNIT_ENABLED)%'
```

```sh
PHPUNIT_TESTSUITE=Foo PHPUNIT_ENABLED=1 ./bin/grumphp run
```

or, you can make it work with environment configurations:

```yaml
grumphp:
  environment:
    variables:
      PHPUNIT_ENABLED: "1"
      PHPUNIT_TESTSUITE: "Foo"
```


src/Configuration/ContainerFactory.php
```diff
@@ -32,6 +32,24 @@ public static function build(InputInterface $input, OutputInterface $output): Co
...
+         *
+         * Investigate if there is a way to simplify this:
+         *
+         * - Load environment config first + register .env files
+         * - Guess paths based on the environment config
+         * - Build container with the environment config already loaded.
+         *
+         * If this is not possible, the LazyTaskConfig can be reintroduced.
+         */
+
```
kira0269: You can totally load your app based on env vars as you want.

I suggest to use the `symfony/runtime` component to achieve that. 
The endpoint `bin/grumphp` would be like the famous `bin/console`. The only difference is that you will use your own runtime implementation.

src/Configuration/Resolver/TaskConfigResolver.php
```diff
@@ -28,14 +36,21 @@ public function listAvailableTaskNames(): array
...
+
         unset($config['metadata']);
+        $resolvedConfig = $resolver->resolve($config);
 
-        return $resolver->resolve($config);
+        return new TaskConfig(
+            $taskName,
+            $resolvedConfig,
+            $metadata
+        );
```
kira0269: `TaskConfg` instances should be kept in memory, non ?
Or instead of having such "TaskConfig resolver", you can:
- create a service definition for each TaskConfig in the container named like `test_config.<task_name>`.
- inject this service into the ones which need it

PS: The logic behind tasks configuration could be simpler. What's the reason of so many steps and layers ?

src/Configuration/Resolver/TaskConfigResolver.php
```diff
@@ -44,7 +59,7 @@ public function fetchByName(string $taskName): ConfigOptionsResolver
             throw TaskConfigResolverException::unknownTask($taskName);
         }
 
-        $class = $this->taskMap[$taskName];
+        $class = $this->taskMap[$taskName]['class'];
```
kira0269: Feels weird to need this kind of logic here.

veewee: Every task class contains its own specific configuration definition. It needs to be loaded somewhere.

This resolver class now contains an array of all tasks options configured in grumphp.yaml and converts it into validated `TaskConfig` for the given task. To me, that is the goal of this class.

Let me know if you have specific improvements in mind in order to to better here.


(this logic didn't really change and was how it originaly worked. The class is now just one layer deeper in the taskMap)

src/Configuration/Resolver/TaskConfigResolver.php
veewee: Every TaskConfig is kept in memory inside the container. This method will only be called once per task since it is in the factory of creating a task instance.

Every task in grumphp.yaml needs its own 'configured' service : It is possible to create the same task with multiple configurations. For example : 2x phpunit with default settings.

This is in general how it works:

- A PHP task has a generic definition in the container
- The tasks config is read from grumphp.yaml
- Config per task is parsed and gets combined with the specific generic task
- This result is stored as a "configured task"


Any concrete suggestions in making it simpler without losing functionality are welcome. 
This logic is here for quite some while and has proven itself though.

src/Configuration/ContainerFactory.php
veewee: This comment is more about:

https://github.com/phpro/grumphp/blob/v2.x/doc/parameters.md

```
grumphp:
    environment:
        files: []
        variables: {}
        paths: []
```

These variables / and files are loaded after the container gets compiled, meaning the env placeholders are already filled in.


Adding symfony/runtime to the mix is not going to make this easier.
It's just a matter of:

- parsing the grumphp.yaml file first
- then loading the env related things
- then loading the container

which will simplify logic like detecting all paths.

kira0269: The `symfony/runtime` is used to centralize how your app would run, taking in count the current context, like environment.

For example, The SymfonyRuntime will parse the .env file in order to know the APP_ENV and the APP_DEBUG en vars.

The steps you describe here:
> parsing the grumphp.yaml file first
> then loading the env related things
> then loading the container

Would be done in this Runtime class

src/Configuration/Resolver/TaskConfigResolver.php
kira0269: The configuration could be injected into the Task service definition, directly.

If you just use this config like that, it's way much simpler.

But if you need to access configuration without instantiating the task, you could also create create specific TaskConfig service (like `PhpStanConfig` for example) for each Task. This way, you could alos implement some specific logic for each task.

src/Configuration/Resolver/TaskConfigResolver.php
kira0269: See my answer here: https://github.com/phpro/grumphp/pull/1177#discussion_r2213475750

veewee:
@kira0269, Thanks for your comments;

In this PR, I focused on getting the environment variables working inside the task configurations.
I agree that some parts during bootstrapping Grumphp or configuring the tasks could be improved. 
Since it works right now and I don't want to reinvent the wheel with no real gain, I'm not sure it's worth the time or trouble to work on this. 

Anyways, thanks for your feedback and for figuring out how to solve the issue :)

oallain:
Thanks for that @veewee 🎉

Issue #1088, opened 2023-05-17
Add support for PSR-4/`composer dump-autoload` testing


TravisCarden:
| Q               | A
| --------------- | ---
| Version         | `v1.16.0`
| Bug?            | no
| New feature?    | **yes**
| Question?       | no
| Documentation?  | **yes**
| Related tickets | n/a

I just discovered that `composer validate` will pass even if there are PSR-4 problems. That's not the fault of GrumPHP, but it may be misleading.

**Steps to reproduce/Result:**
Given a Composer project with PSR-4 configured and GrumPHP installed, just put a class file in the wrong directory and...

```
# `composer validate` reports nothing and returns a success status code.
$ composer validate; echo $?
./composer.json is valid
0

# `composer dumpautoload` reports the error and returns a failure code given the `--strict-psr` option.
$ composer dumpautoload --strict-psr 1> /dev/null; echo $? # (Silence stdout for clarity.)
Class Example\Good located in ./src/Bad.php does not comply with psr-4 autoloading standard. Skipping.
1
```

**Proposal:**
Either add a new task (or an option to the existing `composer` task), to use `composer dumpautoload` to check PSR-4 violations.

veewee:
Hmm interesting. 
The dump autoload changes files, which might not be the wanted scenario here.
Maybe it could make sense for composer to add this check to the validation command somehow?

TravisCarden:
^ I've proposed a `dumpautoload --dry-run` option. 🙂

TravisCarden:
Update: The above-proposed `dumpautoload --dry-run` option has been committed upstream and is scheduled for release in Composer 2.6.

veewee:
This seems solved. Feel free to reopen or open up a new issue if not.

TravisCarden:
This issue was unblocked upstream, but it didn't trickle down automatically because it the functionality wasn't added to `composer validate`. I sent a PR for a new task for it. ^

PR #1186, opened 2025-10-09
Add `composer_validate_autoload` task

TravisCarden:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | https://github.com/phpro/grumphp/issues/1088

This adds a new `composer_validate_autoload` task that checks for PSR-4 or PSR-0 mapping errors. Testing it on the GrumPHP codebase itself surfaced some such errors, as shown for illustration in this error output:

```
composer_validate_autoload
============

Class GrumPHP\Test\Unit\Task\ESLintTest located in ./test/Unit/Task/ESLintTest.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
Class GrumPHP\Test\Unit\Task\StylelintTest located in ./test/Unit/Task/StylelintTest.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
Class GrumPHPE2E\ValidatePathsTask located in ./test/fixtures/e2e/tasks/ValidatePathsTask.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
Class GrumPHPE2E\DummyTask located in ./test/fixtures/e2e/tasks/DummyTask.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
Class GrumPHPE2E\SuccessfulTask located in ./test/fixtures/e2e/tasks/SuccessfulTask.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
Class GrumPHPE2E\ValidateExtension located in ./test/fixtures/e2e/extension/ValidateExtension.php does not comply with psr-4 autoloading standard (rule: GrumPHPTest\ => ./test). Skipping.
```

Since I was in there anyway, I fixed a couple of them. But I left the `GrumPHPE2E` classes alone, because other classes depend on them and I didn't want to break anything because I didn't understand it.

# New Task Checklist:

### DIFF
--- a/doc/tasks.md
+++ b/doc/tasks.md
@@ -15,7 +15,6 @@ grumphp:
         codeception: ~
         composer: ~
         composer_normalize: ~
-        composer_psr: ~
         composer_require_checker: ~
         composer_script: ~
         deptrac: ~
@@ -66,6 +65,7 @@ grumphp:
         tester: ~
         twigcs: ~
         twigcsfixer: ~
+        validate_autoload: ~
         xmllint: ~
         yamllint: ~
 ```
@@ -82,7 +82,6 @@ Every task has its own default configuration. It is possible to overwrite the pa
 - [Codeception](tasks/codeception.md)
 - [Composer](tasks/composer.md)
 - [Composer Normalize](tasks/composer_normalize.md)
-- [Composer PSR](tasks/composer_psr.md)
 - [Composer Require Checker](tasks/composer_require_checker.md)
 - [Composer Script](tasks/composer_script.md)
 - [Doctrine ORM](tasks/doctrine_orm.md)
@@ -135,6 +134,7 @@ Every task has its own default configuration. It is possible to overwrite the pa
 - [Tester](tasks/tester.md)
 - [TwigCs](tasks/twigcs.md)
 - [Twig-CS-Fixer](tasks/twigcsfixer.md)
+- [Validate Autoload](tasks/validate_autoload.md)
 - [XmlLint](tasks/xmllint.md)
 - [YamlLint](tasks/yamllint.md)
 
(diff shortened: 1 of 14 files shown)


veewee:
Looks nice! Thanks for the PR.

I'm wondering: wouldn't something like `composer_autoload` or something in this line be a better name?
psr seems a bit too broad here.


(Those E2E test classes are indeed injected in dynamically created folders, so those shouldnt be changed here :) )

TravisCarden:
You're welcome! I actually called it `composer_autoload` at first and then renamed it because I thought that `composer_psr` better reflected the business value instead of the mechanics. In other words, it _uses_ `composer dump-autoload`, but what it _does_, as the `composer dump-autoload` help says for the `--strict-psr` flag, is check for PSR-4 or PSR-0 mapping errors. I suppose it comes down to what one considers the broader concept--autoloading or PSR--and whether you want the naming to be tool-centric or value-centric. I'll gladly rename it is you like. 🙂

veewee:
IIRC, the dump-autoload command has some additional checks on top of only checking PSR within `--optimize`.
So I'dd say the thing you are validating is wether the classes can be autoloaded according to psr standards.

I think something in lines with autoload or validate_autoload would be more appropiate.

TravisCarden:
^ There you go!

veewee:
Thanks for applying the changes.
Would you mind adding the composer prefix though? `composer_validate_autoload`
Sorry for the confusion here.

TravisCarden:
Okay, I wondered about that. 😛 ^ Here's the update.

veewee:
Thanks!

PR #1188, opened 2025-10-10
Update services.yml to remove suffixes for windows added by default from symfony/process

apohutenumera:
…rom symfony/process

Prevent problem for suffixes for Linux/WSL users

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no
| Fixed tickets | 1185

When using grumPHP with dependencies, if users are on WSL, .bat is present and is taking precedence over "no extension" phpcs and others.

### DIFF
--- a/resources/config/services.yml
+++ b/resources/config/services.yml
@@ -18,7 +18,7 @@ services:
     executable_finder:
         class: Symfony\Component\Process\ExecutableFinder
         calls:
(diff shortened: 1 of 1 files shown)


veewee:
Thanks!

PR #1194, opened 2025-11-23
Add support for Symfony 8.0

Jibbarth:
| Q             | A
| ------------- | ---
| Branch        | v2.x
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | no

:warning: Can't be tested right now, some dependencies don't yet allow SF8


arnedesmedt:
You will also need to get rid of the !tagged attributes in the runner.yml. !tagged_iterator should be used now.
And xml loaders are no longer supported.

Jibbarth:
Agreed, it deserve more work I expected initially :thinking:

veewee:
Thanks!

Jibbarth:
Thank you @veewee for finishing it :star_struck:

arnedesmedt:
Txh! @veewee I think the loader factory would still need a change. If I update to the v2.x version, I still get 

```PHP Fatal error:  Uncaught Error: Class "Symfony\Component\DependencyInjection\Loader\XmlFileLoader" ```

So I think we first should check if the XmlFileLoader class exists in the LoaderFactory, if not, don't initiate it, and add it to the list of loaders. 

WDYT?

veewee:
Ah yes, you are right.

It thought it just got deprecated, but it is actually removed in SF 8.0.
Now wondering why the CI did not fail ... It seems to be installing 7.4 for some (probably dependency) reason.

veewee:
@arnedesmedt continuing here
https://github.com/phpro/grumphp/pull/1199

PR #1199, opened 2025-12-03
Improve SF 8.0 support

veewee:
Follow-up on https://github.com/phpro/grumphp/pull/1194

### DIFF
--- a/.github/workflows/grumphp.yml
+++ b/.github/workflows/grumphp.yml
@@ -45,10 +45,14 @@ jobs:
                   restore-keys: ${{ runner.os }}-composer-
             - name: Install dependencies (highest)
               if: matrix.composer-deps == 'highest'
-              run: composer update --prefer-dist --no-progress --no-suggest
+              run: |
+                composer config platform.php ${{ matrix.php-versions }}
+                composer update --prefer-dist --no-progress --no-suggest --ignore-platform-req=php+
             - name: Install dependencies (lowest)
               if: matrix.composer-deps == 'lowest'
-              run: composer update --prefer-dist --no-progress --no-suggest --prefer-lowest
+              run: |
+                composer config platform.php ${{ matrix.php-versions }}
+                composer update --prefer-dist --no-progress --no-suggest --prefer-lowest --ignore-platform-req=php+
             - name: Install dependencies (lock)
               if: matrix.composer-deps == 'lock'
               run: |
--- a/src/Configuration/LoaderFactory.php
+++ b/src/Configuration/LoaderFactory.php
@@ -26,16 +26,22 @@ final class LoaderFactory
     public static function createLoader(ContainerBuilder $container, array $paths = []): DelegatingLoader
     {
         $locator = new FileLocator($paths);
-        $resolver = new LoaderResolver([
-            $xmlLoader = new XmlFileLoader($container, $locator, self::ENV),
+
+        /** @Deprecated - Remove in a future version of PHP where SF > 7.4 */
+        $xmlLoader = class_exists(XmlFileLoader::class)
+            ? new XmlFileLoader($container, $locator, self::ENV)
+            : null;
+
+        $resolver = new LoaderResolver(array_filter([
+            $xmlLoader,
             $yamlLoader = new YamlFileLoader($container, $locator, self::ENV),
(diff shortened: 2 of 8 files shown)


veewee:
@arnedesmedt Can you validate this works for you?


Composer:
```
  "repositories": [
    {
      "type": "git",
      "url": "https://github.com/veewee/grumphp"
    }
  ],
  "require-dev": {
    "phpro/grumphp": "dev-sf80-upgrade-v2@dev"
  },
```

arnedesmedt:
Hmm, I'm now getting: ` There are no commands defined in the "git" namespace.` 

Is it because of the tagged iterator? that changed? => edit: don't think so.

Probably the loader isn't loading all services now. So the commands are also not tagged as commands...

veewee:
@arnedesmedt It seems that the changes in `AddConsoleCommandPass` require the new `AsCommand` attribute. I changed this

arnedesmedt:
Yes, It works now. I only now get a deprecation error from gitonomy/gitlib:
```PHP Deprecated:  Using Diff::parse without raw information is deprecated. See https://github.com/gitonomy/gitlib/issues/227.```

Since I also needed to update that package.

**Comment by veewee** [comment]
> Yes, It works now. I only now get a deprecation error from gitonomy/gitlib: `PHP Deprecated: Using Diff::parse without raw information is deprecated. See https://github.com/gitonomy/gitlib/issues/227.`
> 
> Since I also needed to update that package.


Is this something that gets triggered from within this package's code or rather something internal in gitlob?
Can you tell me how to reproduce this?

**Comment by arnedesmedt** [comment]
It's triggered here: https://github.com/phpro/grumphp/blob/v2.x/src/Git/GitRepository.php#L69

But it depends on the content of your git diff if it's triggered yes or no. Let me check this further.

**Comment by veewee** [comment]
I see, that is when you use git diff directly as stdin (which happens during commit inside the commit hook)

We might need to add the `--raw` flag to it?

https://github.com/phpro/grumphp/blob/56c6c0583539fdbf6e1d1ab6699c429dc9c29336/resources/hooks/local/pre-commit#L9


Strange that I cant seem to get that error here though.

**Comment by arnedesmedt** [comment]
That worked, but you will need to change some other data processing too. I'm getting for example the following output:

```GrumPHP detected a pre-commit command.

Running task 1/5: phpcs... ✘
Running task 2/5: phpstan... ✘
Running task 3/5: rector... ✘
Running task 4/5: git_branch_name... ✔
Running task 5/5: phpmd... 

phpcs
=====

ERROR: The file ":100644 100644 4cc8f2182 0951c4a95 M   config/reference.php" does not exist.
```
So I think grumphp is now detecting file names from the git diff that aren't files at all. But just data from the git diff, that is now in another format then it was first.

veewee:
@arnedesmedt maybe we should just release this version first and then focus on the gitlib?

I'm trying to reproduce using `gitlib 1.5` but cant seem to find a way:

```
> DIFF=$(git -c diff.mnemonicprefix=false -c diff.noprefix=false --no-pager diff -r -p -m -M --full-index --no-color --staged | cat)

> echo $DIFF

diff --git a/src/test.php b/src/test.php
index 61ace196d4129411be0800919d883e132b0bf075..c9a7fe70994dba1a8a47e7948ec082c260728d3b 100644
--- a/src/test.php
+++ b/src/test.php
@@ -1,2 +1,6 @@
 <?php
 phpinfo();
+
+exit 0;
+
+
```

```
> printf "%s\n" "${DIFF}" | php -d'error_reporting=-1' 'vendor/bin/grumphp' 'git:pre-commit'                                                                                                                                      

GrumPHP detected a pre-commit command.

Running task 1/1: file_size... ✔
```

arnedesmedt:
yes go ahead

arnedesmedt:
@veewee I think, I can already see the problem, why you can't reproduce.
I'm on version `dev-main` for `gitonomy/gitlib`. It's because they haven't yet released a new version to support sf8. 

So if they release a new version, you would be able to reproduce, I guess.

veewee:
Ah ok. Since we dont really need the raw data, I'm wondering if the deprecation warning should be there.
I'll ask it in their repo.

PR #1203, opened 2026-01-16
Cover more of Drupal's custom file extensions

damienmckenna:
The Drupal CMS uses a few additional file extensions that would be worthwhile to add.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | n/a

# New Task Checklist:

### DIFF
--- a/doc/tasks/phpcs.md
+++ b/doc/tasks/phpcs.md
@@ -226,7 +226,10 @@ grumphp:
                 - cfg/
                 - libraries/
             triggered_by:
-                - php
-                - module
                 - inc
+                - install
+                - module
+                - php
+                - profile
+                - theme
 ```


veewee:
Thanks for your suggestion.

This is just documentation, so I'm not going to include it specifically for Drupal since this package is framework agnostic.

damienmckenna:
My rationale was that it would provide a more useful suggested starting point for people starting off - you already have a Drupal section in the page, this just improves it.

veewee:
Oh sorry, totally looked over that. I assumed it was in the main config block.
Will merge :)

Thanks for clearifying.

damienmckenna:
Thank you for your understanding!

PR #1212, opened 2026-03-28
Add support for doctrine/collections 3.x

a-treschilov:
## Summary

- Extends the `doctrine/collections` version constraint from `^1.6.8 || ^2.0` to `^1.6.8 || ^2.0 || ^3.0`
- No code changes required — GrumPHP's collection classes only use core `ArrayCollection` methods that remain fully compatible in 3.x
- doctrine/collections 3.0.0 requires PHP 8.4+, so Composer's platform resolution automatically selects 3.x on PHP 8.4/8.5 and 2.x on PHP 8.2/8.3

## Compatibility notes

- `filter()` and `map()` gained native `Collection` return types in 3.x; `FilesCollection::filter(): self` is a valid covariant override ✓
- Internally `filter()` still uses `createFrom()` → `new static()`, so subclasses continue to receive their own type at runtime ✓
- No GrumPHP code extends any of the classes made `final` in 3.x (`Criteria`, `Expr\*`, `ExpressionBuilder`) ✓

## Checklist

- [x] Branch: `v2.x`
- [x] Bug fix: no
- [x] New feature: no
- [x] BC breaks: no
- [x] Deprecations: no
- [x] Fixed tickets: N/A

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "composer-plugin-api": "^2.0",
         "amphp/amp": "^3.0",
         "amphp/parallel": "^2.1",
-        "doctrine/collections": "^1.6.8 || ^2.0",
+        "doctrine/collections": "^1.6.8 || ^2.0 || ^3.0",
         "gitonomy/gitlib": "^1.6",
         "laravel/serializable-closure": "^2.0",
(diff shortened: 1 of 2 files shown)


veewee:
Thanks!

Issue #1193, opened 2025-11-20
Add Mago as task


yannicka:
Mago is a new linter/analyzer for PHP: <https://github.com/carthage-software/mago>, as PHPStan. It would be good for it to work with GrumPHP.

veewee:
I'dd surely accept a PR for that!

johnatas-x:
Hey,

I opened a PR on this topic: [#1216](https://github.com/phpro/grumphp/pull/1216)
I’m not sure this is the best approach, but I decided to group all Mago commands into a single task.

It is possible to configure which command(s) should run, as well as the options for each command.
I did not make the arguments (paths) configurable, as I believe it is better to manage them directly in `mago.toml`.

By default, the formatter, linter, and analyzer are enabled with the `--staged` option (which seems most appropriate in a GrumPHP pre-commit context). The guard is disabled.

I tested this in a Drupal context by [running the task twice](https://github.com/phpro/grumphp/blob/v2.x/doc/tasks.md#run-the-same-task-twice-with-different-configuration
): once in pre-commit (with priority 1 to run the formatter before other tasks), and once in CI. To enable one or the other, I use a variable on the `enabled` metadata.

Example of my Mago configuration:
```yml
parameters:
  env(IS_DDEV_PROJECT): 'false'
  env(IS_MR_PIPELINE): 'false'
grumphp:
  tasks:
    mago:
      formatter: true
      formatter_options: ['--staged']
      linter: false
      linter_options: []
      analyzer: false
      analyzer_options: []
      guard: true
      guard_options: []
      metadata:
        blocking: true
        enabled: '%env(bool:IS_DDEV_PROJECT)%'
        label: null
        priority: 1
        task: null
    mago_ci:
      formatter: true
      formatter_options: [ '--check' ]
      linter: false
      linter_options: []
      analyzer: false
      analyzer_options: []
      guard: true
      guard_options: []
      metadata:
        blocking: true
        enabled: '%env(bool:IS_MR_PIPELINE)%'
        label: null
        priority: 0
        task: mago
```

What do you think about this first version? I still have a few open questions:

- Is a single task with the 4 commands better than having 4 separate tasks with one command each?
- Should we add a parameter to control the execution order of the commands?
- Considering Mago’s advanced [config file discovery](https://mago.carthage.software/guide/configuration#configuration-file-discovery
), do you think there is value in making the config file path configurable, for more exotic setups?

johnatas-x:
The PR has been updated 😉

PR #1216, opened 2026-04-30
Add Mago as task

johnatas-x:
| Q             | A
| ------------- | ---
| Branch        | mago-task
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | #1193

This PR adds a Mago task

Mago is a recent tool that includes a formatter (like phpcs), a linter (like phpmd), and a static analyzer (like phpstan).
It is known for its performance and is starting to be increasingly adopted (for example, in Drupal).
Its integration into GrumPHP seems valuable.

mago-grumphp

# New Task Checklist:


src/Task/Mago.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+class Mago extends AbstractExternalTask
+{
+
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'formatter' => true,
+            'formatter_options' => ['--staged'],
```
veewee: staged should only be a default during pre-commit I suppose?

src/Task/Mago.php
```diff
@@ -0,0 +1,97 @@
...
+            'linter' => true,
+            'linter_options' => ['--staged'],
+            'analyzer' => true,
+            'analyzer_options' => ['--staged'],
+            'guard' => false,
+            'guard_options' => [],
+        ]);
+
+        $resolver->addAllowedTypes('formatter', ['bool']);
+        $resolver->addAllowedTypes('formatter_options', ['array']);
```
veewee: In other tasks, we provide configuration keys that map the options of the CLI.
Doing it like this introduces a new way of configuring tasks, which I'dd like to avoid at this point.

src/Task/Mago.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function run(ContextInterface $context): TaskResultInterface
+    {
+        $config = $this->getConfig()->getOptions();
+
+        if ($config['formatter'] === false
```
veewee: I was thinking: wouldn't it make more sense to introduce a task per type of command?
Something like following tasks:

- `mago_fmt`
- `mago_lint`
- `mago_analyze`
- `mago_guard`

Every task will have it's own set of configurable options.

veewee (commented):
Thanks for the PR.
I think this task in current state is not in line with other tasks. I've added some remarks inline in the code. Feel free to stat a discussion.

johnatas-x:
Hey @veewee,

Thanks for taking the time to review the PR and share your feedback.
Your comments align with the questions I raised in the [related issue](https://github.com/phpro/grumphp/issues/1193#issuecomment-4353677159).

To be fully transparent, after opening the PR I had already started reworking it to introduce 4 commands. I’ll update the PR this week with those changes; it should end up closer to how the SecurityChecker tasks are structured, which should resolve the three threads.

If you have any additional feedback, feel free to share it so I can address everything at once.

johnatas-x:
Hey,

I just pushed a refactor splitting the `mago` task into 4 distinct subtasks: `mago_lint`, `mago_analyze`, `mago_format`, and `mago_guard`. Each task now follows the same structure as other tasks, with its own configurable options, dedicated documentation, and unit tests.

The `mago` task itself now serves two purposes: it acts as a redirect, pointing users to the 4 subtasks when invoked directly, and as a base class holding the shared logic to avoid code duplication across the subtasks.

Happy to hear any feedback — whether on the option coverage, naming, documentation, or anything else that could be improved.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+use GrumPHP\Task\Config\ConfigOptionsResolver;
+use GrumPHP\Task\Context\ContextInterface;
+use GrumPHP\Task\Context\GitPreCommitContext;
+use GrumPHP\Task\Context\RunContext;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @extends AbstractExternalTask<ProcessFormatterInterface>
+ */
+class Mago extends AbstractExternalTask
```
veewee: Maybe better to have it abstract as wel? There is no such thing as a mago task

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+{
+
+    public static function getConfigurableOptions(): ConfigOptionsResolver
```
veewee: not needed when abstract.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+    {
+        return ConfigOptionsResolver::fromOptionsResolver(new OptionsResolver());
+    }
+
+    public function canRunInContext(ContextInterface $context): bool
+    {
+        return $context instanceof GitPreCommitContext || $context instanceof RunContext;
+    }
+
+    public function run(ContextInterface $context): TaskResultInterface
```
veewee: not needed when abstract.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+            . '(https://github.com/phpro/grumphp/blob/master/doc/tasks/mago/linter.md)'.PHP_EOL
+        );
+    }
+
+    protected static function configureSharedOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults([
+            'retain-codes' => [],
+            'ignore-baseline' => null,
+            'fix' => null,
```
veewee: lets not add the fix option.
Rather use something like this:

https://github.com/phpro/grumphp/blob/732a228acacd062faec211e0e3e4317ce4ea9bd6/src/Task/PhpCsFixer.php#L107-L116

By default grumphp does dry-run.
When failed, it asks if you want it to be fixed or not (meaning it 'll run the fix command)

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+            'fail-on-remaining' => null,
+            'sort' => null,
+            'fixable-only' => null,
+            'reporting-format' => null,
+            'reporting-target' => null,
+            'minimum-report-level' => null,
+            'minimum-fail-level' => null,
+            'dry-run' => null,
```
veewee: remove dry-run option.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+        $resolver->addAllowedValues('reporting-format', [
+            null, 'rich', 'medium', 'short', 'ariadne', 'github', 'gitlab',
+            'json', 'count', 'code-count', 'checkstyle', 'emacs', 'sarif',
+        ]);
+        $resolver->addAllowedValues('reporting-target', [null, 'stdout', 'stderr']);
+        $resolver->addAllowedValues('minimum-report-level', [null, 'note', 'help', 'warning', 'error']);
+        $resolver->addAllowedValues('minimum-fail-level', [null, 'note', 'help', 'warning', 'error']);
+    }
+
+    protected function resolveFixOption(array $config): ?string
```
veewee: not needed. See comment above

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+    {
+        return $config['fix'];
+    }
+
+    protected function validateFixCompatibility(
```
veewee: I dont think we'll need this given that we'll have control about building the fix command from within the grumphp tasks directly with the approach above.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+        $arguments->add('--fix');
+        $arguments->addOptionalArgument('--potentially-unsafe', 'potentially-unsafe' === $fix);
+        $arguments->addOptionalArgument('--unsafe', 'unsafe' === $fix);
+        $arguments->addOptionalArgument('--fail-on-remaining', $config['fail-on-remaining']);
+        $arguments->addOptionalArgument('--dry-run', $config['dry-run']);
+    }
+
+    protected function addSharedArguments(
+        ProcessArgumentsCollection $arguments,
+        array $config
```
veewee: add a psalm-type (shape) maybe to config? (same goes for `addFixArguments`)

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+            default => null,
+        };
+
+        return null !== $error ? TaskResult::createFailed($this, $context, $error) : null;
+    }
+
+    protected function addFixArguments(
+        ProcessArgumentsCollection $arguments,
+        array $config,
+        ?string $fix
```
veewee: no need for the fix var.

src/Task/MagoAnalyzer.php
```diff
@@ -0,0 +1,59 @@
...
+
+class MagoAnalyzer extends Mago
+{
+
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'no-stubs' => null,
+            'staged' => null,
```
veewee: does it make sense to have `staged` here?
We could possibly automatically do this during pre-commit and dont do it during run context?

That way we don't have to pass files ever.. Otherwise we'dd need to pass files during pre-commit.

src/Task/MagoFormatter.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'type' => 'default',
+        ]);
+
+        $resolver->addAllowedTypes('type', ['string']);
+        $resolver->addAllowedValues('type', ['default', 'dry-run', 'check', 'staged']);
```
veewee: lets not hide options behind a custom 'type'

src/Task/MagoGuard.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'no-stubs' => null,
+            'checks' => 'all',
+        ]);
+
+        $resolver->addAllowedTypes('no-stubs', ['null', 'bool']);
+        $resolver->addAllowedTypes('checks', ['string']);
+        $resolver->addAllowedValues('checks', ['all', 'structural', 'perimeter']);
```
veewee: lets not hide options behind a custom 'checks'. make them bools.

src/Task/MagoLinter.php
```diff
@@ -0,0 +1,65 @@
...
+{
+
+    public static function getConfigurableOptions(): ConfigOptionsResolver
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'semantics' => null,
+            'pedantic' => null,
+            'only' => [],
+            'staged' => null,
```
veewee: see previous feedback: lets make staged or not dependant of the context.

src/Task/Mago.php
```diff
@@ -0,0 +1,145 @@
...
+            . '- mago_format '
+            . '(https://github.com/phpro/grumphp/blob/master/doc/tasks/mago/formatter.md)'.PHP_EOL
+            . '- mago_guard '
+            . '(https://github.com/phpro/grumphp/blob/master/doc/tasks/mago/guard.md)'.PHP_EOL
+            . '- mago_lint '
+            . '(https://github.com/phpro/grumphp/blob/master/doc/tasks/mago/linter.md)'.PHP_EOL
+        );
+    }
+
+    protected static function configureSharedOptions(OptionsResolver $resolver): void
```
veewee: are those options actually shared across all the commands in mago as well?
Because this could bite us if not. For example: see formatter tasks, which is not using the shared options.

veewee:
Thanks for the changes.
I currently only looked at the tasks and added some comments.
In general, I think we should do things a bit different here and there.

johnatas-x:
No worries
Did you add comments? I can’t see any

johnatas-x:
Hi @veewee,

Thanks for all your feedback!

I'm working through it bit by bit. I'll push a new version this weekend that incorporates your comments, and I'll also include answers to your questions.

Talk to you soon.

johnatas-x:
Hi, here is a summary of all the changes made based on your feedback.

---

### `staged` option removed from config — auto-detected from context

The `staged` flag is no longer a user-facing option. It is now automatically passed to `mago lint` / `mago analyze` when running in a `GitPreCommitContext`, and omitted in `RunContext`. This matches GrumPHP's philosophy of letting the context drive behavior rather than the user config.

On that note: I kept `--staged` (mago's native flag) rather than passing file paths from the diff context. The reason is that files and directories processed by Mago are defined in `mago.toml`. Passing explicit paths from GrumPHP's diff would *override* that configuration — meaning a file not covered by `mago.toml` could end up being linted/analyzed/formatted unintentionally. Using `--staged` delegates the file selection entirely to Mago's own configuration, which is the safe and expected behavior.

---

### `mago lint` / `mago guard` — always run in dry-run + fix mode

These two tasks always run with `--fix --dry-run --fail-on-remaining`. When the task fails, GrumPHP offers to re-run with `--fix` applied (without `--dry-run`). The fix aggressiveness is controlled by a new `fix-mode` option (`safe` by default, `potentially-unsafe`, `unsafe`).

---

### `mago analyze` — runs directly for full diagnostic output

The analyzer runs `mago analyze` without `--fix --dry-run`, as combining those flags suppresses the detailed diagnostic output. When the task fails, GrumPHP offers a fix command (`mago analyze --fix`) built separately, with the appropriate fix-mode flag if configured.

---

### Removed options incompatible with `--fix`

`fixable-only`, `reporting-format`, `reporting-target`, and `minimum-fail-level` have been removed from the shared options as they cannot be used alongside `--fix`. `fail-on-remaining` is now hardcoded (always on) since it is required to detect unfixed issues during the dry-run step.

---

### `mago_formatter` — always runs `--dry-run`, offers fix in both contexts

The `type` option has been removed. The formatter always runs with `--dry-run`. In both `GitPreCommitContext` and `RunContext`, GrumPHP offers to re-run without `--dry-run` to apply formatting in-place.

---

### `mago_guard` — `checks` replaced by `structural` and `perimeter` bools

Following your feedback, the custom `checks` string option has been replaced by two independent boolean flags `structural` and `perimeter`, consistent with how other options are handled across Mago tasks.

---

### Abstract `Mago` class kept

I kept the `Mago` abstract class, which is extended by `MagoAnalyzer`, `MagoLinter`, and `MagoGuard`. These three share the same execution pattern (shared options, `--fix --dry-run`, `createFailedWithFix`). `MagoFormatter` now extends `AbstractExternalTask` directly as it has its own distinct behavior and does not benefit from the shared abstract.

> Are you happy with this approach, or would you prefer I remove the abstract and duplicate its content across the three tasks?

---

I ran a few tests on a real project and things seem to work as expected. That said, my Mago configuration is fairly minimal, so edge cases with more advanced setups (custom rules, baseline files, analyzer stubs, etc.) haven't been covered. Additional testing from someone with a more complete Mago configuration would be welcome.

Happy to address any further feedback, thanks!

veewee:
Hey @johnatas-x,

Rather than going back and forth on the open threads another dozen times, I pushed a version of how I think the Mago tasks could look. Easier to react to something concrete than to keep describing it.

The biggest change: I dropped the abstract `Mago` class. It was trying to share logic across all four commands, but the commands aren't really that alike. Guard can't auto-fix, format handles staged files differently, and the linter/analyzer don't even run the same way. So the base class ended up carrying behaviour that only fit some of them, which made the whole thing harder to follow than it should be. Each command is now a standalone task you can read top to bottom.

While I was in there I lined the tasks up with how the rest of GrumPHP behaves (dry-run by default, offer to fix when it makes sense) and with what Mago actually supports. I went through every flag against the real `mago` CLI, so guard no longer pretends it can fix things, and the structural/perimeter choice is now a single `mode` option since you can't use both at once anyway.

Tests cover each option per task and the fix commands they generate, and the docs are updated to match.

Could you pull it and try it on a real project, especially the pre-commit runs? Want to make sure it behaves the way you'd expect on a setup more complete than mine. If it feels right to you, I'll get it merged.

johnatas-x:
Hi @veewee,

Thank you for the effort you've put into this.

I'm going to review your changes to better understand the approach and improve my own work going forward.

I'll test the branch on three real projects during the course of next week and will get back to you with feedback based on those tests.

veewee:
Tested it out locally and seems to do its thing :)

johnatas-x:
Hello,

I can confirm that it's working (tested on three real-world projects).

My tests were carried out with very simple configurations, which I'll gradually harden over time. If I encounter any issues or identify opportunities for improvement, I'll be happy to provide feedback and submit enhancements as I go.

I think we can close the related issue.

Thanks a lot for your work!

Issue #1218, opened 2026-05-15
High memory consumption


ckuran:
Q | A
-- | --
Version | 2.20.0
Bug? | yes
New feature? | no
Question? | no
Documentation? | no
Related tickets | ~

### Issue

Executing multiple tasks in a repository with thousands of PHP files can lead to Fatal errors due to exceeding allowed memory and the need to manually increase `memory_limit` just for GrumPHP process. I ended up requiring myself to set the `memory_limit=2G` because otherwise the process fails and cuts the parallel processing. Extending the number of tasks requires constantly increasing number of memory.

### Error
```
Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 8192 bytes) in /app/vendor/phpro/grumphp/src/Collection/FilesCollection.php on line 256
```

### Examples

Repository: ~9900 PHP files only

**Example with just one linter:**
```
Running tasks with priority 0!
==============================

Running task 1/1: jsonlint... ✔

Peak memory usage: 192.0 MB
```

**Example with all linters:**
```
Running tasks with priority 0!
==============================

Running task 1/3: jsonlint... ✔
Running task 2/3: xmllint... ✔
Running task 3/3: yamllint... ✔

Peak memory usage: 368.0 MB
```

**Example with all my current suite:**
```
Running tasks with priority 0!
==============================

Running task 1/12: jsonlint... ✔
Running task 2/12: xmllint... ✔
Running task 3/12: yamllint... ✔
Running task 4/12: openapi... ✔
Running task 5/12: rector... ✔
Running task 6/12: behat... ✔
Running task 7/12: phpmd... ✔
Running task 8/12: phpcs... ✔
Running task 9/12: phpstan... ✔
Running task 10/12: psalm... ✔
Running task 11/12: test-unit... ✔
Running task 12/12: test-integration... ✔

Peak memory usage: 1152.0 MB
```

It's worth noting that setting `parallel: false` reduces the memory consumption, but also raises the time to complete all tasks to unacceptable level.

### Steps to reproduce:
- Checkout some huge repository
- Configure GrumPHP
- Before `return` in  `RunCommand.php`  add the following to check memory usage:
```php
if ($this->io->isVerbose()) {
    $this->io->write([
        PHP_EOL,
        sprintf(
            '&#x3C;comment>Peak memory usage: %.1f MB&#x3C;/comment>',
            (float) memory_get_peak_usage(true) / 1024.0 / 1024.0
        )
    ]);
}
```
[...]

PR #1219, opened 2026-05-15
Reduce memory consumption in FilesCollection

ckuran:
| Q             | A
| ------------- | ---
| Branch        | 2.x
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Documented?   | n/a
| Fixed tickets | #1218 

This PR reduces memory consumption in `FilesCollection`. The key changes:
1. Lower memory in parallel mode. When GrumPHP runs tasks in parallel via `amphp/parallel`, the file collection is serialized and shipped to worker processes. Previously each file was immediately unpacked into a full `SymfonySplFileInfo` object levereging a lot of memory. Now the worker receives plain strings and only builds the objects when something actually iterates over them
2. ensureFiles() - appends one file collection to another without duplicates. It now keeps known paths in a plain hash-map array, so duplicate checks are constant across time.
3. Added printing `Peak memory usage: X.X MB` at the end of commands for easier diagnostic (only in verbose mode)

### Results

Before

```

Running task 1/12: jsonlint... ✔
Running task 2/12: xmllint... ✔
Running task 3/12: yamllint... ✔
Running task 4/12: openapi... ✔
Running task 5/12: rector... ✔
Running task 6/12: behat... ✔
Running task 7/12: phpmd... ✔
Running task 8/12: phpcs... ✔
Running task 9/12: phpstan... ✔
Running task 10/12: psalm... ✔
Running task 11/12: test-unit... ✔
Running task 12/12: test-integration... ✔

Peak memory usage: 1152.0 MB
```

After

```

Running task 1/12: jsonlint... ✔
Running task 2/12: xmllint... ✔
Running task 3/12: yamllint... ✔
Running task 4/12: openapi... ✔
Running task 5/12: rector... ✔
Running task 6/12: behat... ✔
Running task 7/12: phpmd... ✔
Running task 8/12: phpcs... ✔
Running task 9/12: phpstan... ✔
Running task 10/12: psalm... ✔
Running task 11/12: test-unit... ✔
Running task 12/12: test-integration... ✔

Peak memory usage: 126.0 MB
```

### DIFF
--- a/src/Collection/FilesCollection.php
+++ b/src/Collection/FilesCollection.php
@@ -7,26 +7,17 @@ namespace GrumPHP\Collection;
 use Closure;
 use Doctrine\Common\Collections\ArrayCollection;
 use GrumPHP\Util\Regex;
-use Symfony\Component\Finder\Comparator;
-use Symfony\Component\Finder\Iterator;
 use SplFileInfo;
-use Symfony\Component\Finder\SplFileInfo as SymfonySplFileInfo;
 use Traversable;
 
 /**
- * @extends ArrayCollection<array-key, \SplFileInfo>
+ * @extends ArrayCollection<array-key, string>
  */
 class FilesCollection extends ArrayCollection
 {
     /**
-     * @param \Traversable<array-key, \SplFileInfo> $iterator
-     */
-    public static function fromTraversable(\Traversable $iterator): FilesCollection
-    {
-        return new self(array_values(iterator_to_array($iterator)));
-    }
-
-    /**
+     * @deprecated use FilesCollectionFilter::name directly
+     *
      * Adds a rule that files must match.
      *
      * You can use a pattern (delimited with / sign), a glob or a simple string.
@@ -39,10 +30,12 @@ class FilesCollection extends ArrayCollection
      */
     public function name($pattern): self
     {
-        return $this->names([$pattern]);
+        return FilesCollectionFilter::name($this, $pattern);
     }
 
     /**
+     * @deprecated use FilesCollectionFilter::names directly
+     *
      * Adds rules that files must match.
      *
      * You can use patterns (delimited with / sign), globs or simple strings.
@@ -53,12 +46,12 @@ class FilesCollection extends ArrayCollection
      */
     public function names(array $patterns): self
     {
-        $filter = new Iterator\FilenameFilterIterator($this->getIterator(), $patterns, []);
-
-        return self::fromTraversable($filter);
+        return FilesCollectionFilter::names($this, $patterns);
     }
 
     /**
+     * @deprecated use FilesCollectionFilter::notName directly
+     *
      * Adds rules that files must match.
      *
      * You can use patterns (delimited with / sign), globs or simple strings.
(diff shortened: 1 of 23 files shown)


src/Collection/FilesCollection.php
```diff
@@ -250,25 +273,54 @@ public function __serialize(): array
      */
     public function __unserialize(array $data): void
     {
-        $files = $data;
         $this->clear();
-        foreach ($files as $file) {
-            $this->add(new SymfonySplFileInfo($file, dirname($file), $file));
+        foreach ($data as $path) {
+            $this->add($path);
```
veewee, line 222: The signature however is `ArrayCollection<array-key, \SplFileInfo>`.
Now it also contains strings making it result in broken public API contracts (see e.g; the getFirst() vs first() and optionally similar methods like last, current, ...)

I'm also wondering: now every time an iterator goes over it, it will be hydrated.
So if you have more tasks (or file filters), it will always increase memory.

This makes me wonder: what is the ACTUAL memory issue that we are trying to solve?
Can we find a better solution than keeping it as strings intermediate?

It also has been a while but: what benifit does SymfonySplFileInfo  give over regular SplFileInfo here. Which one does add the overhead?

ckuran: Thank you for your comment. It allowed me to rethink the solution and actually understand where the problem is. As you already indicated, `FilesCollection` was holding `SplFileInfo` objects, and I broke the contract by adding pure paths there. I understood that the main problem is that `FilesCollection` has two responsibilities: holding the array of files and filtering the collection.

Now I separated those responsibilities into `FilesCollection`, which holds the array of **strings**, and `FilesCollectionFilter`, which filters the `FilesCollection`. The filter is converting the paths from strings to `SymfonySplFileInfo`, doing the filtering and returning the `FilesCollection` with strings. The `SymfonySplFileInfo` objects are garbaged immediately after filtering. Also we must use it for filtering instead of `\SplFileInfo`.

This solution is fully backwards-compatible, but I marked using the filtering methods in `FilesCollection` as deprecated, and it should be removed in v3. The result? Previously, my app was taking 126 MBs in peak memory usage. Now it's only 50 MBs for the same suite. I had to migrate a few places which were directly working on `SplFileInfo` objects to work with file paths. I haven't fixed phpspec yet, and the tasks themselves, but that can be covered later. Waiting for your opinion on that.

veewee: I consider this a massive breaking change in the contract of the FilesCollection, especially since its not final. + You added a new 'filter' collection class that isnt really a filter, but rather another type of files collection.

I don't think this is the way forward to be fair.

ckuran: I agree it's not a small change in FilesCollection, but the whole solution is in the final state. The contract for public methods in FilesCollection was not changed so that's why I consider this solution backward-compatible. The new 'filter' class can be renamed but basically it takes the FilesCollection as an argument and returns filtered FilesCollection. That was the easiest way to make it backward compatible. Running `bin/grumphp run` returns only failing phpspec, which is todo after we agree what next. If you have a better idea how to organize this change, please let me know.

veewee:
Thanks for sticking with this, and for the rethink after the first round. The second version is a clear improvement.

I want to be straight with you, though. The memory problem is real and your numbers hold up. We profiled it on our side too: most of the cost is in the parallel workers, where each one rebuilds the whole collection as `SplFileInfo` objects in `__unserialize`. An `SplFileInfo` is about 16x heavier than a plain path string, so with a dozen workers it grows quickly. Storing paths instead of objects is the correct fix.

The problem is the contract, and it runs deeper than it looks. `FilesCollection` is public and not final, and our plugin system hands it to third-party code we don't control. Those plugins add and remove files as `SplFileInfo`, and they also pull files back out and call `SplFileInfo` methods on them. We can keep the add and remove side working by converting to a path on input. Reading an `SplFileInfo` back out is what breaks. Doctrine's `ArrayCollection` has around twenty methods that return or pass elements (`first`, `last`, `getValues`, `slice`, `map`, `matching`, and so on), and once we store strings they all return strings. Covering that properly means overriding most of the base class, which is fragile, and it would still change `contains` and identity behaviour. We can't make that invisible to plugins within 2.x.

So this is a breaking change, and it really belongs in a major release. We agree with you that a string-based collection is the better design. Cutting a v3 for this one change is hard to justify on its own, and we don't have much else lined up for a major at the moment.

What I'd like to do is label this as a BC break and keep it parked for v3. The idea is sound and the work isn't lost, it just can't ship in 2.x without breaking the extension contract. I'll leave the PR open and tagged so we can pick it straight back up when a major starts.

Really appreciate the effort, and sorry it can't land sooner.

ckuran:
Well, there is a workaround in raising the `memory_limit` for GrumPHP when needed so I'll stick with that and we don't need to rush with merging it. If you need help with the branch like rebasing or fixing the phpspec tests, please let me know.

PR #1224, opened 2026-06-18
Support git autosquash `amend!` prefix in commit message task

reynkonig:
| Q             | A
  | ------------- | ---
  | Branch        | v2.x
  | Bug fix?      | no
  | New feature?  | yes
  | BC breaks?    | no
  | Deprecations? | no
  | Documented?   | no
  | Fixed tickets | no

`git commit --fixup=amend:` / `--fixup=reword:` produce an `amend! …` subject (sibling of `fixup!`/`squash!`), but `amend!` was missing from the three spots those prefixes are special-cased, so amend commits were rejected. Adds `amend` - centralized into one constant - with unit tests mirroring the existing `fixup`/`squash` cases.


veewee:
Thanks!

PR #1231, opened 2026-09-19
feat(phplint): add php_executable option for parallel-lint configuration

maarsson-io:
| Q             | A
| ------------- | ---
| Branch        | v2.x
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Documented?   | yes
| Fixed tickets | -

## Description

This PR adds an optional `php_executable` configuration option to the `phplint` task.

When configured, the task passes the value to `parallel-lint` using its native `-p` option. If omitted, the current behaviour is preserved.

This makes it possible to explicitly select the PHP executable used by `parallel-lint`, which is useful in environments where the default executable cannot be used reliably (for example, versioned PHP installations such as Laravel Herd).

The change is fully backwards compatible and includes unit tests for the new configuration option.


maarsson-io:
The CI failure appears to be unrelated to this change. All other matrix jobs are green, and the failing tests are E2E tests rather than tests related to the PhpLint task. The stack trace also points to a filesystem/Git maintenance issue (maintenance.lock) during the E2E test cleanup rather than the modified code.

veewee:
Thanks for the PR.

